PR #6, opened 2026-07-29
fix(grid): the record verbatim, an order that reads it, and tests that watch

muhammadbayoumi:
The Data page (`/source/<key>`), reviewed against the project's five rules and repaired. One commit, eight files.

## The record is the record

`table_payload` and `export_source_table` were collapsing whitespace in the values they handed out — names, brands, categories, variants, SKUs. That bought the right **order** by editing the **value**, and it reached the xlsx and Sheets export, so a brand the shop published as a single space left the building as an empty cell.

Nothing normalises on the way out now. The order is fixed where an order belongs: the grid compares a collapsed form when it sorts, when it lists the values a filter offers, and when it buckets a branch — and shows and exports exactly what was captured. `TRIM` stays in the export's `ORDER BY`; it decides the order, never a cell.

## Scraped markup ran

Tabulator renders a **string** group header through `element.innerHTML` and a **node** through `appendChild`. The header was built by concatenation, so grouping by a column whose scraped value contains markup executed it. Demonstrated, not argued: with the old version restored, the test's page really does set `window.__pwned`.

## A control that could remove its own door

Choose Columns was reachable only from the three-dot menu on a column header — and its "Show all columns" tick sends every column to the details. No columns, no headers, no menu, no way back to the chooser or Reset Columns short of the API. The command bar carries a Columns button now, wired before the empty-source return so a dataset with no rows can still be arranged. The chooser also refused to close while a save kept failing, on every press; asking twice now means "close it anyway".

## Counts and branches that lied

- The footer read the state **before** the filter, permanently one behind — "Total Rows: 0" on a table showing 3,417. `dataFiltered` is dispatched from inside Tabulator's filter routine, before the filtered rows become the active set.
- Filtering while nested deleted whole branches: a parent row carries only the nested column, so it failed every other predicate and took its matching children with it — 3,417 rows became "0 of 19". A branch is kept when any child matches.
- Grouping or nesting by Tax put the entire table under one blank heading; its own field is null on every row and the cell renders from `tax_states[tax_ref]`.

## Tests that can actually fail

`tools/grid_harness.py` loads the real `grid.js` against a fixture payload in a real browser — the same trick `panel_harness.py` already plays for the extension panel, which the Data page was the one surface not using. **18 tests** in `tests/test_grid_dom.py` drive the table and ask what it did.

Each fix above was re-broken on purpose and the matching test failed. One brittle substring assertion in `test_vendor.py` is relaxed, and says why: it pinned `groupedBy.slice()` and broke the moment a derived column needed a reader function.

## Also

- `externalLink()` replaces six copies of `target=_blank` + `rel=noopener`. Reading the staged diff caught the helper calling **itself** after a bad automated rewrite — no existing test rendered a link, so all 17 passed over infinite recursion. There is a test for it now.
- 44 hand-written `1px solid var(--line)` become `var(--table-rule)`, the token `table-theme.css` declares for exactly this.
- `remember_` becomes `rememberChoice`.

## Verification

- Full suite on this branch: **no failures** (excluding `test_panel_dom` and `test_connector_live_defects`, which need a browser session and the network and are untouched here).
- `tests/test_grid_dom.py`: **18 passed**.
- Rebased onto `efbe7ca`; the only overlap was `scrapex/tax.py`, checked by hand — their change renames tax_rule's SQL columns, mine removes a duplicate dict key in `as_dict()`. Orthogonal, and no duplicate key remains (verified by AST scan).

## Known and deliberately left

21 column keys are written as literals in `grid.js` and authored again in `reports.py`'s `BROWSE_COLUMNS`. That split has already rotted twice in this file's own history. Fixing it means the server declaring each column's ROLE (`money`, `date`, `link`, `numeric`) — a contract change too large to hide in this commit.

### DIFF
--- a/scrapex/reports.py
+++ b/scrapex/reports.py
@@ -6,7 +6,6 @@ owner curates). This directly answers "did anything actually land?".
 """
 from __future__ import annotations
 
-import re
 import sqlite3
 from dataclasses import dataclass, field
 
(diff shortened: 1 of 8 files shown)


muhammadbayoumi:
Superseded — closing rather than merging.

This branch's content is already on `main` as **546ec34** (`fix(grid): preserve source text and harden table behavior`). Verified by content, not by title:

- 5 of the 8 files hash **byte-identical** between this branch's commit and 546ec34 (`reports.py`, `tax.py`, `grid.js`, `tests/test_grid_dom.py`, `tools/grid_harness.py`).
- The other 3 differ only because **3e4e0c6** later added to them; every change from here survives in main today — the group-header node fix, `collapse()`, `externalLink()`, the `queueMicrotask` footer, the branch-aware filter, `#grid-columns-button`, `rememberChoice`, and zero remaining `1px solid var(--line)` literals.
- Rebasing this branch onto current main resolves to an **empty commit and a zero diff**.

So there is nothing to merge, and merging would duplicate 810 lines.

**On the two CI alerts:** the 49 failures here (43 `test_panel_dom` + 5 `test_phase6_review_fixes` + 1 `test_compaction`) are exactly main's own failures at this branch's base commit efbe7ca — none originate here. Main has since fixed the panel tests in d36c3d8. The merge conflict is the same change arriving from two ancestries; resolving it toward this branch would *delete* the 32 lines 3e4e0c6 added on top.

PR #7, opened 2026-07-29
fix(storage): a sealed archive keeps the rows it promises to keep

muhammadbayoumi:
Fault A of four in the archive/compaction cluster that has kept CI red.

**The promise, from `compaction.py`:** *ScrapeX never unlinks it. Every observation that ever existed is still in that file.* On macOS and Linux that was false, and the failure was total.

## What was happening

Every connection is WAL-mode, and SQLite folds the log back into the main file only when the LAST connection closes. During a compaction the caller still holds the warehouse open, so nothing folds. Measured at the moment of sealing:

    main file 4,096 bytes      write-ahead log 1,709,832 bytes

`_retire` then renamed the main file and deleted the `-wal` beside it. On Windows `os.replace` raises (the open handle blocks it) so the unlink was never reached — **that is the only reason this was never seen.** On POSIX the rename succeeds while handles are open, the log is orphaned by the new name and then unlinked, and the archive becomes a husk: reading its main file alone gives `no such table: scrapex_meta`.

## The fix

`drain_wal()` runs `PRAGMA wal_checkpoint(TRUNCATE)` and reports whether the log is really gone. `_retire` calls it **before** the rename and renames only on success. After: main file 643,072 bytes, log 0, and the file alone reads its seal, its rows and a healthy health check.

*Not deleting the -wal would NOT have fixed this* — the rename orphans the log by itself. The checkpoint is the fix.

When the checkpoint reports BUSY the file stays where it is: a superseded database that still reads beats a tidy name over an empty one, and the seal inside is what the guards read.

## Tests

`test_a_sealed_archive_is_readable_from_its_own_file_alone` asks the promise itself, platform-blind: copy the archive's own file away from every sibling, then read it. **Re-broken on purpose** — fails with *the archive carries no seal once it stands alone* — then restored.

Full suite on Windows: **1465 passed, 1 xfailed, 0 failures.** CI is the real test, because the fault only bites on Linux.

This fault also owns `test_the_seal_is_recorded_inside_the_file_not_only_in_its_name`, `test_undoing_a_compaction_makes_the_archive_live_again` and `test_compaction.py::test_a_compaction_can_be_undone_because_nothing_was_deleted`. Faults B, C and D of the cluster are separate and not in this PR.

### DIFF
--- a/scrapex/storage.py
+++ b/scrapex/storage.py
@@ -1006,6 +1006,44 @@ def migrate_location(db_path: Path | str, new_dir: Path | str, *,
                 "superseded — do not point ScrapeX back at it by hand."))
 
 
+def drain_wal(db_path: Path | str) -> bool:
+    """Fold a database's write-ahead log into the file itself.
+
+    WHY THIS EXISTS, and it is not housekeeping.
+
+    Every connection here is WAL-mode (`db.connect` sets `PRAGMA journal_mode =
+    WAL`), and SQLite only folds the log back into the main file when the LAST
+    connection closes. During a compaction or a move the caller still holds the
+    warehouse open, so nothing folds — and measured on a real warehouse at the
+    moment of sealing, the main file was 4 KB while the log was 1.7 MB. The
+    whole warehouse was in the log.
+
+    That is what made the two lines below a DATA-LOSS bug: `_retire` renamed the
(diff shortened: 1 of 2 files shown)


muhammadbayoumi:

PR #8, opened 2026-07-29
fix(capture): the in-flight guard reads the seal, not a filename

muhammadbayoumi:
Fault C of four in the archive/compaction cluster. **The guard was defeated by the thing it guards against.**

## What was happening

`_refuse_if_superseded` exists for one scenario, and its own docstring states it: `connector.fetch` runs for minutes holding no lock, so a crawl can return to a handle on a database that was sealed and replaced while it worked. Those rows would land in the archive and be invisible to the live warehouse forever.

It identified the warehouse by `PRAGMA database_list` — which reports the path as it was when the **connection opened** — and passed that path to `storage.sealed_at`. A compaction **renames** that file. On macOS and Linux the rename succeeds while handles are open, so the recorded path stops existing, and `sealed_at` opens with:

```python
if not Path(db_path).exists():
    return ""
```

So the guard read *not sealed*, said nothing, and let the crawl write into the sealed archive. Windows hid it: there the rename fails while a handle is open, so the path still resolves.

## The fix

Ask this crawl's **own handle**:

```sql
SELECT value FROM scrapex_meta WHERE key = 'sealed_at'
```

The handle is attached to the database, not to a name, so a rename cannot blind it — and a seal another connection committed is visible to it. A database older than migration 0002 has no `scrapex_meta`; nothing can have sealed it, so that case returns quietly. The path is still used in the message, where it belongs — it tells the owner where the rows would have gone.

## Tests

`test_the_inflight_guard_does_not_depend_on_the_path_it_opened` asks the **property**, not the mechanic, because Windows cannot reproduce the mechanic — it refuses to rename a file a handle holds, which is exactly why this went unseen. With `sealed_at` stubbed to answer as it does for a vanished path, the guard must still refuse; it can only do that by reading through its own handle.

**Re-broken on purpose** by restoring the path-based read — *DID NOT RAISE WarehouseSupersededError* — then restored. Two more tests hold the edges: a live warehouse is not refused, and a pre-0002 database passes without raising or crashing.

Full suite: **1466 passed, 1 xfailed, 0 failures.**

Independent of #7 (Fault A). Faults B and D are separate and not in this PR.

### DIFF
--- a/scrapex/capture.py
+++ b/scrapex/capture.py
@@ -32,6 +32,18 @@ def _refuse_if_superseded(conn: sqlite3.Connection) -> None:
 
     Checked HERE, inside the lock and immediately before the insert, because
     that is the only point where the answer cannot go stale again.
+
+    THE SEAL IS READ THROUGH THIS CRAWL'S OWN HANDLE, not by reopening a path.
+    `PRAGMA database_list` reports the file as it was named when the connection
+    opened, and a compaction RENAMES it: on macOS and Linux that rename succeeds
+    while handles are open, so the recorded path no longer exists. The previous
+    version passed that stale path to `storage.sealed_at`, whose first line
+    returns "" for a path that is not there — so the guard saw no seal, said
+    nothing, and let the crawl write into the sealed archive. Exactly the
+    disaster the paragraphs above describe, produced by the guard against it.
+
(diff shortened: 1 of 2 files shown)


muhammadbayoumi:

PR #9, opened 2026-07-29
fix(storage): say superseded when superseded, and refresh the statistics claimed

muhammadbayoumi:
Faults **B** and **D**, the last two of the archive/compaction cluster. Both are small; both are the code saying something that is not true.

## B — a compaction is not a move

`mark_sealed` stores `"{reason} -> {successor}"`, and two reasons reach the reader: `moved`, a relocation the owner asked for, and `sealed`, a compaction that replaced the file with a smaller successor. `_successor_recorded_nearby` split on `" -> "`, returned the target and **threw the reason away** — so `resolve_db_path` worded both as *this machine's warehouse was moved to …*.

An owner who never moved anything is told their warehouse was moved, and sent looking for a decision they did not make. The reason now travels with the target and the sentence follows it.

This surfaces precisely **because Fault A (#7) was fixed**: with the log drained the rename now succeeds on POSIX, so the default path really is absent and this branch is the one that answers.

## D — "statistics refreshed" was not always true

`repair()` runs `REINDEX` then `PRAGMA optimize` and reports *Indexes rebuilt and statistics refreshed*. `PRAGMA optimize` only analyses tables the **connection** has queried, and before SQLite 3.46 it never analyses one that has no statistics yet — and `repair`'s connection has issued no queries. On ubuntu-24.04's SQLite **3.45**, which is what CI runs, it analysed nothing, produced no `sqlite_stat1`, and the sentence was false.

`ANALYZE` is the unconditional form, and it now runs **only when optimize left nothing behind**, so a newer SQLite keeps its cheaper incremental path.

Fixed in the **product**, not the test. The failing test was asserting something true — that Repair produces planner statistics — and weakening it would have kept a false sentence in the interface.

## Tests

- `test_a_compacted_warehouse_is_reported_as_superseded_not_moved`, with a counterpart holding that a real move still says *moved to*. **Re-broken on purpose**: the compaction case reports *was moved to* and the test fails.
- `test_repair_refreshes_statistics_on_every_sqlite_version` asserts the **outcome** rather than the mechanism, so it holds on both versions.

**Stated plainly:** I could not make D's test fail on this machine and did not pretend to — local SQLite is 3.50.4, where optimize behaves; CI's 3.45 is where the fault bites. B's test was re-broken and observed failing.

Full suite: **1466 passed, 1 xfailed, 0 failures.**

Closes the cluster with #7 (A) and #8 (C). Independent of both.

### DIFF
--- a/scrapex/storage.py
+++ b/scrapex/storage.py
@@ -272,21 +272,35 @@ def resolve_db_path() -> Path:
         # Nothing at the default path, but a sealed sibling means this machine
         # HAS a warehouse — it moved. Starting an empty one here would leave the
         # real history sitting somewhere the owner is never told about.
-        moved_to = _successor_recorded_nearby(fallback)
-        if moved_to:
+        why, went_to = _successor_recorded_nearby(fallback)
+        if went_to:
+            # A compaction and a move leave the same absence behind and mean two
+            # different things. Saying "moved" for a compaction sends the owner
+            # looking for a decision they never made.
+            became = ("was superseded by a compacted database at" if why == "sealed"
+                      else "was moved to")
             raise StorageUnavailableError(
                 f"There is no database at {fallback}, but this machine's warehouse "
-                f"was moved to {moved_to}. ScrapeX will not start an empty one in "
+                f"{became} {went_to}. ScrapeX will not start an empty one in "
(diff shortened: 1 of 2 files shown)


muhammadbayoumi:

PR #43, opened 2026-07-30
feat(images): the pictures both SSR shops publish in markup but not in JSON-LD

### LINKED ISSUE #35: Gallery order is discarded at read time: the card shows the alphabetically-first picture, not the shop's primary one
Every connector ranks a product's pictures and encodes that rank in
`attribute_code` (`image`, `image_1`, `image_2`, …). The read path throws the
rank away and re-sorts the pictures **alphabetically by filename**, so the
picture shown as a product's main image is usually not the one the shop chose.

Found while auditing image capture for #38 / PR #23. Not fixed there: that PR is
about *capturing* images, this is a *read-side* defect that needs no re-crawl.

## The cause

`scrapex/reports.py:1379` — the query behind the product card:

```sql
SELECT spa.attribute_group, spa.attribute_label, spa.attribute_code,
       spa.raw_value, spa.value_url, ...
 ORDER BY spa.attribute_group, spa.attribute_label, spa.raw_value LIMIT ?
```

`attribute_code` is selected (`reports.py:1372`) but never ordered by. And all
six connectors write the **identical** label `"Image"`:

- `custom_json.py:594`, `hybris.py:212`, `jsonld.py:315` (salla + zid),
  `magento.py:1278`, `shopify.py:218`, `woocommerce.py:436`

So `attribute_group` ties, `attribute_label` ties, and the real sort key is
`raw_value` — the filename or alt text, alphabetically.

The UI takes that order at face value: `scrapex/webui/static/grid.js:2473-2478`
filters `group === "Media"` and renders `images[0]` as the main picture, with no
sort of its own.

`scrapex/reports.py:1166-1167` (the details **export**) is worse — it orders by
`attribute_group, attribute_label` with no third key at all, so image order
there is whatever SQLite returns.

## Reproduction

`tests/fixtures/sikaegshop_detail_252.json`, product 252 — 9 image attachments.
Sorted the way `reports.py:1379` sorts them:

```
1. 170A3566.jpg          <-- shown as the main picture
2. 170A3566a.jpg
3. 170A3570.jpg
4. 170A3599.jpg
5. 170A3604.jpg
6. Kleben24.jpg
7. Sika Backing Rod  (1).jpg   <-- the shop's own is_primary: true
8. Sika Backing Rod  (3).jpg
9. sikarod.jpg
```

The shop's chosen primary picture sorts **7th of 9**.

## This is actually two defects

**A. The rank is discarded at read time.** Affects `magento`, `salla`, `zid`,
`shopify`, `hybris`, `woocommerce` — these all encode the site's own gallery
order in the code (`image`, then `image_1`…). The information is in the
warehouse; only the query loses it.

**B. SIKAEGSHOP has no rank to recover, and its primary flag was never
captured.** `custom_json.py:592-594` keys on the attachment id
(`image_{attachment_id}` → `image_230`, `image_2590`, …), deliberately — the
comment at `custom_json.py:583-586` explains an id survives the shop reordering
its list, which a position would not. That reasoning is sound, but it means no
read-side sort can restore primacy for this source.

Worse: the payload states `is_primary` per attachment and **nothing in the
codebase reads it** — `grep -rn "is_primary" scrapex/` returns nothing. The shop
tells us which picture is the main one and we drop it at capture time.

So B needs a capture-side change (store the flag, or a position beside the id)
and a re-crawl; A does not.

## Notes for whoever fixes it

- **Do not just `ORDER BY spa.attribute_code`** — it is lexicographic, so
  `image_10` sorts before `image_2`, and bare `image` sorts after both. It needs
  a natural/positional sort, or better, a stored position.
- Sorting in SQL keeps the export (`reports.py:1166`) and the card
  (`reports.py:1379`) consistent; sorting in `grid.js` fixes only the card.
- The cleanest shape is probably a real position column on
  `source_product_attribute` so the rank stops being smuggled inside a string —
  that would also give B somewhere to put `is_primary`. That is a migration, so
  it is the owner's call whether it is worth it versus a natural sort.

## Not covered by tests

`tests/test_enrichment.py:232` calls `product_attributes` but only asserts the
result is non-empty. No test asserts image ORDER anywhere, which is why six
connectors could each carefully rank their pictures and nobody noticed the
ranking never arrived.

muhammadbayoumi:
**DRAFT — do not merge.** Opened for the main session's review as instructed.

Research first. The short answer is that **there is no general mechanism worth
building**, and the reason is a finding rather than a shrug: the gap is not
"markup vs machine-readable data", it is **summary format vs product record**,
and it exists on exactly two of the ten sources.

---

# 1. What was measured, and on what

Full catalogues, not samples. One request per product, ≥1.2 s apart, the user
agent each source declares, and the script stops the moment a site answers 403
or 429. Neither did. **No image file was downloaded** — the URL is the record.

| source | connector | products fetched | requests |
|---|---|---|---|
| ADVANCEDCASTLE | `zid.py` | **168 of 168** | 168 |
| ALSWEED | `salla.py` | **1,231 of 1,233** (2 pages did not answer) | 1,231 |
| ELSEWEDYSHOP | `shopify.py` | 8 products, API vs page | 9 |
| SAMEHGABRIEL | `woocommerce.py` | 4 products, API vs page | 5 |

The brief warned the coverage numbers might be stale, so nothing here is quoted
from a previous session — the two SSR catalogues were re-crawled end to end
today, and the two API spot-checks were run today.

---

# 2. The measured gap, per source, on the full catalogue

**The 42-product sample that opened this question understated it.** It found 15
missing of 124. Over all 168 products the real figure is **58 missing of 493**.

| source | machine-readable route we read | what it names | what the page names | gap |
|---|---|---|---|---|
| **ADVANCEDCASTLE** | Product JSON-LD | 435 | **493** | **58 on 44 products** |
| **ALSWEED** | Product JSON-LD | 1,170 | **3,322** | **2,162 on 749 products** |
| ELBUROJ | Product JSON-LD | *not measured — see ceilings* | | expected same as ALSWEED |
| ELSEWEDYSHOP | `/products.json` | 1 per product | **fewer** — page markup renders 0 product images server-side | **none** |
| SAMEHGABRIEL | WooCommerce Store API | 6 / 1 / 3 / 6 | 6 / 1 / 3 / 6 — **identical** | **none** |
| SIKAEGSHOP | `/api/products/{id}` | all attachments | — | **none, already solved** |
| MASDAR · MADAR | OCC / GraphQL | platform record | — | **none by construction** |
| HEIDELBERG_EG · GPP_ENERGY · ARAMCO_FUEL_SA | — | — | — | **no product imagery at all** |

Two of those rows are worth reading twice.

**ELSEWEDYSHOP inverts the brief's premise.** Its page markup carries *fewer*
pictures than the API: the theme renders the gallery client-side, and the four
`cdn/shop` URLs in the HTML are `_1200x1200`, `_1024x`, `_crop_center`
renditions of the single file `212549.png` that `products.json` already names.
A markup route there would add nothing and, keyed naively, would have added
three phantom pictures.

**SAMEHGABRIEL nearly produced exactly that error, and it is the best argument
in this PR.** An unscoped sweep of `` tags containing `/uploads/` reported
23–25 "page-only" pictures on *every* product — a suspiciously constant number,
because they were the related-products carousel and site chrome. Scoped to
`.woocommerce-product-gallery`, the API and the page agree exactly: 6/6, 1/1,
3/3, 6/6, **zero either way**. A general "read the gallery" mechanism has to
guess that container per platform, and guessing it wrong does not under-report,
it **invents product associations the shop never made**.

---

# 3. Where each site's fuller truth lives, and how long it survives

Ranked most durable first, per the brief's own hierarchy.

| shop | route chosen | tier | why this one |
|---|---|---|---|
| **zid** (ADVANCEDCASTLE) | `var productImages = [...]` in-page JSON | **JSON bootstrap** — survives a re-skin | The data the theme renders *from*: stable image `id`, `alt_text`, `display_order`, five named sizes. A platform API would rank higher, but zid's own `/api/` is **`Disallow`ed by advancedcastle's robots.txt**, so the in-page bootstrap is both the most durable *and* the only polite route. |
| **salla** (ALSWEED, ELBUROJ) | `a[data-fslightbox="product_{id}"]` slide anchors | **data-attribute contract** — survives some | No JSON bootstrap exists on salla; this is the most durable thing that does. The attribute is keyed on the product id we already have, so a related-products carousel on the same page cannot leak in. |
| *(rejected)* | `#product-images img.carousel-img` | **CSS class — breaks on the next theme update** | This is what the opening sample measured. Rejected on durability, and it is also *measurably wrong*: on 4 of 168 products it returns pictures the page's own list does not have. |
| *(rejected)* | salla's `` inside the slider | CSS/DOM | Lazy-loaded. Every slide after the first carries `cdn.salla.network/images/s-empty.png`, so this route would store **one placeholder as several pictures**. The anchors also appear once per picture where the ``s appear twice. |

---

# 4. Why no general mechanism — and what *is* shared

Every API-backed connector already reads the platform's own product record, and
that record **is** the ceiling. The gap appears only where we read
**schema.org JSON-LD**, which is a *summary*: the spec permits `image` to name
one representative picture, and both SSR shops do exactly that.

So the two readers have nothing in common worth abstracting: one parses JSON
with an id field, the other queries DOM anchors with a data attribute; one shop
publishes a picture at five addresses, the other at one. **A shared "gallery"
concept would fit neither.**

What *is* common is everything after reading, and that is the shared helper in
`jsonld.py`:

- `Picture(identity, url, label)` — [`jsonld.py:319`](scrapex/connectors/jsonld.py:319)
- `merge_pictures(record, summary)` — [`jsonld.py:347`](scrapex/connectors/jsonld.py:347)
- `jsonld_pictures(node, identity=…)` — [`jsonld.py:374`](scrapex/connectors/jsonld.py:374)
- `WalkTally.picture_route_defects()` — [`jsonld.py:97`](scrapex/connectors/jsonld.py:97)

Per-connector: `zid.page_pictures` + `zid.image_identity`
([`zid.py:194`](scrapex/connectors/zid.py:194),
[`zid.py:182`](scrapex/connectors/zid.py:182)) and `salla.page_pictures`
([`salla.py:89`](scrapex/connectors/salla.py:89)).

---

# 5. No duplicates — the trap that would have passed review

**Of advancedcastle's 435 JSON-LD image URLs, ZERO are string-equal to a
gallery ``, and ZERO name a picture the gallery lacks.** They are the
same pictures at different addresses — JSON-LD publishes the `large` rendition,
the gallery renders `full_size`.

A URL-keyed merge would therefore have stored **every picture twice, under two
attribute codes**, while every spot-check looked correct. Pictures are merged on
**identity**, which each connector defines for its own shop:

- zid → the media uuid (`image_identity`), because one picture has five URLs.
- salla → the URL, *checked*: 1,160 of 1,170 JSON-LD URLs are byte-identical to
  a slide `href`, and no product repeats an href.

And zid records the **`large`** size specifically because that is the one its
own JSON-LD names — **429 of 429 matched URLs byte-identical, 0 differ**. That
makes the merge provably duplicate-free *and* means **no URL already in the
warehouse churns**. `full_size` is published too and is a larger file on 406 of
464 pictures; switching is one line and I have left it as the owner's call
rather than change every stored link inside a coverage fix.

**The summary is merged in, never replaced by.** On 4 of 168 advancedcastle
products the two lists disagree and 6 pictures exist only in the JSON-LD; on
alsweed, 10 YouTube thumbnails do. Replayed over every censused page:
**0 pictures lost.**

---

# 6. The drift guard — the part that matters more than the extra images

A markup reader's failure mode is **silence**. It matches nothing, the prices
still land, every product keeps the one picture its JSON-LD names, and the run
reports plain success forever.

`WalkTally` now counts the pages where the route was readable and the pages that
published images while it was not, and raises a **defect — not a warning** —
when it fails on more pages than it works on. Per `base.py`'s own distinction,
only a defect reaches `crawl_run.errors_count`.

[`tests/test_product_pictures.py`](tests/test_product_pictures.py) leads with
the tests that **fail if a drift goes unannounced** — the bootstrap and the
anchor attribute are renamed, everything else still works, and the run is
required to say so:

```
test_a_zid_theme_that_stops_publishing_its_picture_list_is_announced
test_a_salla_theme_that_renames_its_slides_is_announced
test_a_healthy_run_raises_no_picture_defect          # and stays quiet when fine
test_the_defect_fires_only_when_the_route_fails_more_than_it_works
```

**20 tests in this file, and the full suite is green — 1,637 items, 0 failures.**

---

# 7. Measured before / after

Replaying the **shipped** merge functions over every censused page — the real
code, not an estimate:

| source | image rows before | after | gained | products gaining | **lost** | complete? |
|---|---|---|---|---|---|---|
| **ADVANCEDCASTLE** | 435 | **499** | **+64** | 48 of 168 | **0** | ✅ complete — every picture the page names |
| **ALSWEED** | 1,170 *(would be, once enrichment is declared)* | **3,332** | **+2,162** | 749 of 1,231 | **0** | ✅ complete — 61 products publish none and stay blank |
| ELBUROJ | — | — | — | — | — | ⚠️ **stated ceiling**, not measured |
| ELSEWEDYSHOP · SAMEHGABRIEL · SIKAEGSHOP · MASDAR · MADAR | *unchanged* | | | | | ✅ already at the platform ceiling |
| HEIDELBERG_EG · GPP_ENERGY · ARAMCO_FUEL_SA | 0 | 0 | | | | ✅ no product imagery published |

190 advancedcastle rows and 3,322 alsweed rows now carry the shop's **own** alt
text instead of a file name.

### The ceilings that remain, stated plainly

- **ELBUROJ** is `active: false` and asks for a 10-second crawl delay; a
  3,441-product census would be ~9.5 hours of knocking. It runs the same
  connector as ALSWEED and is *expected* to behave identically. That is an
  expectation, not a measurement, and is the one thing here a future crawl
  should confirm rather than assume.
- **Videos are not pictures.** 13 alsweed slides are `data-type="youtube"` and
  are not read as images. Where the JSON-LD separately lists a YouTube
  *thumbnail*, the merge keeps that row because it already existed.
- **No second language to capture.** advancedcastle serves the **same Arabic
  `alt_text` on its English locale** (verified on the same product, both URLs),
  and salla's connector fetches one page per product — asking for a second
  locale would double every crawl. So there is no bilingual pair, no `_ar`
  column, and **no contract change**: `PAYLOAD_VERSION`, `PAYLOAD_COMPAT_VERSION`
  and `GENERATION_OF_VERSION` are untouched. Images already live in
  `source_product_attribute`.

---

# 8. Coordination — #23 has landed, #35 has not

Re-checked against `origin/main` immediately before this push, not from memory.

## #23 is MERGED — the earlier "dormant" caveat in this PR is void

`origin/main` has moved 15+ commits since I branched. **PR #23 merged**, which
means its `enrichment` manifest lines for ALSWEED and ELBUROJ are now on main.
So salla's reader here is **live, not dormant**, and ALSWEED's 3,332 pictures
land as soon as this merges.

`origin/main` is merged into this branch (`7999b2e`). Main had also touched
`salla.py` and `zid.py` since I branched (the `declare_frontier` progress-bar
work) — **the merge was clean, and I verified `declare_frontier` survived it**
in both connectors rather than trusting the auto-merge. Full suite after the
merge: **1,724 passed, 1 xfailed, 0 failures**, including #23's new
`tests/test_images_every_connector.py`.

I still did not touch `hybris.py`, `shopify.py` or `magento.py`, and my
`sources.yaml` edit remains confined to ADVANCEDCASTLE's notes.

## #35 — open, no branch, no PR, and **no file conflict expected**

Issue #35 is a **read-side** defect: `reports.py` throws away the rank each
connector encodes in `attribute_code`. It touches `scrapex/reports.py` and
`scrapex/webui/static/grid.js`. **This PR touches neither** —
`git diff origin/main --stat -- scrapex/reports.py scrapex/webui/static/grid.js`
is empty. I expect **no merge conflict in either direction.**

But there is a **semantic interaction, and it runs in #35's favour** — three
measured facts that session should have:

**1. I changed the column the read path actually sorts on.** Because every
connector writes the identical label `"Image"`, #35 correctly identifies
`raw_value` as the real sort key. This PR changes `raw_value` for image rows
from *always the filename* to *the shop's own alt text when it published one*
([`jsonld.py:435`](scrapex/connectors/jsonld.py:435)).

**2. That makes the current ordering non-deterministic, not just wrong.** Salla
repeats one `data-caption` across a product's slides, so after this PR the
`raw_value` sort key **ties within a product on 719 of 1,170 alsweed products**
(and 11 of 168 on advancedcastle). Tied rows come back in whatever order SQLite
chooses. So this PR turns #35's symptom from *stably wrong* into *arbitrary* on
those products — which makes landing #35 more valuable, not less, and means
"it looked fine yesterday" will not be evidence of anything.

**3. A latent trap in the obvious fix.** `ORDER BY attribute_code` sorts as a
**string**, so `image_10` falls between `image_1` and `image_2`. That is safe
*today* only because no product has 11 pictures — measured maxima after this PR
are **9 (advancedcastle)** and **10 (alsweed)**. This PR raises that ceiling
substantially (alsweed went from ~1 picture per product to as many as 10), so
the eleventh picture is now a plausible near-term event. **#35 should sort on
the numeric rank, not the code string.**

Line numbers in the issue have drifted with main; current positions:

| issue says | now |
|---|---|
| `reports.py:1379` (card ORDER BY) | [`reports.py:1492`](scrapex/reports.py:1492) |
| `reports.py:1372` (select list) | [`reports.py:1485`](scrapex/reports.py:1485) |
| `reports.py:1166-1167` (details export, no third key) | [`reports.py:1275`](scrapex/reports.py:1275) |
| `jsonld.py:315` (the `"Image"` label) | [`jsonld.py:435`](scrapex/connectors/jsonld.py:435) |

**I have deliberately not fixed #35 here.** It is a read-side defect that needs
no re-crawl, it is that session's ground, and the write side this PR provides —
the shop's own order, encoded in `attribute_code`, main picture first — is
exactly what their fix needs to read. Two sessions in `reports.py` is how a day
gets reverted.

Migrations untouched (0056/0057 not involved). The live warehouse was never
written to; every "after" figure is a replay of the shipped code over captured
pages.

---

# 9. Politeness

Roughly 1,450 requests in total, all GET: 168 for the advancedcastle census,
1,233 attempted for alsweed's, 5 sitemaps, 14 for the API spot-checks, 4
robots.txt, and about 20 exploratory page reads. All >=1.2 s apart, with the
user agent each source declares, and a hard stop on 403/429 that never fired. `robots.txt` was read for every host
touched: advancedcastle allows `/products/` and disallows `/api/` (which is why
no platform API was tried), alsweed allows `/` with no crawl delay. **No extra
request is added to any crawl** — salla now keeps the HTML it already fetched
instead of discarding it, and zid reads a bootstrap that was already in the
response.

Fixtures and their provenance:
[`tests/fixtures/live/product_pictures_2026-07-30.CAPTURE.md`](tests/fixtures/live/product_pictures_2026-07-30.CAPTURE.md).


muhammadbayoumi:
## Not merged, and the reason is on `main` rather than in this diff

Rebased onto current `main` (79 commits ahead) in an isolated worktree and reviewed the **merged result**, not this patch. Three findings, in order of weight.

### 1. The zid half was deliberately held, six days ago, for a measured reason

`tests/fixtures/live/product_pictures_2026-07-30.CAPTURE.md` on `main` says so in its own words:

> **The ADVANCEDCASTLE column does not ship in this branch.** Its zid reader was split out and held: re-checked 2026-08-05, that site's sitemap now lists 2 product URLs, 8 of 8 stored product links answer 404, and the URL this fixture was captured from is itself 404.

**The site stopped serving the catalogue this work was measured against.** Every number in this branch — 435 → 499 image rows, 48 products gaining, 429 of 429 URLs byte-identical — was true on 30 July and describes pages that now answer 404. Merging would land a reader for a catalogue that is not there, and reverse a decision someone made on purpose with evidence.

### 2. Two of the three code files are already on `main`

`scrapex/connectors/jsonld.py` and `scrapex/connectors/salla.py` are **byte-identical** to `main` — the salla half landed as #102. The only content still unique here is the zid gallery route.

### 3. A naive merge loses coverage in both directions

Ten conflicts. Nine are pure additions. The tenth is the block `main` calls *"salvaged from the tests that covered BOTH readers before the zid half was split out"* — taking either side whole is wrong:

| test | on `main` | here | if you take one side |
|---|---|---|---|
| `test_a_healthy_run_raises_no_picture_defect` | salla only | **both** | `main`'s is narrower |
| `test_no_two_image_rows_share_a_url_or_a_code` | salla only | **both** | `main`'s is narrower |
| `test_a_picture_only_the_summary_names_is_never_dropped` | **salla** | **zid** | one silently shadows the other — same name, different shop |
| `test_a_page_with_no_pictures_at_all_yields_no_image_rows` | **salla** | **zid** | same |
| `test_the_merge_keys_on_identity_and_not_on_the_url` | **only here** | — | lost |

Python takes the last definition of a duplicated name without a word, so the doubled pair would have looked green while covering one shop instead of two.

I authored the resolution that preserves every assertion from both sides, then discarded it on finding (1). **Nothing was pushed and this branch is untouched.**

### What would revive it

Re-measure `advancedcastle.com`. If the catalogue is back, the zid route is worth landing on its own — `image_identity`, the `productImages` bootstrap over a theme class, and the id-keyed merge are all well argued and well measured. If the catalogue is gone, this is a reader for a shop that no longer publishes, and the salla half already shipped.

Closing rather than leaving it open a third month. The branch stays; reopening costs nothing.

Issue #32, opened 2026-07-30
Technical Requirements


muhammadbayoumi:
### 1.1 Version Updates

The tool must use a structured versioning mechanism. The version number must be updated whenever a major functional, architectural, or behavioral change is introduced.

### 1.2 Version Visibility

The currently installed version must be clearly displayed within the Extension and, where applicable, in the web interface.

### 1.3 Latest Version Detection

The system must be able to determine whether the user is running an outdated version compared with the latest active release.

### 1.4 Outdated Version Notification

When the installed version is older than the version required to support the currently deployed features, the user must receive a clear notification.

The notification should include:

* The currently installed version.
* The latest available version.
* The minimum required version.
* A brief summary of the missing or unsupported functionality.
* Clear instructions for updating the Extension.

### 1.5 Feature-to-Version Traceability

Each major feature or significant change must be associated with the version in which it was introduced.

This information should be maintained in a changelog or release history to prevent confusion regarding:

* Whether a feature has already been implemented.
* Which version introduced the feature.
* Why the feature may not be available or functioning in the user's current environment.

### 1.6 Compatibility Validation

Before executing a feature, the system must validate that the installed Extension version supports it.

If the feature is unsupported, the system must return a clear version-related error instead of failing silently or displaying a generic error message.

---

### 2.1 Primary Execution Layer

The Extension must be the primary application and the main execution layer for all core functionality, configurations, and modifications.

### 2.2 Complete Functional Coverage

All core features and operational actions must be fully available through the Extension.

The web interface must not provide any essential functionality that cannot also be performed through the Extension.

### 2.3 Web Interface Scope

The web interface should primarily be used for:

* Displaying data.
* Viewing statuses and results.
* Monitoring activities.
* Reviewing reports or logs.
* Providing read-only or limited management views where appropriate.

### 2.4 No Web-Only Core Features

Core features, configuration changes, or operational actions must not be implemented exclusively in the web interface.

Any functionality introduced on the web must either:

* Already exist in the Extension, or
* Be added to the Extension as part of the same release.

### 2.5 Feature Parity

The system must maintain functional parity between the Extension and the web interface.

Where the same feature is available in both interfaces, both must use the same backend logic, validation rules, permissions, and data source.

### 2.6 Extension-First Development

New functionality must be designed and implemented in the Extension first.

The web interface may subsequently expose the same functionality for presentation, monitoring, or administrative convenience, but it must not become the exclusive execution channel.

[...]

PR #40, opened 2026-07-30
Version Management (issue 32 section 1): one version, a capability ledger, and a panel that says what it is

muhammadbayoumi:
**DRAFT — do not merge.** Opened for review under the gate. CI green on both commits
(runs `30537389431`, `30537421378`).

Implements **section 1 of #32**. Section 2 is verified and left alone; the audit is below.

---

## The two incidents, and how this catches each

**Incident 1 — the crawl pace was asked for as if new.** It was built and plumbed to
`HttpFetcher` in `c63ec21` and rendered only on the display-only web page. Nothing could
answer *has this shipped, and where?*

Caught by the ledger, three ways:

- `scrapex/version.py:102` records it with the version it is guaranteed from, its commit,
  the surfaces that execute it, the settings it owns and the control it is reached by.
- The panel prints that ledger in About — `extension/app.js:283`, asserted by
  `test_about_lists_what_the_engine_deploys_and_the_version_it_arrived_in`: the answer is
  on the surface he works from, not in a commit message.
- `CHANGELOG.md` is generated from the same ledger (`scrapex/cli.py:260`), so the readable
  form cannot drift from the machine-readable one.

**Incident 2 — `crawl_parallel_sources` shipped with no field in the panel.** The engine
learned it at `63dc24b` (07-29 18:33); the panel got a field at `76a4b14` (07-30 08:23),
eleven commits and fourteen hours later. In between, two jobs queued and he asked why.

Caught twice, at two different moments:

- **At build time.** `tests/test_version.py:150` fails when a capability says the panel
  executes it and names a control neither `extension/app.html` nor `extension/app.js`
  contains; `tests/test_version.py:170` fails when a `crawl_*` setting the engine accepts
  belongs to no capability at all. Shipping the engine half alone stops being possible.
- **At run time.** An extension older than the engine gets a notice carrying all five
  facts §1.4 lists, and the missing capability is named in words — *"Crawl several
  different sites at the same time…"* (`extension/app.js:291`,
  `test_an_extension_older_than_its_engine_is_told_all_five_facts`).

---

## Which §1 clauses are implemented

| clause | where | evidence |
|---|---|---|
[...]

PR #45, opened 2026-07-30
feat(ui): a display-only time zone, one formatter, nothing written differently (#33)

### LINKED ISSUE #33: User Interface Time Zone Settings
### 6.1 Time Zone Selection

The system must provide a setting that allows users to select their preferred time zone for displaying dates and times across the user interface.

### 6.2 Display-Only Conversion

The selected time zone must affect only how dates and times are displayed in the Extension and web interface.

It must not modify:

* Stored timestamps.
* Original source data.
* Database records.
* Server-side timestamps.
* API payload values.
* Audit or system logs at the data level.

### 6.3 Standard Time Storage

All timestamps must continue to be stored and exchanged using a standard time reference, preferably UTC.

The selected user-interface time zone must be applied only during the presentation layer conversion.

### 6.4 Supported Time Zones

The time zone selector must use standard IANA time zone identifiers, such as:

* `Asia/Riyadh`
* `Europe/London`
* `America/New_York`
* `Asia/Dubai`

Fixed UTC offsets should not be used as the primary configuration method because they may not correctly account for daylight saving time changes.

### 6.5 Default Time Zone

The system should automatically detect the user's local time zone as the default value.

The user must be able to override the detected time zone manually from the Settings page.

### 6.6 Persistent Preference

The selected time zone must be saved as part of the user's preferences and remain active across:

* New sessions.
* Browser restarts.
* Extension restarts.
* Supported devices associated with the same account.

### 6.7 Consistent Display

The selected time zone must be applied consistently to all user-facing date and time values, including:

* Activity timestamps.
* Status updates.
* Reports.
* Notifications.
* Scheduled actions.
* History records.
* Dashboard data.
* Extension views.
* Web interface views.

### 6.8 Time Zone Visibility

Where time interpretation may be important, the interface should clearly display the active time zone or its abbreviation next to the date and time.

Example:

`30 July 2026, 11:05 AM — Asia/Riyadh`

### 6.9 Extension and Web Synchronization

The Extension and web interface must use the same saved time zone preference.

Changing the time zone in either interface must update the user's preference and be reflected in the other interface.

### 6.10 Immediate UI Refresh

When the user changes the selected time zone, all visible date and time values should be refreshed immediately without requiring the underlying data to be reloaded or modified.

### 6.11 Fallback Behavior

If the selected time zone is unavailable or invalid, the system must fall back to:

1. The user's detected local time zone.
2. The system default time zone.
3. UTC, if neither option is available.

The system must not modify the stored timestamps during fallback.

### 6.12 Audit and Diagnostic Context

System and audit logs must retain their original standardized timestamps. The interface may display a converted value, but the original UTC timestamp should remain available for diagnostic and auditing purposes.

---

## 7. Time Zone Acceptance Criteria

The time zone functionality will be considered successfully implemented when:

1. The user can select a time zone from the Settings page.
2. The selected time zone affects only the user-interface presentation.
3. Stored timestamps and source data remain unchanged.
4. All user-facing timestamps are converted consistently.
5. The selected preference persists across sessions and supported devices.
6. The Extension and web interface display the same converted time values.
7. Changing the time zone refreshes displayed times immediately.
8. Daylight saving time is handled automatically through IANA time zone rules.
9. Invalid time zone configurations fall back safely without modifying any data.
10. Original UTC timestamps remain available for auditing and troubleshooting.

muhammadbayoumi:
Closes #33.

**DRAFT until the main session reviews it.** Do not merge.

A display-only time zone preference. One formatter, one shared preference, and
not one stored value changed.

---

## The §6.9 question — resolved by the owner, 2026-07-30

Issue #32 §2.3/§2.4 says the web interface must not provide functionality that
changes state, and `tests/test_settings_live_in_the_extension.py:33` fails the
build if that page grows a control. Issue #33 §6.9 says changing the zone in
*either* interface must update the preference. A selector on the web page would
have satisfied one and broken the other.

**The owner ruled: the selector lives in the extension only; the web page
displays the active zone and never offers to change it.** §6.9's intent — one
preference, never two conflicting ones — is met by a *single shared preference*
rather than by two selectors. §2.3 explicitly allows display.

That test was not weakened. It is now guarded from the other side too:
[`test_the_web_page_shows_the_active_zone_and_offers_no_control`](tests/test_display_time_zone.py)
fails if the page stops *showing* the zone **or** if it ever grows a
`data-time-zone-select`. Please do not reopen this — it is settled.

---

## What was already built, verified rather than reinvented

**§6.2/§6.3 need no new enforcement.** Confirmed with a run, not by reading:

```
$ python -m pytest tests/test_payload_compat.py -q -k "timezone or invalid_fixtures"
5 passed
```

`scrapex/payload.py:241` refuses any `scraped_at` whose offset is not UTC, and
`contracts/fixtures/payload_invalid_timezone.json` is that test's vector. So the
storage contract was already correct; this PR's only obligation was not to break
it, and [`test_the_payload_contract_still_refuses_a_non_utc_timestamp`](tests/test_display_time_zone.py)
now keeps that obligation mechanical.

**§6.9's sharing mechanism already existed**, so it was followed rather than
replaced: `ui_time_zone` sits beside `ui_appearance` in `scrapex/settings.py:83`,
with `GET`/`POST /api/timezone` mirroring `/api/appearance`
(`scrapex/webui/app.py:1731`). One settings row, both surfaces, last-write-wins
on `updatedAt`. No second sharing scheme.

**§6.4's Python validation reuses `scheduler.zone_exists`** (`scrapex/scheduler.py:39`)
rather than a second `zoneinfo` check — same tz database, so a zone that can be
*scheduled* can be *displayed* and the two can never disagree.

**No migration.** A preference belongs in settings; 0056/0057 untouched.

---

## Section 7 acceptance criteria

| # | Criterion | Verdict |
|---|---|---|
| 1 | Zone selectable from Settings | **Met** — `extension/app.html:916` (Display group, `#ui_time_zone`). `test_the_time_zone_can_be_chosen_from_the_panels_settings_tab` |
| 2 | Affects presentation only | **Met** — see criterion 3's evidence |
| 3 | Stored timestamps and source data unchanged | **Met** — `test_choosing_a_time_zone_changes_exactly_one_row_in_the_database` diffs *every table* before/after; `test_no_writer_can_read_the_display_preference` |
| 4 | All user-facing timestamps converted consistently | **Met** — 22 template sites + 5 client sites, one formatter; `test_no_surface_grows_a_second_date_formatter`, `test_every_template_that_renders_an_instant_uses_the_macro` |
| 5 | Preference persists across sessions and devices | **Met** — localStorage + the engine-held settings row; `test_a_zone_saved_on_the_other_surface_arrives_here` |
| 6 | Extension and web show the same converted value | **Met** — one preference, one byte-identical module; `test_the_two_copies_of_the_module_are_identical`, `test_choosing_a_zone_shares_it_with_the_engine_for_the_web_page` |
| 7 | Changing the zone refreshes immediately | **Met** — `test_changing_the_zone_re_renders_visible_times_without_refetching` asserts **no data route is called at all** |
| 8 | DST handled through IANA rules | **Met** — `Intl.supportedValuesOf('timeZone')` for the list, `Intl.DateTimeFormat` for the conversion. No fixed offset is offered as a setting |
| 9 | Invalid configuration falls back safely, no data modified | **Met** — `test_an_invalid_zone_falls_back_down_the_chain_and_says_which_step`; `test_an_unknown_identifier_is_refused_rather_than_stored` |
| 10 | Original UTC remains available | **Met** — `title="Stored as …Z (UTC)"` + `data-utc` + `<time datetime>` on every converted value; `test_the_raw_utc_stays_reachable_on_every_converted_time` |

Nothing is **not met**, and nothing is blocked.

---

## How the DRY requirement (§6.7) is actually held

One formatter in `extension/timezone.js` — copied byte-for-byte to
`scrapex/webui/static/timezone.js` for the same origin reason `appearance.js` has
two copies, with a test that fails if they drift. Three thin adapters, because
this codebase emits a time in exactly three ways:

- **Jinja templates** → `{{ stamp(value) }}` (`templates/_time.html`), which writes
  the raw UTC into `data-utc` and lets the module convert it in place;
- **HTML strings** → `ScrapeXTime.markup(value, mode, {zone})`;
- **DOM nodes** → `ScrapeXTime.node(value, mode, {zone})`.

All three land on the same `format()` and all three leave the raw instant in the
element. That is what makes §6.10 a re-read of the DOM (`paint()` — no fetch, no
data touched) and §6.12 free.

Two static guards keep it that way. `test_no_surface_grows_a_second_date_formatter`
refuses `toLocaleDateString`, `toLocaleTimeString`, `toLocaleString([], …)`,
`Intl.DateTimeFormat().format`, and the hand-sliced-instant idiom
(`something_at … .slice(0, 10)`) anywhere in the UI scripts. Both patterns were
really in the tree before this PR — the list is a record of what was fixed, not
a guess. `test_every_template_that_renders_an_instant_uses_the_macro` does the
same for Jinja, so a page added next month gets caught by CI rather than by a
reader noticing its times are four hours out.

### The sites that were converted

`overview.html` (3), `jobs.html` (2), `history.html` (2), `logs.html`,
`changes.html` (1 server + 2 in its inline script), `offer.html` (4),
`schedules.html` (1 server + 1 inline), `excel.html`, `sync.html` (2),
`_retention.html`, `_storage.html` (2), `extension/app.js` (3),
`static/grid.js` (3).

### Deliberate exclusions, each with a reason

- **`business_date` / `price_changed_on` / `last_confirmed_on`** — calendar
  dates, not instants. `price_changed_on` is `po.business_date` and
  `last_confirmed_on` is already `[:10]` server-side (`scrapex/reports.py:348`).
  Shifting a calendar date into a zone either does nothing or moves it a day,
  which would be a false fact about the market. The formatter refuses any value
  that is not an instant and returns it verbatim
  (`test_a_business_date_is_never_shifted_by_the_display_zone`).
- **`usd_rate_as_of`** (`grid.js:1391`) — the date *of an exchange rate*. Moving
  it could misstate which day's rate was used.
- **The export filename stamp** (`grid.js:3080`) — a written value, not a display.
- **A schedule's own `timezone` field, and the `run_at` wall clock** — these
  decide *when a job fires*. Converting them would be a behaviour change, not a
  display change, and is outside this issue. `Intl…resolvedOptions().timeZone`
  remains their default at `app.js:1252` and `schedules.html:221`, untouched.
- **The panel's log copy-to-clipboard** (`app.js`) — a diagnostic export; §6.12
  wants UTC there.

Per-row zone labels would be noise on 200 log lines, so dense tables carry one
`time_zone_note()` line instead (`logs.html`, `history.html`) while single
prominent values name their zone inline (`zone=true`).

---

## §6.11 — the fallback chain

`selected → detected → system → UTC`, reported rather than silent.
`resolution()` returns the zone, the **step**, and a sentence that names both the
rejected identifier and where it landed; that sentence is what
`[data-time-zone-note]` shows on both surfaces.

Two properties worth checking in review:

1. **A fallback never rewrites the stored preference.** An unresolvable
   `Mars/Phobos` stays saved exactly as typed, so fixing the tz data restores the
   owner's choice instead of finding it overwritten.
2. **`""` means "follow detected"** and is a *storable decision*, not an absence.
   `test_the_empty_zone_is_valid_and_means_follow_the_detected_one` explains why:
   if resetting to Detected deleted the row, the other surface would keep its
   older explicit zone, nothing would contradict it, and it would push it back.

The `system` step is the machine's own UTC offset (`Etc/GMT±N`) for a browser
that names no zone. §6.4 rules fixed offsets out as the *primary* configuration
— which is exactly why this is the last step before giving up and not an option
in the selector.

---

## Bugs found and fixed while testing

- `markup()` set `data-utc-zone="show"` but built its initial text with
  `format()` instead of `label()` — so a value shipped without its zone and grew
  one the first time anything re-painted it. The same element said two different
  things depending on when you looked.
- The panel harness writes every page into one `file://` origin, so Chromium
  shares **one** localStorage across them. A zone chosen by one test had become
  the starting state of the next, which made the §6.10 test pass for the wrong
  reason (it read the previous test's Riyadh as its own "before"). Both harnesses
  now clear the key before the module reads it.
- `test_vendor.py`'s innerHTML guard correctly caught the zone `<select>` being
  built from a string; it builds DOM now. The same reasoning was then applied to
  the schedules row rather than introducing an `innerHTML` on a page that renders
  scraped values.

---

## Rules observed

- **`PAYLOAD_VERSION` / `PAYLOAD_COMPAT_VERSION`: untouched.** `git diff main -- scrapex/payload.py contracts/ apps_script/` is empty. No payload field changes, so nothing here adds a second reason for an Apps Script paste.
- **No migration.** The only change outside `scrapex/webui/` is one `Setting` line.
- **The live warehouse was never opened.** Every test builds its own `tmp_path` database; the engine was not restarted and no crawl was started.
- **English UI text**; IANA identifiers are not translated.
- Named files staged, `git diff --cached` read in full. No `git add .`.

## Verification

Rebased onto `main` at 99264bd (which landed while this was in progress). The one
conflict was in `grid.js`, where that branch changed `money(…, offer.unit)` to
`money(…, basisOf(offer))` on a line whose date cells this PR also changes — both
sides kept.

```
python -m pytest        # 1686 passed, 1 xfailed
node contract/parity/parity.test.mjs       # 3/3
```

---

## Follow-up: the owner's last-crawl report, and what it exposed

The owner reported that the last-crawl date under the source table did not follow
a unified format. It did not, in **four** places, and finding out why mattered
more than the fix.

**Why the guard was blind.** `test_every_template_that_renders_an_instant_uses_the_macro`
keyed on the *name* of the field — anything ending in `_at`. The field is
`summary.last_run`. A vocabulary of field names is always one product decision
behind, so the guards now key on the **manipulation** instead: slicing to 10/16/19
characters, or swapping the ISO `T` for a space. That shape is what
"hand-format an instant" *is*, and it cannot be renamed out of view. Measured, not
assumed: the template rule fires **4/4** on the reported miss at `source.html@main`
and **zero** times on the current tree; the JS rule needed exactly one named
exemption — the CSV/JSON export **filename** stamp in `grid.js`, which must stay
UTC or it would disagree with the UTC inside the file it names. Both name-based
and shape-based rules are kept, because they catch different failures, and the
residual blind spot they share is stated in the code: a bare `{{ summary.last_run }}`
with no manipulation at all is invisible to both and still needs a reader.

**A five-angle sweep then found three more** (idiom / semantics / server-field /
page-read / client-file, every candidate adversarially verified):

| Site | What it was |
|---|---|
| [`_source_list.html:46`](scrapex/webui/templates/_source_list.html:46) | The *other half* of the same freshness line — `Last crawled … UTC` in the dataset picker that the source page itself includes. The owner was reading two different renderings of one fact on one page. |
| [`extension/app.js` `freshnessLine()`](extension/app.js:1434) | The panel's copy of that line, **landed on `main` today**. The new shape guard caught it within a minute of being written. |
| [`jobs.html:111`](scrapex/webui/templates/jobs.html:111) and [`logs.html:63`](scrapex/webui/templates/logs.html:63) | The two that actually mattered — see below. |

### The polls were worse than the original defect

Both pages server-render a time through the macro and then **overwrite it on a 4s
poll** with the raw instant. The page is correct until the first tick and silently
wrong afterwards. And because `textContent` *deletes* the `<time data-utc>`
element the macro rendered, `ScrapeXTime.paint()` has nothing left to select — so
a later zone change can never repair it. There is no `MutationObserver` anywhere
in the tree; I checked.

Two aggravating details, both verified in the source: `jobs.html` paints **every**
`[data-job-ref]` row (the early return only decides whether polling *starts*), so
one live job did this to every finished card on the page; and `logs.html` calls
`paint()` *before* it consults `TERMINAL`, so even a finished job's Time column
was clobbered once on load. Both now build the cell with `ScrapeXTime.node()`
via `replaceChildren`/`append`, and
`test_no_template_script_writes_a_raw_instant_into_the_page` fails the build if
either goes back.

### Four look-alikes that must never be "fixed"

Each now says so where it lives, because each sits beside a value that *is*
converted: `business_date` in `offer.html` and `grid.js`, and `usd_rate_as_of`.
The most dangerous is `retention.prune_before` — the **same string** is posted
back as the prune boundary, so converting the display would put a different date
on screen from the one the button acts on, and prune rows the owner never saw
listed. That is a display feature silently becoming a data change.

### One question for the owner, not answered here

`scrapex/reports.py` truncates instants to dates with `[:10]` **on the server**
(`last_confirmed_on`, `last_seen_at`, `usd_rate_as_of`). Those columns are
correctly not client-side conversion sites, but their day boundary is UTC's, not
the display zone's — the same off-by-one-day error §6.7 forbids, relocated to the
server. It is a server-side question with a real trade-off, so per the standing
rule it is put to him as a decision rather than changed quietly.

### Re-verified

Rebased again onto `main` at 322e205 (16 further commits landed during the work;
conflicts in `app.html` and `panel_harness.py` were additive on both sides and
both sides kept).

```
python -m pytest        # 1745 passed, 1 xfailed
node contract/parity/parity.test.mjs       # 3/3
```

One pre-existing test was updated rather than worked around:
`test_the_data_tab_states_when_each_source_was_last_crawled` asserted the old
hand-formatted `"Last crawled 2026-07-29 08:30"` — the exact shape the owner
reported. It now pins the zone (so it asserts the product's format, not the CI
machine's zone) and additionally checks the raw UTC is still on the element.


muhammadbayoumi:
Brought `main` in (the branch was behind #40, #66, #68, #72, #76) and resolved. Reviewed on the merged result.

## Five conflicts

Four were mechanical — both sides adding a new block at the same offset: two functions in `scrapex/webui/app.py`, two blocks in `extension/app.js`, new tests in `tests/test_panel_dom.py`, and `tools/panel_harness.py`, whose signature is now the union of **three** sessions' parameters (`logs`, the four version ones, and `timezone`/`schedules`).

**The fifth was a real design conflict**, and both sides were right. #68 builds the dataset-list caption once and uses that one string for both the visible text and the `title`, because the line is clamped to one row and a `title` that is a prefix recovers nothing. #45 routes the time through `stamp()` so it reads in the owner's zone. But `stamp()` emits a `<time>` **element**, which cannot go inside a `title=` attribute.

Resolved so both rules hold: the element **asks** to be titled (`data-clamped-title`) and `timezone.js` fills it from the rendered text **after** converting. Seven lines, applied to both copies, which remain byte-identical — checked by this PR's own `test_the_two_copies_of_the_module_are_identical`.

## This PR's guard caught four places in work that landed after it was written

`test_every_template_that_renders_an_instant_uses_the_macro` named them precisely and described the fix: the rates dataset card in `_source_list.html`, two instants in `google_finance_dataset.html`, and one in `settings.html` — all from #72, all hand-formatting `[:16]|replace("T"," ") ~ " UTC"`. All four now go through `stamp()`.

## One correction to my earlier report

I first saw ~80 failures here and guessed playwright contention. **That was wrong, and the cause was my own resolution.** The conflict boundary fell *inside* `refreshGoogleFinance`, so stripping the markers and keeping both sides joined its first half to the timezone section and lost its two closing braces. `extension/app.js` did not parse, so the whole panel bundle failed and every DOM test with it — the same shape as the `engine.js` import incident.

Worth recording: `node --check extension/app.js` **passes on a broken file**. Node treats a lone `.js` containing `import` as ambiguous and does not parse it. `tests/test_panel_wiring.py` catches it because it pipes through `node --input-type=module --check`, and its docstring says exactly why.

## Verification, run alone

Full suite: **exit 0, zero FAILED.** `extension/app.js` parses as a module; panel wiring 17/17; contract parity exit 0.

The standing rule holds: **zero** controls added to any web template — no zone selector on the page, exactly as ruled. The selector is in the panel; the page displays the active zone.

PR #63, opened 2026-08-01
feat(data): identity, then the offer, then the filing

muhammadbayoumi:
The owner: his eye hunts for the price in a different place in every table. **Measured**, before this — the position of Price in each source's own table:

| MADAR | SIKAEGSHOP | ALSWEED | ELBUROJ | GPP_ENERGY |
|---|---|---|---|---|
| **19** | 10 | 8 | 8 | 6 |

## What I expected to find, and did not

I expected the **order** to differ between sources. It does not. Comparing every adjacent pair against the central list: **zero out-of-order pairs, in any source.**

An earlier check of mine said otherwise and was simply wrong — it filtered the reference list by membership in the shorter one, so a source showing a column MADAR does not show read as a reordering. Task #72 asks to "unify the column order"; **there was nothing to unify.**

## The real cause

Price sat at index **33** of `BROWSE_COLUMNS`, behind **twenty-four classification columns**, and each source is shown only the columns it actually fills. So the number of levels a source publishes decides how far right its price drifts.

**The order was consistent and the position was not** — and the position is what the eye tracks.

## The fix

Three named blocks instead of one flat list: **identity → the offer → the filing.**

| | after |
|---|---|
| MADAR | **6** |
| SIKAEGSHOP | 4 |
| ALSWEED · ELBUROJ · GPP_ENERGY | 3 |

A spread of **three** rather than thirteen, and the remainder is honest: a source with no variants is not shown a variant column.

**Nothing is pinned and no column is forced to appear.** The per-source gates still decide *presence* — the ruling that keeps a fuel site's USD estimate off a shop's table — and it is untouched. Only the order changed.

## Tests

- every classification column comes **after** Price, and identity before it — the **rule**, not the current arrangement, so a future insertion cannot quietly undo it
- the reorder lost no column and duplicated none, and every browse column is still exportable — the grid and the spreadsheet read this same list, and hand-reordering a list is exactly how a column vanishes from both

**1706 passed, 1 xfailed.** Contract parity **3/3**.

### DIFF
--- a/scrapex/reports.py
+++ b/scrapex/reports.py
@@ -605,84 +605,80 @@ def _level_columns() -> list[tuple[str, str]]:
 
 
 BROWSE_COLUMNS: list[tuple[str, str]] = [
-    # The two languages of one field sit TOGETHER, English first. They used to
-    # be filed in separate blocks — every English column, then Country, then
-    # every Arabic one — so a bilingual shop showed Record (AR), Record,
-    # Category, Country, Category (AR): the same fact twice with an unrelated
-    # column wedged between. The label for each pair is written once, by
-    # BILINGUAL_COLUMNS below. The "(AR)" is AUTHORED into the label, not
-    # appended at runtime: the mark is a property of the column, not of the
-    # pair, so an Arabic-only source is still told which language it is
-    # reading. (It also makes /api/fields and the grid agree by construction
-    # rather than by both remembering to run the same loop.)
-    ("product_name", "Product name"),
-    ("product_name_ar", "Product name (AR)"),
-    ("country_code_alpha2", "Country code"),
-    ("brand", "Brand"),
-    ("brand_ar", "Brand (AR)"),
-    # Classification (owner ruling 2026-07-22): part of the MAIN table, with
-    # every level the source publishes. "category" carries the source's full
-    # path ("Cables > Low voltage") or, for a source that only files products
-    # under flat labels, the labels themselves. The per-level columns split
-    # the path so the table can sort and group by any layer; presence gating
-    # keeps each level to the sources that actually reach that depth.
-    ("category", "Category"),
-    ("category_ar", "Category (AR)"),
-    # The DEEPEST level this row actually reaches, in one column. The owner's
-    # ask (2026-07-28): «عاوز عمود يكون مجمع اخر تصنيف لكل صف» — sources like
-    # MADAR and ADVANCEDCASTLE classify to different depths row by row, so no
-    # single Category L-column holds "the most specific thing this is". Reading
-    # it off the row means never asking which level to look in.
-    ("category_leaf", "Category leaf"),
-    ("category_leaf_ar", "Category leaf (AR)"),
-    *_level_columns(),
-    ("variant", "Variant"),
-    ("variant_ar", "Variant (AR)"),
-    ("sku", "SKU"),
-    ("price", "Price"),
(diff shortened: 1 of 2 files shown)


muhammadbayoumi:
Reviewed on the merged result against today's main. **The reorder itself is right** — same 50 keys, no label changed, per-source presence gates including the USD-estimate leak guard provably unchanged, and `grid.js` takes column order from the server payload, so no column disappears and the JS/Python contract is untouched. Three things need doing before it lands.

**1 — it orphans a generator main introduced on purpose.** `d9e6f7a` replaced hand-typed `category_l1..l4` tuples with `*_level_columns()` and wrote down why; this types twenty of them back in and leaves the generator dead. Measured, raising `CATEGORY_LEVELS` from 10 to 12:

| | 10 | 12 |
|---|---|---|
| main | 22 category columns | **26** |
| this PR | 22 | **22** |

The constant is inert, and `reports.py:593` still promises "raising it further is one line". `*_level_columns(),` placed inside the filing block yields the identical list — the reorder never required unrolling it.

**2 — only one of three consumers moves.** The Choose-Columns panel and the Current-View export read `dataset_field.display_order`, and `ensure_fields` is documented never to rewrite an existing row (`scrapex/fields.py:21-44`, "an existing field is left completely alone"). So on any warehouse whose sources have been opened once — the owner's included — the grid reorders and the column chooser and the export keep the OLD order. The same code then gives two different answers depending on whether a source has ever been opened. "The price sits in the same place in every table" is met in one surface out of three.

**3 — two written rulings still say the opposite.** `scrapex/reports.py:582-585`, eighteen lines above the list, and `docs/data-page-schema.md`, which calls itself "the ruling". Neither is touched, so the code now contradicts them.

Two smaller ones worth folding in: `country_code_alpha2` moves from index 2 to 13, putting Country to the **right** of the price on commodity sources where `_tree_shape` already treats it as half the row's identity; and ~55 lines of per-column rationale are deleted that `COLUMN_NOTES` does not carry.

One thing to know: `85ec935` — this branch's only commit — also rode inside PR #67, which is now closed. This PR is the sole home for the work.

muhammadbayoumi:
Closing: all three requirements this was sent back with have since landed, by other routes. The idea was right and it is now in main — this branch is not the way it got there.

**1 — the generator stayed load-bearing.** This PR re-typed twenty `category_lN` tuples by hand and orphaned `_level_columns()`; measured, raising `CATEGORY_LEVELS` from 10 to 12 gave 22 → 26 columns on main and **22 → 22** on this branch. #85 sorts the existing list by block instead, so the generator is still what builds the levels and `reports.py`'s promise that raising the ceiling is one line is true again — asserted by `test_the_category_levels_still_come_from_the_generator`.

**2 — all three surfaces move.** This was the blocker, and it was worse than *only the grid moves*: the grid was the one surface that **could not** move, because it read the literal list and never consulted `dataset_field` at all. #85 gives the grid, the chooser and the export one reader — `fields.column_order` — and 0059 adds `arranged_at` so an order you arranged is never overwritten by an improved default.

**3 — the ruling was updated.** #88 generates `docs/data-page-schema.md` from the same lists, with a golden test. It had drifted in five ways, including still stating the order this PR was replacing.

What this branch actually achieved is on main: **price moved from index 32 to 7**, identity then the offer then the filing, exactly as agreed.

Reopen if you want anything from the diff that #85 did not carry — the reordering intent was correct and only the mechanism differed.

PR #66, opened 2026-08-01
feat(sikaegshop): the stock count the price row never carried

muhammadbayoumi:
SIKAEGSHOP publishes a remaining-stock count for every product. The connector
read that number **twice** and put it nowhere a query can reach:

- `_availability` reduced it to the word `in_stock` / `out_of_stock`
- `enrichment_rows` filed it in the attribute bag, as prose keyed by a name

The PRICE row — the one column a reader can filter, chart and compare against
another shop — was never given it.

## Measured on the live warehouse (read-only, 2026-07-30)

| source | observations | carrying a count | offer_state rows | carrying a count |
|---|---|---|---|---|
| **SIKAEGSHOP** | **252** | **0** | **87** | **0** |
| MADAR | 6,146 | 0 | 3,425 | 297 |
| MASDAR | 617 | **617** | 617 | **617** |

MASDAR is the control: the column works end to end, because `hybris.py` has
passed `stock_quantity=` since it was written.

## No migration, no `PAYLOAD_VERSION` bump

Nothing is being added. `stock_quantity` is already in the row contract
(`scrapex/rowspec.py:83`), already in the 34-column header, and ingest already
stores it (`scrapex/ingest.py:685, :760, :1229, :1267`). The connector simply
never filled it in — the fix is one keyword argument and the reasons for its
shape.

## The rule: store what the shop says, never invent a zero

A live census of all 87 products (captured for this PR, 8 pages, read-only GETs
on the same open endpoint the connector already crawls):

- **87/87** publish the `stock_quantity` key, all non-null, `int` on all 87
- **16/87 publish `0` — and all 16 carry `is_active: true`**
- range 0 … 4000

Those 16 are the whole argument. The shop lists them and says none are left at
the same time, so a falsy guard would silently discard the most
decision-changing thing it states, on 18% of the catalogue. The inverse error is
worse: defaulting an absent count to `0` would announce "sold out" about a
product the shop never discussed. Hence an explicit `is not None`, never
truthiness — and both halves get a test.

`_stock_count` is extracted so `_availability` and the price row read the same
number through the same fallback chain; deriving it twice is exactly how one of
them could answer `in_stock` from a figure the other never wrote down. `bool` is
excluded because `isinstance(True, int)` is True and this family's sibling keys
are flags.

The enrichment row **keeps its opposite, falsy guard**, deliberately: an
attribute bag is prose, where an absent row and a `0` row read the same to a
human, and letting the 0 through there would also start printing
`Maximum stock level: 0` on 85 of 87 products. A price column is a measurement,
where NULL and 0 are different claims. Its comment now says so.

## Why every warehouse test crawls twice

A single ingest into an empty database **cannot fail**. The first observation of
an offer is always appended — there is no open period to compare against — so
the count lands whatever the append gate does, and the test is green while the
value is dropped on every crawl after the first. Each warehouse test here
ingests twice and asserts the row written by the second, following the whole
trip: row → payload → ingest → warehouse, with the payload asserted in the
middle so a downstream failure can never be blamed on transport.

## ⚠️ Merge #42 first — 3 of the 8 tests wait on it

**#42 was still open when this was measured.** Verified by applying its
`ingest.py` diff to this tree and removing it again:

| state | this file | full suite |
|---|---|---|
| with #42 | **8/8 pass** | **green** |
| without #42 | 5/8 pass | only those 3 fail |

Without #42, crawl 2 reports `confirmed=1 / observations=0`: the figure reaches
`offer_state` and nothing else. That is the bug seen from the other end —
`offer_state` is a cache rebuilt from the latest observation, so a count landing
only there does not survive a rebuild. `price_observation` is the durable home,
which is why the assertions anchor on it and read `offer_state` second.

**Nothing else in the suite moves either way**, so the connector change is
compatible with `main` on its own; it is these three assertions, and only these,
that are red until #42 lands.


muhammadbayoumi:
Brought `main` in — **merged clean, no conflicts**; main never touched `custom_json.py`. Full suite on the merged result: **exit 0, zero FAILED**. The three red tests were entirely the stale base: this branch was written against main's `8fe16e6` ingest-gate fix and sat 16 commits behind it.

Reviewed and **not redundant** with `8fe16e6`. That commit fixed the INGEST gate (`_still_the_same_price`) so a moved stock figure forces an observation append, and its own message says of MADAR "the connector was not the gap". For SIKAEGSHOP the connector **is** the gap, and main has never closed it — no `stock_quantity=` on the price row anywhere in `custom_json.py`. These are complementary halves.

**One condition, not a blocker.** `stock_quantity` is invisible to a reader: zero occurrences in `scrapex/reports.py`, `scrapex/webui/static/grid.js` and `extension/`. After this merges, a re-crawl stores the number and no column, export or panel shows it — the same shape as the accident that stranded PR #51. Capturing the value is still strictly right (an uncaptured observation is lost forever, a display can be added any time), so this merges; the display half needs its own issue.

PR #68, opened 2026-08-01
Stop the freshness line from crushing the dataset name

muhammadbayoumi:
Owner report (item 2 of the UI defect list): «تشوه المحتوى فى قائمة dataset بعد اضافة تاريخ اخر جولة» — the dataset list was mangled after the last-crawl date was added. In the screenshot `madar`, `masdar`, `samehgabriel` and `sika` all render as four characters and an ellipsis, with the `[Row N]` badge overlapping the text.

## Cause

The caption went into `grid-column: 2` — the same narrow `auto`-width column the badge lives in. `Last crawled 2026-07-30 10:15 UTC · 19,548 rows seen` is much wider than `Row 19,548`, so that column grew to fit the caption and took the width from column 1, where the name is. The name was never at fault; it had nowhere left to go.

## Fix

The caption describes the dataset, not the badge, so it spans the card on its own row — `grid-column:1/-1; grid-row:2` — with the identity and the badge pinned to `grid-row:1` so it cannot be auto-placed back beside them. It stays one line (`min-width:0` so the grid item will shrink below its content, then ellipsis) and gains a `title` so the full value is still readable when cut: the card decides **where** a fact is shown, never **what** it says.

Asset version `source-ui-8` → `source-ui-9` — a cached stylesheet would keep the old placement, which is the entire defect.

## Verification

Full suite green; contract parity 3/3. Each of the three new guards was proved to bite by re-breaking the fix:

| re-break | result |
|---|---|
| caption back in `grid-column:2` | `test_..._does_not_crush_the_name` fails |
| `min-width:0` dropped | `test_..._stays_readable_when_truncated` fails |
| `title` removed | `test_..._stays_readable_when_truncated` fails |

### DIFF
--- a/scrapex/webui/static/pages/data-workspace.css
+++ b/scrapex/webui/static/pages/data-workspace.css
@@ -50,15 +50,24 @@
   box-shadow:inset 3px 0 var(--accent)}
 .dataset-choice[hidden]{display:none}
 .dataset-choice>.source-identity{grid-column:1;grid-row:1;min-width:0}
-.dataset-choice>.source-identity-meta{grid-column:2;grid-row:1;align-self:end;font-size:.58rem;white-space:nowrap}
+/* Kept as it was. This selector matches NOTHING: source_metric nests
+   .source-identity-meta inside .source-identity-footer, so the [Row N] badge is
+   never a direct child of the card and never a grid item. Verified by rendering
+   a card: its ancestors are source-identity > source-identity-footer. Left in
+   place because removing pre-existing dead CSS is not this fix's business, but
+   named here so nobody reasons from it again — including this file's own
+   previous comment, which did. */
+.dataset-choice>.source-identity-meta{grid-column:2;align-self:end;font-size:.58rem;white-space:nowrap}
 /* THE FRESHNESS LINE IS A CAPTION ON THE CARD, NOT A FIGURE BESIDE IT.
-   It was placed in grid-column 2 — the narrow auto-width column the [Row N]
-   badge lives in — so "Last crawled 2026-07-30 10:15 UTC · 19,548 rows seen"
-   widened that column until the NAME had nothing left: madar, masdar,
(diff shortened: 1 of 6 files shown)


muhammadbayoumi:
## Reviewed my own PR against latest main and found three real defects in it

All three sat where no test looked, which is why the suite was green throughout. Fixed in `b110e7d`.

**1 — the title was a strict prefix (blocking).** It stopped at `UTC` and omitted `· 19,548 rows seen`: precisely the tail that the `text-overflow:ellipsis` added by this same PR removes. It recovered only the part you could already read, and on main that figure had been legible because the caption wrapped. The line is now built once and used for both the visible text and the title, so they cannot drift.

**2 — a guard protected a rule that styles nothing (blocking).** The test asserted `grid-row:1` on `.dataset-choice > .source-identity-meta`. That selector matches no element: `source_metric` nests `.source-identity-meta` inside `.source-identity-footer`. Verified by rendering a real card — ancestors are `source-identity > source-identity-footer`. The assertion is gone, the `grid-row:1` I had added to that rule is withdrawn, and the rule is left otherwise exactly as main had it; removing pre-existing dead CSS is not this fix's business.

**3 — the comment named the wrong cause.** A comment is the permanent record, and mine blamed the `[Row N]` badge for occupying column 2. Nothing occupies column 2. The card is `grid-template-columns:minmax(0,1fr) auto`; the freshness line was the only item in the `auto` track, so it set that track's width and `minmax(0,1fr)` let column 1 pay for it. The badge is **inside** column 1 — it appeared to collide because the same track was squeezing it.

Also: the clamp applies to whichever caption renders, so `Has data, but no crawl has succeeded yet` and `Never run` are titled too. The rule belongs to the clamp, not to how long a string happens to be today.

**Verification.** Full suite exit 0, zero FAILED; parity exit 0. Both new guards proved to bite — a title reduced to its prefix fails the first, a caption shipped without one fails the second.

PR #67, opened 2026-08-01
fix(panel): Google Finance reaches the panel it was already advertised in

muhammadbayoumi:
The owner opened Settings looking for **Google Finance** and found nothing. The web page told him where to go — *"Controls live in the ScrapeX side panel under Settings → Google Finance"* — and **sent him to an empty place**.

## What happened

PR #51 was merged. Its server half, its web read-out and its tests are all on `main`. **Its extension half is not:** `extension/app.html` on `origin/main` contains **zero** mentions of Google Finance, and the two `google` hits in `app.js` are the Sheets navigation entry, unrelated. **50 lines of markup and 70 of wiring went missing** between the branch and main.

I do not know which step dropped them — the rebase merge reported nothing and **I took silence for success**. What I do know is that the branch commit still has them, and this restores exactly those two files from it.

## Verified not a revert

Taking whole files from an older commit is how a day's work disappears, so everything merged since was checked **by name** and is still present:

- the schema-lag banner
- `crawl_parallel_sources`
- the shared split button
- the crawl-pace section
- the resume wiring

The only deletions are five cosmetic lines the renumbering re-inserts.

## And the test caught what I missed

The restore also brought back four `<small>` descriptions that **#47 had deliberately removed** from the split-button menu, and `test_the_log_controls_are_one_shared_split_button` failed on the exact sentence.

That is the difference between a test that works and a test that decorates — and it was written today by the session that removed them.

Settings sections renumber to 01–07 with no duplicate index.

**1706 passed, 1 xfailed.** Contract parity **3/3**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #72, opened 2026-08-01
Google Finance controls and dataset — the feature PR #51 merged into a branch that had died five seconds earlier

muhammadbayoumi:
**This PR re-targets work that is already written, already tested, and completely absent from `main`.** Nothing here is new code.

## What happened

| time | event |
|---|---|
| `2026-08-01T05:03:54Z` | **PR #48** `codex/fix-idle-ui-cpu` → **main**, squash-merged as `0c1571a`. The squash carried four files: `jobs.py`, `rates.py`, `tests/test_idle_cost.py`, `tests/test_rates.py`. |
| `2026-08-01T05:03:59Z` | **PR #51** `codex/google-finance-control` → **`codex/fix-idle-ui-cpu`** — *not* main. Merged into a branch that had been consumed five seconds earlier. |

GitHub reports #51 as **MERGED**, truthfully: it was merged into its stated base. That base had already been squashed into main and no longer fed anything. So the sixteen files and 934 insertions went into a dead end, and `main` has never seen them.

Proof `main` has none of it: `GET /api/rates/google-finance` → 404, `GET /data/google-finance` → 404, `POST /api/rates/google-finance/refresh` → 404, `POST /api/settings` → 400 `unknown setting 'google_finance_auto_refresh'`. `git grep -in "google.finance" origin/main` returns only prose in a migration, the backlog and two comments — no route, no setting, no template.

This is why the owner reports **"Google Finance does not appear in settings"**, and why **#67** — which restores the extension half believing the server half is on `main` — would ship four dead controls.

`codex/fix-idle-ui-cpu` and `codex/google-finance-control` now have **identical trees**, because #51's merge put the second inside the first. Either branch carries this work; this PR uses the better-named one.

## It honours the settings rule

Controls live in the side panel — `google_finance_auto_refresh`, `google_finance_refresh_hours`, `finance-save`, `finance-refresh`, `finance-dataset` are all asserted present in `extension/app.html`. The web page gets a **read-only** table plus a line pointing at the panel. The branch **extends** `tests/test_settings_live_in_the_extension.py` with a test asserting the controls are in the panel, rather than adding an exemption to it.

## One conflict, and the recommended resolution

`scrapex/rates.py` only. Four hunks:

1. **Docstring wording** — take this branch's; "the selected interval" is accurate once the interval is configurable.
2. **`if not automatic_refresh_enabled(conn): return False`** — take this branch's; it is the new off switch.
3. **`REFRESH_AFTER_S` → `refresh_interval_hours(conn) * 3600`**, plus the new `refresh_now()` — take this branch's. `DEFAULT_REFRESH_HOURS = 6.0`, so behaviour with no setting stored is unchanged.
4. **`refresh_if_due` body → `return refresh_now(conn, fetcher, now=stamp)`** — take this branch's, **but keep main's longer comment**. Main's version documents why the throttle stamp is committed *on its own* before the fetch (an escaping `CrawlBlocked` otherwise rolls it back and a blocked quote page becomes a request every half-second). `refresh_now()` does preserve that commit-before-fetch ordering — verified — but its two-line comment is the weaker record of a hard-won incident.

## Verified

I merged `main` + **#67** + this branch (hunk 4 resolved as above) + **#70**, in a scratch worktree:

- `python -m pytest -q` → **exit 0, zero FAILED**
- `node contract/parity/parity.test.mjs` → foldDigits 8/8, optionFingerprint 5/5, recordHash 3/3
- `node --test extension/tests/*.test.mjs` → exit 0 · `node --test apps_script/tests/*.test.mjs` → exit 0
- `validate-manifest` → OK, 11 sources

No duplication with #67: after both merges each panel id appears **exactly once** in `extension/app.html`, and the `app.js` handler once. `extension/app.js` is byte-identical between the two; only `app.html` differs, where #67 adapted it to today's markup.

**I have not merged this.** The conflict is real and the resolution above is a recommendation, not a decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,15 @@ jobs:
           echo "panel tests collected: ${count:-0}"
           test "${count:-0}" -ge 40
       - name: Tests (Python engine + Python-vs-frozen contract)
+        # tests/conftest.py restores a prebuilt schema template instead of
+        # replaying 57 migrations per test, which is what turns a ~30-minute
+        # local suite into ~6 minutes. It is a developer-loop convenience, and
+        # this is the line that keeps it one: CI runs the REAL stream 805 times,
+        # so the authority on green tests exactly what it tested before that file
+        # existed. Same reasoning as the panel gate above — a guard that can
+        # vanish quietly is the defect.
+        env:
+          SCRAPEX_FULL_MIGRATIONS: "1"
         run: python -m pytest -q
 
   contract-parity:
--- a/design/components.css
+++ b/design/components.css
@@ -1193,6 +1193,7 @@ code {
  * placement; everything visual and interactive is this. Behaviour: split-button.js.
  * ======================================================================== */
 .split-button {
+  position: relative;
   display: inline-flex;
   align-items: stretch;
   border: 1px solid var(--line-strong);
@@ -1234,7 +1235,7 @@ code {
   transform: none;
 }
 .split-button-menu {
-  position: relative;
+  position: static;
   display: flex;
 }
 .split-button-menu > summary { list-style: none; }
@@ -1250,6 +1251,12 @@ code {
   border-end-end-radius: calc(var(--radius) - 1px);
   cursor: pointer;
 }
+.split-button-trigger:focus-visible {
+  position: relative;
+  z-index: 1;
+  outline: 2px solid var(--focus);
+  outline-offset: -2px;
+}
 .split-button-trigger .sx-icon { width: 1.05rem; height: 1.05rem; }
 .split-button-chevron {
   margin-inline-start: auto;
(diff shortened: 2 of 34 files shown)


muhammadbayoumi:
Brought `main` in and resolved, so this is now judged on the merged result.

**`scrapex/rates.py`** — the four hunks as recommended in the description: the branch's configurable interval, its `automatic_refresh_enabled` gate, its `refresh_now`, and `refresh_if_due` delegating to it. **Main's long comment moved into `refresh_now`** rather than being replaced by the two-line summary: it is the record of a real incident (an escaping `CrawlBlocked` rolled the stamp back and a blocked quote page became a request every half-second, each preceded by rebuilding the 1.3s HTTPS client). `refresh_now` does preserve the commit-before-fetch ordering; the paragraph is what stops someone reordering those statements later.

**`tests/test_rates.py`** — both sides kept. `--theirs` would have taken main's whole file and dropped this branch's own new tests that had merged cleanly.

**One real semantic conflict, caught by a test rather than by git.** This branch adds a third dataset card for the Google Finance rates, and its caption shipped with no `title`. PR #68, merged an hour ago, clamps everything inside `.dataset-choice-detail` to a single line — so "Ready - refresh after price data is collected" (44 chars) became truncatable with no way back to it. Git reported no conflict; `test_every_dataset_freshness_state_can_be_read_in_full` failed. The caption now follows the same rule as its two siblings: built once, used for both the text and the title.

Full suite exit 0, zero FAILED. Parity, extension tests, apps_script tests all exit 0. `validate-manifest` OK, 11 sources.

PR #76, opened 2026-08-02
The version notice compares before it accuses

muhammadbayoumi:
Reported from the panel: a card headed **"The ScrapeX engine is older than this extension"**, with `Installed extension 0.1.0` and `Engine 0.1.0` printed directly beneath it, ending in **"Update the engine to 0.1.0."**

Two false statements on one card. The heading is contradicted by the two numbers under it, and the remedy is the version already installed.

## Cause

The branch fired on one fact — the engine published no capability report — and asserted a second one it never checked:

```js
if (!report && engine.reachable && engine.version) {
  notice.innerHTML =
    `<div class="setup-title">The ScrapeX engine is older than this extension</div>` + ...
    `Update the engine to ${esc(installed || "the extension's version")}.`
```

No comparison anywhere. And the remedy is built from the *extension's* version, so whenever the two sides match it prescribes what is already installed.

## Fix

Silence has three causes and they do not share a remedy:

| state | sentence | remedy |
|---|---|---|
| engine really is older | "The ScrapeX engine is older than this extension" | update the engine to the extension's version |
| both below the reporting floor | "This ScrapeX engine cannot say what it deploys" | update **both** to `CAPABILITY_REPORTING_SINCE` — the only number that changes anything |
| engine claims a reporting version and reported nothing | "This ScrapeX engine is not reporting what it deploys" | **restart it** — the files moved and the process did not; downloading anything is wasted effort |

`CAPABILITY_REPORTING_SINCE` is named on both sides and carried in the shared `version-vectors.json`, so the remedy cannot name a version one half does not believe in. It also reaches the live `/api/version` report, not just the test vectors.

## Guards

- `test_two_sides_at_the_same_version_are_not_told_one_is_older` — the reported case exactly
- `test_an_engine_at_a_reporting_version_that_stays_silent_is_told_to_restart` — the third cause
- `no refusal ever prescribes a version you already have` — walks every contract vector and fails any "Update the engine to X" where the engine is already at or past X
- `the reporting floor is the contract's, not a second opinion` — binds the two copies of the constant

**Proved to bite:** restoring the unconditional `engineIsOlder = true` fails both DOM tests; restored, all three pass. `node --test extension/tests` 19/19.

### DIFF
--- a/contracts/version-vectors.json
+++ b/contracts/version-vectors.json
@@ -1,6 +1,7 @@
 {
   "version": "0.2.0",
   "minimum_extension_version": "0.2.0",
+  "capability_reporting_since": "0.2.0",
   "note": "Both copies of capabilityProblem must reproduce every `problem` string exactly. Python: scrapex/version.py. JavaScript: extension/version.js.",
   "cases": [
     {
--- a/extension/app.js
+++ b/extension/app.js
@@ -9,7 +9,7 @@
 // Arabic renders right-to-left without disturbing the English chrome around it.
 import { checkEngine, getBackend, setBackend } from "./engine.js";
 import { autostartStatus, setAutostart, startEngine } from "./transport.js";
-import { capabilityProblem, deployedFrom, installedVersion } from "./version.js";
+import { capabilityProblem, deployedFrom, installedVersion, CAPABILITY_REPORTING_SINCE, isOlder } from "./version.js";
 
 const $ = (id) => document.getElementById(id);
 const esc = (v) => String(v ?? "").replace(/[&<>"']/g,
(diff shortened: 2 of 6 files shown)


muhammadbayoumi:
Confirmed live, and improved.

With the extension reloaded to 0.2.0 and the engine process still on 0.1.0, the card now reads correctly — the engine genuinely is older, and both numbers agree with the heading. The self-contradiction is gone.

But the remedy stopped at *Update the engine to 0.2.0*, and for this engine that is only half true. It runs on the owner's own machine from the same checkout, and it reports the version of the code **running**, not the code on disk. After a pull those differ until the process restarts — which is the ordinary case, not the rare one. The sentence was sending the owner to look for a download they already had.

It now says both, without claiming to know which: *Update the engine to 0.2.0 — and if its files are already 0.2.0, restart it so it loads them.*

Guarded in `test_an_engine_older_than_the_extension_says_so_in_different_words`. `node --test extension/tests` 19/19; the three notice tests pass.

muhammadbayoumi:
The reader asked what they were supposed to *do*, and they were right to.

The sentence was **Update the engine to 0.2.0 — and if its files are already 0.2.0, restart it so it loads them.** Nothing in the panel or on the page tells anyone which version is on disk. That is a decision the reader has no way to make, so it is not an instruction.

Restart first. It is free, it takes a second, and it fixes the common case — a process started before the last pull. Then the number itself answers the rest: if the engine still reports the old version afterwards, its files really are older.

All three sentences now open with the action and name where the control is — **"Restart engine"** on the ScrapeX Settings page, in the row that sits beside the very versions the card is talking about:

| state | what it now says |
|---|---|
| engine older | Press "Restart engine" on the ScrapeX Settings page. If it still reports 0.1.0 afterwards, its files really are older and need updating to 0.2.0. |
| both below the floor | Version reporting starts at 0.2.0, and both sides are below it. Update the files, then press "Restart engine" and reload this extension from chrome://extensions. |
| reporting version, silent | …the running engine is older than its own files. Press "Restart engine" on the ScrapeX Settings page — there is nothing to download. |

New guard: the free action must come before any mention of updating. `node --test extension/tests` 19/19; the three notice tests pass.

PR #90, opened 2026-08-04
Spark eShop comes back, whole this time

muhammadbayoumi:
SPARK_ESHOP was **crawled and then had its manifest entry deleted while the data stayed**: 1,789 products, 3,149 offers and 3,149 observations from a successful run on 2026-08-03, sitting in the warehouse marked active with nothing in the code able to crawl, update or explain them. That is **22% of every offer this warehouse holds**. This gives them a definition again.

## The owner asked whether I had inspected the site. I had not

The first draft copied `currency: EGP`, `default_region: EG` and `vat_mode: incl` out of the deleted entry and invented `min_expected_rows` / `max_drop_pct` from ELSEWEDYSHOP. Nothing in it had been checked against anything.

`/meta.json`, read 2026-08-03:

```json
{"name":"Spark","city":"El Weili","province":"Cairo","country":"EG",
 "currency":"EGP","ships_to_countries":["EG"],"money_format":"LE {{amount}}",
 "published_products_count":1789}
```

1,789 — matching the warehouse exactly. EGP and EG were right. **A guess that turns out right is still a guess**, so they are cited now rather than asserted. `vat_mode` remains unchecked and says so in place, instead of passing as a finding.

## The rows already in the warehouse are not trustworthy, and the entry says so

| the first draft asserted | the 3,149 stored rows say |
|---|---|
| currency EGP | **UNKNOWN**, all 3,149 |
| region EG | **`*`**, all 3,149 |
| — | `product_name` **empty on all 1,789** |
| — | **zero** attribute rows — not one published image |

And the **English** titles are in `product_name_ar`. `ABB 1SDA100487R1 | XT5H 630 TMA 630-6300` is not Arabic.

The images are explained by the **missing `enrichment` kind**, which this PR adds — `tests/test_images_every_connector.py` named it the moment the draft appeared. The other three are not explained. All four are recorded beside the source they describe, so nobody reads this entry and believes the data behind it is sound.

## The English shop now says so out loud

`<html lang="en">`, every title like *"Himel Variable Speed Drive VFD, 380V 3-Phase Motor Control"*, and **no `/en` locale at all** — that URL answers 404, verified live.

The connector's default is *"assume Arabic at the root, fetch /en for English"*, which is right for the eleven sources crawled under it and wrong here. Under that default the crawl files English titles into `product_name_ar`, asks a shop with no `/en` for its `/en`, fails, and leaves `product_name` — which `config.py` calls **required** — empty on every row. Then it **reports success**. That is exactly how the 3,149 rows were written.

`default_language: en` could not be added before #91 landed; the manifest would have rejected a key the model did not have. It is inert until someone activates the source, which is the point — **the declaration is in place before the first crawl rather than discovered after it.**

## Two guards, both proved to bite

Deleting `default_language: en` would fail nothing — no test, no validation, no error. It would silently recreate the defect on the next run. So a test asserts it, and asserts `active: false` beside it, because active would mean the scheduler replaces 3,149 rows tonight without the owner deciding to.

Dropping the line and flipping the flag each turn the suite red; the unmutated manifest stays green.

My first attempt at that proof reported both mutations **SILENT**. The guards were fine — my harness was reading pytest's output for a string it does not print, and its `active: false` edit hit the first source in the file rather than this one. **A proof that cannot fail proves nothing**, so it was rebuilt on exit codes and re-run.

## Gates, on the rebased result

`pytest` full suite · contract parity · extension **19/0** · apps_script **13/0** · `validate-manifest` **12 sources, active list unchanged at 7**.

Rebased onto `main` after #91; the two original commits are preserved.

## What this does not do

It does not repair the 3,149 rows. A re-crawl replaces them — **rewriting a stored observation is the one thing this project does not do** — and that is 9 requests whenever the owner wants it. The entry stays inactive until then.

---

## The guard this branch claimed to have (added after review)

An adversarial review found that `test_no_source_holds_data_the_manifest_has_forgotten` **could not fail**. It inserted a row called `GHOST` into an empty temp database and asserted that `GHOST` came back — true for any manifest, including one with every source deleted.

The review proved it end-to-end: it removed SPARK_ESHOP from the manifest, recreating the 2026-08-03 condition, and **the two tests beside this one went red while this one stayed green.** The test named after the incident was the only one that could not see it.

Worse, it guarded nothing that ships — **no production code read `source_site.active` anywhere in the repo**, so the "check" was a query living inside a test.

**A correction to commit `d66e8ac`**, which said *"Proved against the live warehouse rather than a fixture — deleting the entry again reports SPARK_ESHOP as an orphan."* The shipped test never opens the warehouse; it connects to `tmp_path/"w.db"`. That claim was false in the same commit that made it.

**What replaces it:** `storage.undeclared_sources(conn)` — real code, reading the *current* manifest each call because the failure mode is precisely that the manifest changed. It rides `storage.health()`, which the Storage page runs on every visit, so the fact reaches a screen the owner opens. `ok` stays true: a forgotten source is not corruption, and it is the one thing `quick_check` can never see.

Three tests, including **the negative half the old one never had** — a declared source must *not* be reported. Without it, a check that returned its whole input would pass. Each mutation hits exactly its own test.

### DIFF
--- a/scrapex/storage.py
+++ b/scrapex/storage.py
@@ -504,6 +504,35 @@ def space_warning(db_path: Path | str) -> str:
     return ""
 
 
+def undeclared_sources(conn) -> list[str]:
+    """Sources this warehouse holds data for that the manifest no longer names.
+
+    SPARK_ESHOP was crawled on 2026-08-03 — 1,789 products, 3,149 offers, 3,149
+    observations, a successful run — and its manifest entry was deleted
+    afterwards while the rows stayed, marked active. That was 22% of every offer
+    here belonging to a source no code could crawl, update or explain, and it
+    was found by hand-censusing the database two days later.
+
+    The manifest is the definition; the warehouse is the consequence. A
+    consequence with no definition is not a small untidiness: every count
+    includes it, every table shows it, and nothing can ever refresh it.
+
+    Reads the CURRENT manifest each call rather than a cached one, because the
+    failure mode is precisely that the manifest changed.
+    """
+    from .config import MANIFEST_FILE, load_manifest
+    try:
+        declared = {entry.source_key for entry in load_manifest(MANIFEST_FILE).sources}
+    except Exception:  # noqa: BLE001 — a broken manifest is validate-manifest's job
+        return []
+    try:
+        stored = {row[0] for row in conn.execute(
+            "SELECT source_key FROM source_site WHERE active = 1")}
+    except sqlite3.DatabaseError:
+        return []                      # not a warehouse; health says so already
+    return sorted(stored - declared)
+
+
 def health(db_path: Path | str) -> dict:
     """SQLite's own verdict, reported as a word plus the raw findings.
 
@@ -533,6 +562,7 @@ def health(db_path: Path | str) -> dict:
         freelist = conn.execute("PRAGMA freelist_count").fetchone()[0]
         page_size = conn.execute("PRAGMA page_size").fetchone()[0]
         identity_problem = _warehouse_identity(conn)
+        forgotten = undeclared_sources(conn)
     except sqlite3.DatabaseError as exc:
         return {"status": "unreadable", "ok": False,
                 "detail": f"SQLite could not read this file: {exc}",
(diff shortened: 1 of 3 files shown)


muhammadbayoumi:
**The owner asked whether I had inspected the site and reviewed the data. I had not**, and the answer changes this PR.

The entry copied `currency: EGP`, `default_region: EG` and `vat_mode: incl` straight out of the deleted draft, and I invented `min_expected_rows` and `max_drop_pct` from ELSEWEDYSHOP. Nothing in it had been checked against anything.

## What the shop says

`/meta.json`, read 2026-08-03:

```json
{"name":"Spark","city":"El Weili","province":"Cairo","country":"EG",
 "currency":"EGP","ships_to_countries":["EG"],"money_format":"LE {{amount}}",
 "published_products_count":1789}
```

1,789 — matching the warehouse exactly. `<html lang="en">`, and every product title is English: *"Himel Variable Speed Drive VFD, 380V 3-Phase Motor Control"*.

So EGP and EG were right. **A guess that turns out right is still a guess**, and they are cited now rather than asserted. `vat_mode` remains unchecked and says so in place instead of passing as a finding.

## What the stored rows say, which is worse

| the manifest asserted | the 3,149 stored rows say |
|---|---|
| currency EGP | **UNKNOWN**, all 3,149 |
| region EG | **`*`**, all 3,149 |
| — | `product_name` **empty on all 1,789** |
| — | **zero** attribute rows — not one published image |

And the English titles are in `product_name_ar`. `ABB 1SDA100487R1 | XT5H 630 TMA 630-6300` is not Arabic — that inverts the bilingual rule at its root.

The images are explained by the missing `enrichment` kind, which this PR fixes. **The other three are not explained.** All four are now recorded with their counts beside the source they describe, so nobody reads this entry and believes the data behind it is sound.

A re-crawl is what would replace those rows — not a repair of them, because rewriting a stored observation is the one thing this project does not do. The entry stays inactive until that is decided.

PR #107, opened 2026-08-05
Console and the Workspace toggle wore the same icon

muhammadbayoumi:
Stacked on #106 — GitHub will retarget this to `main` when that merges.

## What happened

The three new pages had never been captured, so the first time the rail was seen as a picture was after it shipped. **Console had `dashboard` — the icon the Workspace toggle four buttons above it already wore.**

The rail is icons only; every label is a tooltip or screen-reader text. A repeated symbol there is not a blemish — it is two destinations that cannot be told apart at the one moment the owner is choosing between them.

Console now wears `file-upload`, which is what it does: publish datasets outward into the six sheets the Excel add-in reads.

## Why it took a picture to find

Nothing in the suite compared one rail button against another. It does now:

```
Console wears the Workspace icon again   BITES
a rail button loses its icon entirely    BITES
```

The screenshot tool also gained the three pages as scenarios (`14-profile-welcome`, `15-engines`, `16-console`), so the next change to them is visible without anyone remembering to look. Every existing screenshot is regenerated — the merged rail grouping changed all of them.


muhammadbayoumi:
Closed by GitHub when its base branch was deleted on merge, not by a decision. Superseded by the PR above; the branch is the same work rebased onto `main`.

PR #116, opened 2026-08-06
[verification] Prove the CI split chooses the extension gate

muhammadbayoumi:
A one-line comment in `extension/app.js` and nothing else. If the split works, this pull request runs `pytest -m extension` — 328 tests — and never reaches the 2022-test engine suite.

Opened to verify #115 against real CI rather than against my own reasoning, and closed once the run is read.

### DIFF
--- a/extension/app.js
+++ b/extension/app.js
@@ -54,6 +54,7 @@ const state = {
 };
 
 // ---- views ----------------------------------------------------------------
+// (a one-line change used to prove the CI split chooses the extension gate)
 const VIEWS = [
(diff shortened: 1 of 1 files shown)


muhammadbayoumi:
Verified against real CI, which is what it was for:

| event | scope | duration |
|---|---|---|
| `pull_request` | **extension** | **3m28s** |
| `push` | full | 9m59s |

The pull-request path chose the extension gate and ran 328 tests. The push path fell back to `full` because a brand-new branch's `github.event.before` is all zeros — the safe fallback working as designed, but it means an extension-only change still pays for the whole suite once. Fixed in the follow-up.

PR #152, opened 2026-08-10
perf(extension): make the Side Panel shell immediate

muhammadbayoumi:
## Summary

- Render and wire the static Profile shell immediately on `DOMContentLoaded`.
- Resolve Account and Engine startup independently with finite deadlines, cancellation, and generation guards.
- Keep the bounded 100 ms paint opportunity as renderer protection, not as the host-opening fix.
- Replace implicit toolbar opening with one explicit `chrome.action.onClicked` handler that calls `chrome.sidePanel.open({windowId})`.
- Record toolbar, Side Panel host, document lifecycle, FCP, shell, Account, and Engine phases on one epoch clock.
- Keep Local Web and PR #151 unchanged.

## Release-blocking host-opening correction (`c8cd099`)

The owner confirmed that the bounded animation-frame fallback in `047e523` did **not** fix the defect: Chrome opened the Side Panel frame and title, but the extension content stayed blank until the webpage outside the panel was clicked.

### Controlled A/B status

| Strategy | Revision | Actual unpacked Side Panel result |
|---|---|---|
| A — implicit `openPanelOnActionClick: true` | `047e523` | **Failed (owner-confirmed):** host/title opened; content remained blank until an outside click. |
| B — explicit action opening | `c8cd099` | **Pending owner real-container verification.** This session cannot control installed Chrome, so no pass is claimed. |

Strategy B disables implicit action opening, retains the manifest's global `side_panel.default_path: app.html`, explicitly reinforces `{enabled: true, path: "app.html"}`, and registers exactly one action-click listener. The handler captures timestamps in memory, then calls `sidePanel.open({windowId})` as its first Chrome API call—without an `await` or configuration work that could consume the user gesture. Trace persistence happens only after `open()` has been invoked.

This follows Chrome's supported action-click opening path: [Side Panel API](https://developer.chrome.com/docs/extensions/reference/api/sidePanel). `minimum_chrome_version` is now 116, the first version supporting `sidePanel.open()`.

### Combined boundary trace

The session-local trace records:

- `set-panel-behavior-request` plus resolved/failed
- `set-panel-options-request` plus resolved/failed
- `action-click`
- `side-panel-open-request` plus resolved/failed
- `side-panel-on-opened` where Chrome 141+ exposes it
- `document-start`, `dom-content-loaded`, `pageshow`
- `first-contentful-paint`
- `focus`, `blur`, `visibilitychange`
- animation-frame request/resolution and bounded paint-opportunity resolution
- `shell-interactive`, `shell-post-opportunity`
- Account and Engine check start/finish

The combined trace is stored only in `chrome.storage.session` under `scrapexSidePanelStartupTrace`; it does not contain account identity or tokens. After a no-interaction attempt, it can be inspected from the extension service-worker console with:

```js
chrome.storage.session.get("scrapexSidePanelStartupTrace").then(
  ({scrapexSidePanelStartupTrace: trace}) =>
    console.table([...trace.events].sort((a, b) => a.at - b.at))
)
```

`shell-visible` has been renamed to `shell-post-opportunity`. Only the browser's native FCP entry claims an actual visual paint.

### Boundary still requiring evidence

The installed-Chrome controller is unavailable in this Codex session, so these facts are intentionally **not** claimed yet:

- whether the owner's failing implicit attempt reached `document-start` before or only after the outside click;
- whether explicit Strategy B removes the blank host state;
- the minimal static-page result;
- three consecutive cold and three consecutive warm no-interaction passes.

No diagnostic HTML page is shipped. If Strategy B still fails, the next controlled real-container step is the requested minimal static `default_path` A/B, not another timing workaround.

## Before / after in-document measurements

Baseline is the audit at `3dbf962`. These are the existing traceable in-document measurements at `047e523`; `c8cd099` changes the toolbar/host boundary and instrumentation, so it does not manufacture a new real-container result. Each current row has 10 headed Chromium samples; values are median / worst.

| Scenario / boundary | Baseline audit | In-document result at `047e523` |
|---|---:|---:|
| Cold unavailable, document → FCP | 260–656 ms observed range | 133.6 / 587.7 ms |
| Cold unavailable, document → shell interactive | not separately instrumented | 121.9 / 140.2 ms |
| Cold unavailable, document → shell post-opportunity | not separately instrumented | 147.1 / 224.9 ms |
| Warm unavailable, document → FCP | 68–240 ms observed range | 94.2 / 120.3 ms |
| Warm unavailable, document → shell interactive | not separately instrumented | 90.9 / 106.9 ms |
| Warm unavailable, document → shell post-opportunity | not separately instrumented | 105.9 / 122.6 ms |
| Warm unavailable, shell → Account settled | serialized within prior init | 6.5 / 11.8 ms |
| Warm unavailable, shell → Engine settled | serialized within prior init | 11.5 / 20.8 ms |
| Warm nonresponsive Engine, document → shell interactive | not separately instrumented | 193.1 / 224.4 ms |
| Warm nonresponsive Engine, document → shell post-opportunity | not separately instrumented | 213.3 / 252.5 ms |
| Warm nonresponsive Engine, shell → Account settled | serialized within prior init | 8.2 / 14.2 ms |
| Warm nonresponsive Engine, shell → Engine timeout | ~2,506 ms health; ~2,535 ms init (single baseline) | 2,517.2 / 2,528.2 ms |

FCP remains authoritative. `shell-interactive` means the local DOM shell is wired; `shell-post-opportunity` means startup yielded the bounded renderer opportunity. Chrome did not expose toolbar click timing to the document in the old strategy, so those numbers never claimed toolbar click → host/document startup.

## Automated verification

- `python -m pytest tests/test_panel_startup.py -q` — 23 passed.
- `python -m pytest -m extension -q` — 462 passed, 1 intentional skip.
- `node --test extension/tests/*.test.mjs` — 52 passed.
- Host-opening regression proves global `app.html` configuration, implicit opening disabled, one action handler, `open()` as the first Chrome API call, one request per click, and combined trace propagation.
- Initially hidden/unfocused regression with withheld animation frames and no click/focus/blur/visibility event — passed; shell, Account, and Engine settle once through the bounded fallback.
- `node contract/parity/parity.test.mjs` — 16/16 frozen parity cases passed.
- Exact CI ESLint command — passed.
- Manifest validation — passed.
- `python tools/sync_design_assets.py --check` — passed.
- `git diff --check` — passed.

The prior 320/400 px light/dark/default/alternative harness matrix remains green. It is not presented as a real Chrome Side Panel container test.

## Required real-container acceptance before approval/merge

Using `C:/Users/User01/source/repos/ScrapeX-extension-startup/extension` as the unpacked extension:

- reload the extension after `c8cd099`;
- run three consecutive cold and three consecutive warm opens;
- leave focus on the webpage and do not click, hover, move focus, or trigger visibility for at least 15 seconds;
- verify Profile shell/navigation appear automatically in under one second;
- verify Account and Engine start and update independently;
- verify no duplicate requests/panels;
- preserve the trace from any failure before further code changes.

Real-container acceptance remains release-blocking. Do not merge until it passes and final review approves.

## Parallel integration / PR #151

Expected textual overlap remains limited to the existing full-PR overlap in `extension/app.html`, `extension/app.js`, and `tests/test_panel_dom.py`; this correction additionally changes Side Panel host/instrumentation files that PR #151 is not expected to touch. No changes were made to PR #151, `main`, or Local Web. No rebase was performed.

Auto-merge remains disabled.


muhammadbayoumi:
@codex review

muhammadbayoumi:
@codex review

Please review the no-interaction startup correction in 047e523, including the bounded paint fallback, corrected shell-visible semantics, deterministic first-frame test, and initially hidden/unfocused frame-starvation regression. Real Chrome Side Panel cold/warm owner verification remains release-blocking and is not claimed by the automated results.

muhammadbayoumi:
@codex review

Please review `c8cd099`, which replaces the owner-confirmed failing implicit host-opening path with the official explicit action-click strategy.

Review focus:

- `openPanelOnActionClick` is disabled and exactly one action handler calls `sidePanel.open({windowId})`.
- `open()` is the first Chrome API call in the click handler; trace persistence occurs afterward, so no setup await or diagnostic API consumes the user gesture.
- The manifest and runtime configuration keep the global panel enabled at `app.html`.
- Toolbar/config/open/document/FCP/focus/visibility/shell/Account/Engine events share one epoch-clock session trace.
- `shell-visible` is now `shell-post-opportunity`; only native FCP claims paint.
- The 100 ms frame fallback remains bounded protection and is not presented as the host-opening fix.
- New tests cover single-handler/single-open behavior, API ordering, trace propagation, and initially hidden/unfocused no-interaction startup.

Automated checks are green, including 462 passed + 1 intentional skip in the full extension gate and 52/52 extension Node tests. Owner cold/warm real Chrome Side Panel verification of Strategy B remains release-blocking and is not claimed by this update. Do not merge yet.

muhammadbayoumi:
## Root cause found and measured — `app.js` delays `load`, and Chrome shows the panel on `load`

**Five measurements on the owner's Chrome 151**, one toolbar click each, nothing clicked afterwards:

| page | carries | result |
|---|---|---|
| `diagnostic-panel.html` | nothing at all | **painted immediately** |
| `diagnostic-styles.html` | 3 render-blocking stylesheets | **painted immediately** |
| `diagnostic-scripts.html` | 3 synchronous head scripts | **painted immediately** |
| `diagnostic-both.html` | both halves together | **painted immediately** |
| `diagnostic-app-nomodule.html` | app.html **minus app.js** | **painted immediately** |
| `app.html` | all of it | **blank until a click** |

The last two differ by **one line**: `<script type="module" src="app.js">`.

### The mechanism, from this branch's own trace

| | `load` | → `visibilitychange` (visible) | gap |
|---|---|---|---|
| trace 1 | 1259.4 ms | 1268.3 ms | **+8.9 ms** |
| trace 2 | 2363.5 ms | 2370.0 ms | **+6.5 ms** |

**Chrome makes the side panel document visible immediately after `load`.** Twice, under 10 ms apart.

`load` waits for the module graph and the requests it opens. While `app.js` calls an engine that is not running and retries, `load` does not fire → the document is never marked visible → rAF never runs, timers are throttled, nothing paints. The click outside the panel is what ends that state.

**The shell was `shell-interactive` at 168 ms in every run.** It was ready the whole time.

### What this refutes — all measured, not argued

- **Windows native occlusion.** Owner ran Chrome with `--disable-features=CalculateNativeWinOcclusion --disable-backgrounding-occluded-windows`: no change. This is `HIDDEN`, not `OCCLUDED`.
- **The opening strategy.** Strategy A (`openPanelOnActionClick`) tried against the real panel: no change.
- **Zero-width first layout.** Reproduced in a harness and refuted — layout recovers pixel-identically on resize.
- **CSS, appearance tokens, stylesheets, head scripts** — each cleared by its own measurement above.
- **"The panel waits for the engine before showing its shell"** — I claimed this; the trace disproves it. `shell-interactive` 168 ms, `engine-check-start` 1158 ms. The engine's absence matters **only** because it holds `load` open.

### The fix, not in this PR

Nothing that touches the network may hold `load`. Engine and account work starts **after** `load`, so `load` fires in milliseconds and Chrome shows the panel whether or not an engine is running — the state every user is in before they install one.

### Status

`OPENING_STRATEGY` is back to `"explicit"`. `node --test`: 77 passed, 0 failed. Diagnostic pages live in `extension/tests/`, which the release workflow deletes and then fails the build over, so they cannot ship.

**Not ready to merge as a fix for the blank panel** — it does not fix it and does not claim to. It carries the instrumentation that found the cause, the `setOptions` race removal, and two self-inflicted startup waits removed.

muhammadbayoumi:
## Correction: `app.js` does not delay `load`. I measured it and I was wrong.

I recorded here that `app.js`'s module graph holds `load` open. **It does not.** Served over HTTP and loaded in an ordinary browser tab:

| | |
|---|---|
| `domContentLoadedEventStart` | 47 ms |
| `loadEventStart` | 64 ms |
| **gap** | **17 ms** |

Exactly **one** resource finishes after DOMContentLoaded — the favicon, at 53 ms. There is no resource in this page capable of holding `load` for a second.

On the owner's side panel the same gap is **1138 ms** (121 → 1259). So the delay is not a resource and not the module graph. It is **throttling**: the document is created hidden, and a hidden document's timers are throttled — proved by the trace itself, where a 250 ms fallback arrived at **1131 ms**.

### And the biggest consumer of that window was ours

`afterNextPaint` waited for a frame that a hidden document never gets, then fell through to a timer that is itself throttled:

```
785 ms   paint-opportunity-resolved   source: "timer"
1148 ms  shell-post-opportunity
1259 ms  load
```

**That is already fixed in `c709a2a`** — a hidden document no longer requests a frame at all, and resolves immediately.

### It has never been tested

Every run after `c709a2a` used a *diagnostic* page. `app.html` itself has not been opened once with the fix in place. That is the next measurement, and it is one click.

### What the bisect actually established

The five-page bisect is still valid and still useful: nothing in the stylesheets, the head scripts, or both together loses the first frame, and removing `app.js` restores it. But the reason I gave for *why* removing `app.js` helped was wrong. The correct reading is that `app.js` is what brings the throttled paint-opportunity wait into the page — not that it holds `load` on a resource.

muhammadbayoumi:
## Confirmed by the owner: the panel now paints immediately

`c2e5bf7` — verified in real Chrome 151 on the owner's machine. One toolbar click, no click anywhere else, content appears at once.

### The deadlock, and how it was broken

Chrome creates the Side Panel document **hidden** and reveals it only after `load` — 8.9 ms and 6.5 ms after it, in two independent traces. A hidden document also has its **resource loading deprioritised**. So the panel stayed hidden until `load`, and `load` was slow *because* the panel was hidden.

The third trace settled it. Taken after `startup.js` stopped waiting for a frame a hidden document never gets — a fix that worked exactly as intended, moving `paint-opportunity-resolved` from 785 ms to **177.6 ms** with source `"hidden"`:

| | |
|---|---|
| all of our own work finished | **825 ms** (engine reachable, fully settled) |
| `load` | **2024 ms** |
| nothing of ours running in between | **1199 ms** |

The same page in an ordinary visible tab, served over HTTP and measured with resource timing: `domContentLoaded` 47 ms, `load` 64 ms — a **17 ms** gap, one resource finishing after DCL (the favicon).

**The module graph was never slow. It was slow because the panel was hidden.**

### The fix

`app.js` is appended by `boot-app.js` when `load` fires, so nothing in front of `load` depends on it. The shell is static markup and needs no JavaScript to be correct — established separately by auditing the whole ancestor chain, and by the owner's own bisect: `app.html` minus that one line painted on the first click while `app.html` unchanged did not.

**The trade, stated rather than hidden:** the panel becomes visible before it is interactive. Someone may see the shell for a moment before the buttons respond. Strictly better than a blank panel that stays blank until they click somewhere else.

### What this cost, and what it refuted

Four hypotheses were measured and killed before the right one: Windows native occlusion (owner ran Chrome with the occlusion switches — no change), the opening strategy (Strategy A tried against the real panel — no change), a zero-width first layout (reproduced in a harness and refuted), and the engine's absence (shell was interactive at 168 ms, engine check started at 1158 ms).

I also recorded two wrong mechanisms on this PR and corrected both: that the panel waits for the engine before showing its shell, and that `app.js` holds `load` on a resource. Same conclusion in the end, wrong reasoning twice.

### Ready for review

`node --test`: 77 passed, 0 failed. `pytest test_panel_dom.py test_ui_kit.py`: all passed. The ordering test now asserts `app.js` is **absent** from the markup — putting the tag back fails two tests, which was verified by putting it back.

Diagnostic pages remain under `extension/tests/`, which the release workflow deletes and then fails the build over, so they cannot ship. `OPENING_STRATEGY` is `"explicit"`.

muhammadbayoumi:
## Confirmed working: the panel paints on the first click **and** the tool functions

`cf2cc37`, verified by the owner in real Chrome 151.

### The two commits that did it

**`c2e5bf7`** — `app.js` is appended by `boot-app.js` when `load` fires, so nothing in front of `load` depends on the module graph.

**`cf2cc37`** — and the module actually starts when it arrives. The first commit made the panel paint instantly and left every control dead: `app.js` began with an unconditional `DOMContentLoaded` listener, that event fires at ~120 ms, `load` at ~2000 ms, so the listener was registered for something that would never happen again and `init()` was never called. **A blank panel traded for a silent dead one is not an improvement.** The entry point now asks `document.readyState`, which has an answer in both worlds.

### The deadlock, for the record

Chrome creates the Side Panel document **hidden** and reveals it only after `load` (8.9 ms and 6.5 ms after it, two traces). A hidden document has its **resource loading deprioritised**. So the panel stayed hidden until `load`, and `load` was slow because the panel was hidden.

The decisive numbers, from the trace taken after our own paint-opportunity wait was removed:

| | |
|---|---|
| all of our work finished | **825 ms** |
| `load` | **2024 ms** |
| nothing of ours running between | **1199 ms** |
| same page, ordinary visible tab | DCL 47 ms → load **64 ms** |

### What it took

Four hypotheses measured and killed first: Windows native occlusion (occlusion switches — no change), the opening strategy (Strategy A against the real panel — no change), zero-width first layout (reproduced in a harness, refuted), and the engine's absence (shell interactive at 168 ms, engine check started at 1158 ms).

Then a five-page bisect, one owner click each, which cleared the stylesheets, the head scripts, and both halves together — and convicted the one line that loaded `app.js`.

**I recorded three wrong mechanisms along the way and corrected each on this PR**: that the panel waits for the engine before showing its shell; that `app.js` holds `load` on a resource; and I shipped a fix without checking what the module does when it arrives late. The conclusions converged; the reasoning was wrong twice and incomplete once.

### Gates

`node --test` 78 passed, 0 failed. `pytest test_panel_dom.py test_ui_kit.py` all passed. The ordering test asserts `app.js` is **absent** from the markup, and a second test reads the entry point — both fail when their change is reverted, verified by reverting them.

Diagnostic pages stay under `extension/tests/`, deleted by the release workflow which then fails the build if they survive. `OPENING_STRATEGY` is `"explicit"`.

**Ready for review.**

PR #178, opened 2026-08-12
Pick a spreadsheet you already had, and a policy that cannot go stale

muhammadbayoumi:
Two halves of one request: **pick an existing spreadsheet**, and **make sure the published privacy policy stays current, tied to a date**.

---

## Half one — the chooser

`drive.file` reaches two kinds of file: ones ScrapeX created, and ones the owner **hands** it. Until now only the first existed, so "use a spreadsheet I already have" was impossible. The handing is done by Google Picker.

**Picker cannot run in this panel.** It loads `apis.google.com/js/api.js`, and MV3 forbids an extension executing code fetched from a server. There is no flag for this and no workaround inside the panel — the chooser has to run on an ordinary web origin.

So [docs/picker/scrapex-picker.html](docs/picker/scrapex-picker.html) is that origin. What it is, and what it deliberately is not:

| | |
|---|---|
| Gets the token | in the URL **fragment** — browsers do not send fragments to servers, so it never reaches a log |
| Keeps the token | no. The fragment is erased with `history.replaceState` on load, so it does not outlive the click |
| Calls any API | no. It draws the Picker and nothing else |
| Sends anywhere | only `chrome.runtime.sendMessage` to one hard-coded extension id |
| Returns | a file id and a name. Not a token, not a document |

The panel then opens that file with **its own** token through `sheets.js`, exactly as it opens one it created.

The page ships with an **empty API key and refuses by name**, exactly as `WEB_CLIENT_ID` did until the client existed. A placeholder would fail against Google with a message about the key rather than about the thing that is actually missing. (See *What only you can do* below.)

### The door this opens, and what stands in it

`externally_connectable` lets a web origin talk to the extension. That is a real widening, so it is guarded three times:

1. **The manifest** admits exactly one origin.
2. **`background.js` checks the origin again.** A match pattern is one character from being widened, and this listener is the last thing between a web page and the panel. `https://muhammadbayoumi.github.io.example.com` is a host a careless prefix match would admit; the test uses exactly that.
3. **Only two fields are copied**, by name. A message carrying a token, a URL and a mime type loses all three here — a future field cannot arrive in the panel without someone adding it on that line.

The choice goes to `storage.session`, not `local`: a document chosen once is not a preference worth writing to disk after the browser closes.

### What the tests do

Six drive the **real listener** out of `background.js`: the look-alike origin, an absent origin, four wrong message shapes, a message carrying fields it was never asked for, a cancellation, and a non-string id.

Both mutations were run:

| Mutation | Result |
|---|---|
| Delete the origin check, trust the manifest | **2 fail** |
| `{...message}` instead of copying two fields | **3 fail** |

Six more read the **public page itself**, because it is served to everyone for ever: no `GOCSPX-` secret, no `client_secret`, no refresh token, no `ya29.` token, no token in a query string, and the extension id read *and checked* rather than trusted.

Both node harnesses gained `onMessageExternal`. Without it `background.js` throws while loading and every test in those files fails with a message about `addListener` rather than about the panel — which is how it presented, and is worth knowing.

---

## Half two — a date that cannot drift

A privacy policy is a promise with a date on it, and a date that drifts is a lie nobody notices.

Two guards compare the **stated** date against the file's **last real change in git**. They fired the moment they were written:

| File | Said | Last actually changed |
|---|---|---|
| `privacy-policy.md` | 6 August 2026 | **12 August 2026** |
| `support.md` | 6 August 2026 | **8 August 2026** |

Both corrected. The policy had been edited twice — new storage rows, three new hosts — while still claiming to be a week old.

`publish-docs.yml` now also runs **daily**, and a new `drift` job compares the sha256 of each *published* file against this repository's copy, reporting both `Last updated` lines. `publish` runs only `if: needs.drift.outputs.drifted == 'yes'` — so a scheduled run that finds the site already correct is silent, rather than a stream of empty commits.

---

## What only you can do

1. **Create the Picker API key.** Cloud Console → *APIs & Services* → *Credentials* → *Create credentials* → *API key*. Restrict it to the **Picker API** and to the page's own origin — an unrestricted key in a public page is a key anyone can spend your quota with. Then it goes in `API_KEY`.
2. **Decide whether this page may be published to your site.** I have not pushed anything to `mbiXsite`. Publishing is a public act under your name, and it is yours to make.

Until both, the button exists and explains itself instead of failing.

## Measured

| | |
|---|---|
| node | 174 / 174 |
| engine (`-m "not extension"`) | 1830 passed, 2 skipped, 1 xfailed |
| extension (`-m extension`) | 57 passed |
| policy | 27 passed |
| eslint, ruff | clean |

No new OAuth scope. No new permission beyond the one origin named above.

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -110,13 +110,22 @@ jobs:
 
   publish:
     needs: drift
-    # The tests still run on every path; only the WRITE is skipped when there is
-    # nothing to write. A scheduled run that finds no drift ends green having
-    # proved the published policy is the one this repository checks.
-    if: needs.drift.outputs.drifted == 'yes'
+    # NO `if:` HERE, ON PURPOSE. It was on the job, and the comment beside it
+    # claimed the tests still ran on every path -- which the `if:` itself made
+    # false: skipping the job skipped the guards too. A daily run that found no
+    # drift then proved nothing at all, while reporting green.
+    #
+    # The condition now sits on the two steps that WRITE. Everything that CHECKS
+    # runs every time.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          # FULL HISTORY, because the privacy-policy guards below ask git when a
+          # document last really changed. A depth-1 clone grafts HEAD, so every
+          # file looks newly added and the guard answers with HEAD's date for
+          # everything -- confidently, and wrongly.
+          fetch-depth: 0
       - uses: actions/setup-python@v5
         with:
           python-version: "3.12"
@@ -132,6 +141,7 @@ jobs:
           python -m pytest -q tests/test_the_privacy_policy_is_true.py
 
       - name: Publish to the public delivery endpoint
+        if: needs.drift.outputs.drifted == 'yes'
         env:
           GH_TOKEN: ${{ secrets.PUBLIC_SITE_TOKEN }}
         run: |
@@ -151,6 +161,7 @@ jobs:
           done
 
       - name: The pages the store depends on must actually work
+        if: needs.drift.outputs.drifted == 'yes'
         # PRINTING A URL IS NOT CHECKING IT. This step used to end by echoing
         # the two page addresses, which proves nothing: publishing can succeed
         # while the page stays blank for ever, and the chain that follows is the
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -33,6 +33,26 @@ on:
     paths:
       - "docs/privacy-policy.md"
       - "docs/support.md"
+  # A DAILY LOOK AT WHAT IS ACTUALLY PUBLISHED, because the push trigger above
+  # only fires when this repository changes and says nothing about whether the
+  # publish SUCCEEDED or survived.
+  #
+  # Three ways the published copy drifts from this one, none of which touches
+  # `docs/` here and so none of which the push trigger can see:
+  #   * a publish run that failed after the tests and before the write
+  #   * an edit made directly in mbiX-hub
+  #   * a revert, a force-push, or a restore in that repository
+  #
+  # The owner asked for this on 2026-08-12 in terms of the DATE — "if the
+  # published date differs from ours, run the workflow". The check below
+  # compares the whole file rather than the date line, which is strictly
+  # stronger: a date can match while the text underneath has moved, and it is
+  # the text that makes promises. The date is reported because it is the line a
+  # human reads first.
+  schedule:
+    # 04:11 UTC. An odd minute on purpose: GitHub queues every job scheduled on
+    # the hour, and a policy check is not worth waiting in that queue.
+    - cron: "11 4 * * *"
 
 permissions:
   contents: read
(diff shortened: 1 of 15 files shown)


muhammadbayoumi:
## Attacked before merging, and it did not survive intact

I wrote this branch, so I did not review it myself. Five agents attacked it with distinct lenses — the web→extension door, the public page holding a token, the panel side, the two guards, and the whole diff — and a sceptic tried to refute everything they raised. **20 findings, 8 verified, 6 confirmed**, three of them distinct. All three were real.

### 1. The feature was a lie · critical

Choosing a spreadsheet confirmed the file, painted its name into the panel's spreadsheet line, and printed *ScrapeX can now write to it*. Then it threw the id away. `exportSourceToSheet` went on calling `ensureSpreadsheet(DEFAULT_WORKBOOK)` unconditionally — which resolves "ScrapeX Data" **by name**.

Both halves were separately defensible. The Picker's grant really is permanent, and the export line really did name the file it wrote to. Together they told the owner their rows were in a workbook that never received a row.

Fixed: the choice is kept, `whereExportsGo` is the single answer to *where do the rows go*, the panel states which workbook is in force the moment it opens, and **Create a spreadsheet** is the way back and says so.

**My first test for this passed against the mutation.** It drove `whereExportsGo` directly — proving the helper worked and nothing about the export. It now drives `exportSourceToSheet` and asserts on the id `writeTab` receives. A fourth test covers what my third got backwards: a *failed* creation must **not** abandon the workbook in force.

### 2. The token was readable by every extension with `tabs` · critical

It travelled in the tab's fragment, defended by the true and irrelevant fact that browsers do not send fragments to servers. `chrome.tabs.create` **commits** that URL, and the committed URL is delivered whole — fragment included — to every extension holding `tabs`, via `onCreated` and `onUpdated`.

The sceptic corrected me on the shape of it: **this is not a race the page can win.** The delivery has already happened before any script runs, and `history.replaceState` fires one *more* `onUpdated`.

Fixed at the root: the URL carries a single-use nonce, traded once through `background.js` for the token. A watching extension copies a number that is spent within a second, refused thereafter, expires inside the same window, and is useless without also being able to send from the chooser's origin. A wrong guess deliberately does **not** consume the handoff — a guesser must not be able to break the owner's chooser at will.

### 3. My own date guard answered confidently and wrongly · critical

In a shallow clone HEAD is grafted, so `git show` prints the **whole file** as added lines for every path, touched or not. The date-only skip never fires and `_last_changed` returns HEAD's date for everything — it does not skip, it asserts something false.

`publish-docs.yml` and `release-extension.yml` both checked out at depth 1 and both run this file. **This would have failed on main after the merge.** Both now fetch full history; the helper refuses to guess if one ever stops.

### And a comment that was lying

`publish-docs.yml`'s `if:` was on the **job** while the comment beside it claimed the tests still ran on every path. Skipping the job skipped the guards, so a daily run finding no drift proved nothing and reported green. The condition now sits on the two steps that *write*.

### Refuted, and not changed

- *The remote api.js reads the fragment before replaceState* — true mechanically; the proposed fix prevents nothing under its own threat model, and the nonce makes it moot.
- *A choice made while the panel is closed is stranded* — nothing durable was lost.

### Measured

| | |
|---|---|
| node | 181 / 181 |
| extension | 570 passed |
| engine | 1830 passed, 2 skipped, 1 xfailed |
| eslint, ruff | clean |

The one engine failure in a full run is OP-19's known load sensitivity — it passes twice in isolation, and this branch changes no engine file at all (`git diff --stat -- scrapex/` is empty).

The Picker API key is in place, restricted to the Picker API and one site. It is public by design; a test refuses anything in that slot not shaped like a browser key.

PR #186, opened 2026-08-13
The Console opens, and refuses any workbook but the right one (A3)

muhammadbayoumi:
A page in a **tab** — not a view in the panel, not a page on the website.

| | why not |
|---|---|
| a panel view | six sheets of configuration are a table editor; the panel is 360px |
| a page on the site | the token would have to leave the extension, and the `externally_connectable` widening the Picker needs is not worth repeating for a surface that does not need it — the Sheets API is a plain `fetch` that `host_permissions` already covers |

Cost in the manifest: **zero**. `release-extension.yml` copies the whole directory, so a new page ships by existing.

## The check that comes before every other check

The add-in does not find its configuration by name. `EndpointCatalog` carries **six tab ids compiled into the build** and fetches each directly. So a workbook can have all six tab *names*, look perfect, and be a copy the add-in has never read.

The Console compares the six gids first and **refuses to display anything** if they disagree — naming which tab is which id. Checking the wrong file and reporting it correct is worse than saying nothing.

Then it reports **before** it offers to change: findings first, sheets after. A Console that opened on an editor would invite a change while the file already had faults nobody had been shown.

## One picker, not two

`extension/picker.js` is the nonce handoff extracted whole. The panel picks the workbook exports go into; the Console picks the add-in’s configuration. A second copy of a token flow is a second place for the rules to drift.

## Two guards were blind to every page but `app.html`

Both widened — and the first one immediately earned it:

- `test_ui_kit.py::_markup` globbed one file. Widened to every shipped page, it found **`class="body"` in `onboarding.html`** — three elements carrying a class no stylesheet defines, shipping since it was built. The class is deleted and the elements stay: `.step` is a flex row and that div is what holds the text beside the number.
- `test_the_interface_stays_english` read `app.html` alone. Now parametrised over every page — which matters most for the Console, because it edits a workbook whose own **data** is bilingual.

## And one guard caught me

`test_a_page_that_is_only_a_shape_says_so_on_itself` failed when I deleted the Console view’s "not built" note. **Correctly** — the Datasets card below is still a shape.

But that test could not express a page that is *half* built, which is exactly what Console became. So the rule moved from a list of pages onto the **class**: every `.planned-note` must name its milestone, must sit inside the card whose controls it excuses, and every control in that card must be dead.

| mutation | result |
|---|---|
| a live button inside the card that says it is unbuilt | fails |
| a note with no milestone | fails |
| a note moved outside its card | fails |

It now catches a live button beside a "not built" note **wherever it appears**, rather than only on pages somebody remembered to list.

## Verification

584 extension tests · 210 node · eslint clean · ruff clean · **no engine file changed**.

The privacy policy gained the row the Console’s remembered workbook needs — the guard added earlier today refused the commit until it did.

### DIFF
--- a/contract/addin-contract.json
+++ b/contract/addin-contract.json
@@ -139,6 +139,18 @@
   "constants": {
     "TRANSFORM_SEPARATOR": "|",
     "TRANSFORM_ARGUMENT_SEPARATOR": ":",
-    "BLOCKS_SYNC_FROM": "Error"
+    "BLOCKS_SYNC_FROM": "Error",
+
+    "_comment_uri": [
+      "The three literals SourceUriValidator searches for inside the WHOLE",
+      "address, case-insensitively (SourceUriValidator.cs:80-118). They are",
+      "contract facts exactly like the transform separator above: change one in",
+      "the add-in and every Console warning about an address becomes wrong.",
+      "They live here rather than inline for the same reason the boolean",
+      "spellings do — so a change to the add-in has one place to land."
+    ],
+    "URI_GOOGLE_SHEETS_MARKER": "docs.google.com",
+    "URI_TSV_MARKERS": ["output=tsv", "format=tsv"],
+    "URI_TAB_MARKER": "gid="
   }
 }
--- a/docs/privacy-policy.md
+++ b/docs/privacy-policy.md
@@ -34,6 +34,7 @@ your email address. The extension contains no telemetry of any kind.
 | The time zone dates are shown in | The same | You |
 | The address of your local engine — `127.0.0.1:8000` unless you change it | The same | You |
 | The spreadsheet you chose for exports — its name, its link, and Google's id for it | The same | You |
+| The spreadsheet holding your Excel add-in's configuration, if you open the Console — its name and id | The same | You |
 
 Nothing in that table leaves your computer unless you press a button that says
 it will.
--- a/docs/privacy-policy.md
+++ b/docs/privacy-policy.md
@@ -1,6 +1,6 @@
 # ScrapeX — Privacy Policy
 
-*Last updated: 12 August 2026*
+*Last updated: 13 August 2026*
 
 ScrapeX is a Chrome extension and a companion program, **ScrapeX-Engine**, that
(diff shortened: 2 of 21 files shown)


muhammadbayoumi:
## A second commit: thirteen of the fifteen orphans I reported were mine

While reading `DataSourceEntity`'s own rules to build the editor, I found one the checker did not have — and it invalidates most of what this PR's predecessor reported.

**A blank `PROFILE_KEY` is not an absent one.** The add-in resolves an empty `PROFILE_KEY` — and the literal `DEFAULT`, case-insensitively — to the row's `TARGET_ENTITY_KEY`, then looks up `4.DataMap` under that name.

So every source that leaves the column blank looked like a source referencing nothing, and every profile named after its entity looked orphaned.

Measured against the real workbook, both ways:

| | orphans |
|---|---|
| my rule | **15** |
| the add-in's | **2** — `GARB`, `GARB2` |

The thirteen — `T_DIESEL`, `T_EG_GARB_BOQ`, `T_EUROCODE_LIB`, `T_LV_WIRES`, `T_SUPPLIER`, `T_GPP_ENERGY`, `T_SA_Aramco`, `T_SIKAEGSHOP`, `Egypt_Compensation_Rates_Library`, `mbiXaddin_COUNTRY_REFERENCE`, `mbiXaddin_SHEET_LIBRARY`, and the two Saudi Contractors Authority tables — are all wired correctly.

**The whole file's count goes from 16 to 3.**

### Why this mattered more than the feature it delayed

A checker that reports thirteen false problems teaches its owner to stop reading it, and then it is not there for the two that are real. Overstating is not the safe direction.

The resolution now happens in one place and **both** readings of `PROFILE_KEY` use it — the orphan check and the attribute check. A test covers the second specifically, because teaching only the first would have stopped the false orphans and started missing real attribute faults: the same defect wearing the opposite sign.

### And the first draft of these tests passed for the wrong reason

They wrote over `values[0]`, which is the **header** row, so the sheet did not parse and nothing was reported — including the thing being asserted. Fixed to rows 1 and 2, with the comment saying so. Removing the resolution now fails three of the four.

214 node tests, eslint clean.

PR #253, opened 2026-08-22
The release the documents ask for is the release that would run

muhammadbayoumi:
## What he saw

    Installed version   Not detected
    Latest version      0.2.1   [Available to install]
    Protocol            Not available

— with `scrapex/version.py:76` reading `VERSION = "0.3.0"`. *«المحرك الموجود على
github 0.2.1»*.

## The root cause: (a) a publish step that exists and was not run

**Nothing on the install path is broken, and every link is already pinned by a
test.** Traced end to end:

| link | where |
|---|---|
| the host is granted | `extension/manifest.json:39` — `https://raw.githubusercontent.com/muhammadbayoumi/mbiX-hub/*` |
| the panel fetches | `extension/app.js:14` imports `latestEngineRelease`; called at `:3755` and `:3777` |
| the URL | `extension/releases.js:32` — `.../mbiX-hub/main/ScrapeX/json/version.json` |
| the field read | `extension/releases.js:83` — `body.version` |
| what is printed | `extension/app.js:3514` — `latest.state === "ok" ? latest.version` |
| the field written | `.github/workflows/release-engine.yml:344` — `"version": VERSION` |
| writer and reader already pinned together | `tests/test_the_two_release_paths.py:276` |

So it is **not** (c) — the panel is not reading the wrong field. And it is not
(b) — the publish step exists: `release-engine.yml` builds, checks, publishes to
`mbiX-hub` and rewrites the manifest last (`:441`–`:462`).

**It is (a).** The workflow triggers on `engine-v*` tags and nothing else
(`release-engine.yml:39-42`), and:

```
$ git tag -l "engine-v*"
engine-v0.2.1
```

`VERSION` moved to 0.2.2 at `adf31b2` (2026-08-10) and to 0.3.0 at `e963269`
(#247, 2026-08-22). Neither was tagged. The manifest says 0.2.1 because 0.2.1 is
the only release that has ever existed. That is `OP-32`, open since 2026-08-21,
and **cutting the release is his** (PLATFORM-PLAN Decision 4, `R-37`/`R-42`).

*The other two rows are a different fact and not a version-publishing defect:
`extension/app.js:3484` prints "Not detected" when `state.engineVersion` is
empty, and `:3472` returns "Not available" for the protocol on the same
condition. The panel never reached a running engine.*

## What this PR fixes — the way out had gone stale

`.github/workflows/release-engine.yml` opens with `test "$tag" = "$version"`
(`:111`), so a tag that disagrees with `VERSION` is refused **before anything is
built**. Six places named `engine-v0.2.2` — line numbers as they were **before**
this PR:

| where | shape |
|---|---|
| `docs/STATE.md:248` | the sentence telling him to cut it |
| `docs/STATE.md:249` | the whole command to copy |
| `docs/REQUESTS.md:1217` | the sentence |
| `docs/BACKLOG.md:1124` | the whole command to copy |
| `docs/BACKLOG.md:1427` | a note about a past failure — history, reworded not renumbered |
| `docs/BACKLOG.md:1499` | the sentence |

Nothing compared any of them with the source, so the moment `VERSION` moved, the
repository was telling him to run a release that would fail at step one — in the
registers **C1** sends every session to read first.

## The guard

`tests/test_the_release_the_documents_ask_for_is_the_one_that_would_run.py`:
**an engine tag named as an INSTRUCTION must equal `VERSION`.** Two shapes are
matched, because they are the two a person acts on — `git tag engine-v…` and
`cut engine-v…`. A tag in narrative prose is deliberately left alone: `engine-v0.2.1`
shipped the black window and every sentence saying so is true for ever, so a guard
demanding *every* tag equal `VERSION` would force history to be rewritten to stay
green.

**It found all six on the untouched documents before a single mutation was tried.**

### Why it does not fetch the hub, and does not ask git for tags

The published version lives on the hub, and the released set lives in the tag
list. Both were rejected, for the reason this repository has already paid for
twice:

* `actions/checkout@v4` fetches **no tags**, so `git tag -l` in the docs tier
  would compare against an empty set and pass over anything.
* A network fetch is red on a train, red behind a proxy, and **vacuously green**
  wherever it is skipped — the failure recorded in `docs/LESSONS.md` under *A
  skip is not a failure* and in
  `tests/test_the_workflows_check_out_enough_history.py`.

So the guard reads two committed files and compares two strings. It **cannot**
prove a release was cut, and says so in its own docstring. `Q-16` in
`docs/BACKLOG.md` puts that question to the owner with three shapes and no
recommendation, because the answer depends on how often he intends to release.

### What has to change in `mbiX-hub`, and nothing here can do it

Nothing, by hand. `ScrapeX/json/version.json` there is **written by the release
workflow**, not maintained: pushing `engine-v0.3.0` publishes the release, sends
the two store documents, and rewrites the manifest last. One command:

```
git tag engine-v0.3.0 && git push origin engine-v0.3.0
```

The "must be newer than what is published" check passes — 0.3.0 > 0.2.1 — and the
publishing job needs `PUBLIC_SITE_TOKEN` as a repository secret. Whether it is set
is not visible from here; the workflow names its absence rather than crashing on it
(`release-engine.yml:394-401`), so a first attempt reports that in one sentence
rather than as a permissions error.

### `MINIMUM_EXTENSION_VERSION` — the constraint, checked

`R-07`: the floor is *"a fact the engine owns"* and it **stays**; the advert goes.
So the question is whether cutting 0.3.0 would advertise a floor above the
extension people actually have. Measured: `MINIMUM_EXTENSION_VERSION` derives to
**0.2.2** and `extension/manifest.json` reads **0.2.2** — equal, so the release
demands nothing new. That is already guarded by
`tests/test_version.py:474` (`test_an_engine_release_does_not_raise_the_floor_under_the_published_extension`),
which reads the shipped manifest rather than trusting the ledger. Nothing in this
PR touches either number, and `R-07`'s standing prohibition on re-pinning
`extension/manifest.json` to `VERSION` is untouched.

## Mutation — eight killed, and the first pass was thrown away

| | |
|---|---|
| M1 | the state document names the dead tag again → **KILLED** |
| M2 | `VERSION` moves and no document follows → **KILLED** |
| M3 | the workflow stops refusing a tag that disagrees with `VERSION` → **KILLED** |
| M4 | the pattern stops reading the sentence shape → **KILLED** |
| M5 | the pattern widens to every tag name, history included → **KILLED** |
| M6 | the two documents carrying four of the six copies leave the scope → **KILLED** |
| M7 | a newly pinned citation drifts off its subject → **KILLED** |
| M8 | the workflow's own header instruction goes stale → **KILLED** |

Control green before, **post-control green after** — and that last word is the
finding, because the first pass was not. Every restore wrote the original bytes
back and was verified by re-hashing the file on disk; all eight reported
`restored=True`; `git status` was clean. And the post-control run was **red**:

```
grep '^VERSION = ' scrapex/version.py                      ->  VERSION = "0.3.0"
python -c "import scrapex.version as v; print(v.VERSION)"  ->  0.4.0
```

The `.pyc` compiled while M2 was in place was still being trusted — CPython
invalidates on the source's mtime and size, and a same-length restore inside one
timestamp tick changes neither. Every run after M2 carried an extra failing test,
and M8's kill criterion was that same test, so it could have reported KILLED over
a mutation that did nothing. The harness now purges `__pycache__` between runs and
the table above is from the clean pass. Recorded in `docs/LESSONS.md` beside the
existing *`restored: True` can be true while the file on disk has changed*.

## Documents (C2)

* **`docs/STATE.md`** — the header names the real `main` (`4615a14`, #250) and the
  unreleased engine; *STILL HIS* item 2 now names `engine-v0.3.0` and says the
  number is derived; item 3 and the dead `cd .claude/worktrees/determined-liskov-0c89fe`
  instruction are struck (that worktree is gone — checked).
* **`docs/BACKLOG.md`** — `OP-32` carries the second report, the corrected next
  action, and the guard's stated limit; `OP-37`'s and `OP-39`'s stale tag
  references corrected; **`Q-16`** added.
* **`docs/REQUESTS.md`** — `REQ-28` records the second report and the corrected
  number; its board row follows; the closed `OP-37` blocker is struck.
* **`docs/LESSONS.md`** — two entries: the instruction that rots on the next bump
  (§7), and the `__pycache__` trap (§8).
* **`tests/test_the_documents_cite_what_they_claim.py`** — five `PINNED` rows for
  the citations added here, including `scrapex/version.py:76`, the citation that
  stayed correct while the value on the line changed under it.

**No register drift:** he asked → `REQUESTS.md` (`REQ-28`, second report); we
found it → `BACKLOG.md` (`OP-32`, `Q-16`); no decision was taken, so
`RULINGS.md` is untouched.

## Verification

* `SCRAPEX_FULL_MIGRATIONS=1 python -m pytest -q` — green (`R-22`).
* `ruff check scrapex/` — clean. No file under `scrapex/` is touched by this PR.
* `python -m pytest -m docs` — green.

## Not merged

`R-42`: this is a secondary session. Opened and reported; the merge and the tag
are his.

---

### DIFF
--- a/.github/workflows/release-engine.yml
+++ b/.github/workflows/release-engine.yml
@@ -2,7 +2,15 @@ name: Release the engine
 
 # PATH B of PLATFORM-PLAN §5c. A tag starts it; nothing else does.
 #
-#     git tag engine-v0.3.0 && git push origin engine-v0.3.0
+#     git tag engine-v<VERSION> && git push origin engine-v<VERSION>
+#
+# <VERSION> IS DELIBERATELY NOT SPELLED OUT HERE. It is `scrapex/version.py:VERSION`
+# and nothing else — the first step below refuses any other tag — and this line used
+# to carry the literal, which made it one more copy of a number that moves on every
+# contract change (R-35). It read `0.3.0` on the day 0.3.0 was released and would
+# have been wrong by the next migration. Read it instead of typing it:
+#
+#     python -c "from scrapex.version import VERSION; print(VERSION)"
 #
 # Decision 4: releasing is manual, updating is not. This is the manual half —
 # the owner decides when. The extension notices the result on its own.
--- a/docs/BACKLOG.md
+++ b/docs/BACKLOG.md
@@ -1203,9 +1203,11 @@ many workers are asked for, and the cost is bounded by arithmetic rather than by
 hoping. (a) as an immediate mitigation.*
 **Not started.**
 
-### OP-32 · The fix for the black window has sat in the repository for twelve days, unreleased
+### OP-32 · ~~The fix for the black window has sat in the repository, unreleased~~ — RELEASED 2026-08-22 as `engine-v0.3.0`
 
-**Status: OPEN. The gate that let it happen is closed in this PR; publishing is his.**
+**Status: CLOSED 2026-08-22 — the gate that let it happen was closed first, then he
+published.** The entry is kept in full rather than trimmed: it is the only record of
+why an install path where every component worked still handed him a black window.
 Reported by the owner on 2026-08-21 as *"it did not install — black screen"*, and
 captured as [REQ-28](REQUESTS.md#req-28--the-engine-would-not-install-and-showed-a-black-screen).
 
@@ -1245,9 +1247,33 @@ this guard accepting a data root that was named but never assigned.
 0.2.1"* — and filed it as a **wording** defect about the two meanings of
 "installed". The numbers were the finding: the release had not happened.
 
-**Next action, and it is his:** `git tag engine-v0.3.0 && git push origin engine-v0.3.0`.
-Nothing needs bumping — `scrapex/version.py:76` already carries the number, and the
-tag is derived from it rather than chosen.
+~~**Next action, and it is his.**~~ **— HE TOOK IT. `engine-v0.3.0` IS PUBLISHED,
+2026-08-22.** *«اقطع الوسم»*, said after reading this entry, so the finding is what
+produced the release. Recorded as facts rather than as a green workflow run:
+
+| | |
+|---|---|
+| tag | `engine-v0.3.0` → `451468d`, which is `main` |
+| the version at that tag | `scrapex/version.py:76` **0.3.0**, `pyproject.toml` mirror **0.3.0** — so the workflow's `test "$tag" = "$version"` had two numbers that agreed |
+| workflow | completed, 28m36s |
+| the manifest the panel reads | `"version": "0.3.0"`, `"tag": "engine-v0.3.0"`, `"published_at": "2026-08-22T13:17:13Z"`, `"minimum_extension_version": "0.2.2"`, protocol 1 |
+| what it said before | `0.2.1`, unchanged since 9 August |
+
+**The floor it published is 0.2.2, and that is the number that matters most here:**
+`extension/manifest.json` is also 0.2.2, so this release demands nothing the
+installed extension cannot do — checked before the tag by
+`tests/test_version.py`'s floor guard, not after it.
+
+**This entry is CLOSED as the defect it described.** What is not closed is the
+confirmation that the published build installs on his machine; that stays on
+`REQ-28`.
+
+> **AND THE VERSION GAP THAT OPENS NEXT IS NOT THIS DEFECT RETURNING.** Migration
+> `0010` is a contract change, so `R-35` moved the source to **0.3.1** against a
+> published **0.3.0** — which is the ordinary state between two hand-cut releases,
+> not a fault. `OP-32` was three things at once: nothing released across two bumps,
+> the newest installable engine silent on a double-click, and the documents naming a
+> tag the workflow would refuse. A gap alone is none of them.
 
 > **STILL OPEN A DAY LATER, AND THE INSTRUCTION HAD GONE STALE WHILE IT WAITED.**
 > He reported it again on 2026-08-22 — *«المحرك الموجود على github 0.2.1»* — with the
@@ -1256,7 +1282,7 @@ tag is derived from it rather than chosen.
 > written about it.** `extension/releases.js:32` reads
 > `ScrapeX/json/version.json` from the hub, `extension/app.js:3514` prints the
 > `version` it finds, the workflow writes that same field
-> (`.github/workflows/release-engine.yml:344`), and
+> (`.github/workflows/release-engine.yml:352`), and
 > `tests/test_the_two_release_paths.py:276` already pins the writer's output to the
 > reader's input. The manifest says 0.2.1 because 0.2.1 is the only engine tag that
 > exists.
@@ -1655,10 +1681,11 @@ process that waits for a named pid to exit, for exactly this reason. **This is
 why `OP-36` had to be fixed first** — before that, the helper a frozen engine
 spawned was a silent native messaging host and would have waited for ever.
 
-**Next action, and it belongs after a release rather than before one:** cut
-`engine-v0.3.0` (`OP-32` — the number comes from `scrapex/version.py:76`, never from
-a document), then implement the swap against a binary that exists. The order is not
-a preference — the first artifact that can exercise any of this is the next release.
+**Next action, and its precondition is now met.** This belonged after a release
+rather than before one, and the release happened on 2026-08-22 (`OP-32`): a published
+`engine-v0.3.0` is the first frozen artifact that can exercise any of this. So the
+swap can now be implemented against a binary that exists rather than against
+`sys.frozen` set by a test.
 
 ### OP-40 · His warehouse has moved ahead of `main` again — v9 now, and this is the third time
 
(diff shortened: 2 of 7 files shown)


muhammadbayoumi:
## Rebased onto `5f63bb0`; the citation repair is **dropped** at the primary session's request

Rebased `fix/the-release-the-documents-ask-for` onto `origin/main` (`5f63bb0`,
after #251 and #252). Force-pushed with `--force-with-lease`.

**The `2710 → 2725` repair is no longer in this PR.** It lands on
`feat/the-profile-page-becomes-columns`, which already carries it committed. Two
branches editing the `PINNED` table at the same place is a guaranteed conflict for
whichever merges second, and only one needed to hold it. `tests/test_the_documents_cite_what_they_claim.py`
and `docs/BACKLOG.md`'s prose citation are back at main's content, byte for byte.

What this PR keeps is the **class**, in `docs/LESSONS.md` §7 — written to stand on
its own, and saying plainly that the fix landed elsewhere.

### The three collisions predicted, and a fourth

| | |
|---|---|
| **`docs/REQUESTS.md`** — REQ ids | No collision. I claim **no** REQ id; `REQ-28` is an existing entry I extended, so `REQ-30` is untouched. Auto-merged. |
| **`docs/BACKLOG.md`** — OP/Q ids | No collision, **no hole**: `OP-1`–`OP-42` contiguous (`OP-42` intact), `Q-1`–`Q-16` contiguous with my `Q-16` last. Auto-merged. |
| **`docs/STATE.md`** | **Conflicted.** Took main's rewritten paragraph, **corrected the pointer to `5f63bb0`** rather than restoring `4615a14`, and re-applied my engine-release paragraph beneath it. |
| **`tests/test_the_documents_cite_what_they_claim.py`** *(not predicted)* | **Conflicted** — #252 and I both appended to `PINNED`. Both blocks kept, main's first. |

`tests/test_the_registers_cannot_collide.py` run rather than eyeballed: **green.**

> Main's new header says *"a commit pointer written into prose is stale by the time
> it is read"* — and it was, within the hour, reading `4615a14` with #251 and #252
> already in. Corrected and the sentence kept, since it is now evidence for itself.

## Both corrections taken, and they change what the lesson says

**1. #252 was not merged untested.** It reported MERGEABLE/CLEAN with every check
passing — against #252's own merge commit **against `4615a14`**, a base that stopped
existing when #251 landed. So the suite did not fail to run and did not fail to
notice: it **passed, truthfully, about a `main` that no longer existed.** "Green"
without a base is not a claim about anything. My first account said "no test either
could have run to see it" — the second half of that is right, the framing was wrong,
and the corrected version is what is committed.

**2. Disjoint in files, coupled in content** — now the entry's title and its point.
*"Check whether the files overlap"* is the reflex that fails here: git finds nothing
to conflict on and merges both cleanly, while the coupling sits entirely in the
content. One moved a line; the other wrote that line's number down.

Both feed the same conclusion, which `docs/STATE.md`'s branch-protection item now
states: **`require branches up to date` is the clause that catches this, and
`require the check suite` on its own would have passed the pair through** — because
both checks genuinely passed. Second reason for that setting after `ac3a5af`.

## R-22, stated rather than hidden

**This branch's suite has one failure, and it is main's:**

```
FAILED tests/test_the_documents_cite_what_they_claim.py::
  test_a_pinned_citation_still_points_at_its_subject[BACKLOG.md-app.py-2710]
```

Inherited, proved from git objects without running anything:

```
$ git show 5f63bb0:tests/test_the_documents_cite_what_they_claim.py | grep -n "source_key not in known"
202:    ("docs/BACKLOG.md", "scrapex/webui/app.py", 2710, "if source_key not in known:"),
$ git show 5f63bb0:scrapex/webui/app.py | grep -n "if source_key not in known:"
1831:        if source_key not in known:
2725:            if source_key not in known:
```

Dropping the repair is what re-introduces it, deliberately and on instruction. It
clears the moment `feat/the-profile-page-becomes-columns` is on `main` — which is
the merge order already set — so **R-22 is satisfied by that order, not by this
commit**, and I am not claiming a green suite that this tree does not have.

Everything else is green: **302 passed, 1 skipped** in the docs tier, the one
failure above, and nothing else in the full run (recorded below). `ruff check
scrapex/` clean — **no file under `scrapex/` is touched by this PR.**

## Still not merged

`R-42`. Yours to take, in your order: parser branch first, then this.

muhammadbayoumi:
## Rebased onto `bcb8f6e` (#255) — green, and the inherited failure is gone

Head `e1f7fdd`, `MERGEABLE`, no conflicts outstanding.

### Full suite, verified non-vacuously

Run **bare** so the reported code is pytest's own rather than a trailing `echo`'s:

```
SCRAPEX_FULL_MIGRATIONS=1 python -m pytest -q     ->  exit 0
```

| check | value |
|---|---|
| output file | 4249 bytes, 52 lines — not empty |
| reached | `[100%]` at line 45 |
| `grep -c '^FAILED'` | **0** |
| `grep -c '^ERROR'` | **0** |
| migrations really ran | `schema template: DISABLED (SCRAPEX_FULL_MIGRATIONS) — every migrate() ran for real` |

The `[BACKLOG.md-app.py-2710]` failure this branch carried on the previous base is
gone: `main` now holds `2787`, re-derived by #255, and this branch never touched that
row.

Docs tier **310 passed, 1 skipped**. `ruff check scrapex/` clean — **no file under
`scrapex/` is touched by this PR.**

### Re-derived, not carried

Every line number this branch cites, read from the file at `bcb8f6e`:

| citation | at `bcb8f6e` |
|---|---|
| `extension/releases.js` | 32 |
| `extension/app.js` | 3514 |
| `.github/workflows/release-engine.yml` | 352 |
| `tests/test_the_two_release_paths.py` | 276 |
| `scrapex/version.py` | 76 |
| `tests/test_version.py` | 73 |

**I cite nothing in `scrapex/webui/app.py`**, so the `2710 → 2725 → 2787` walk is
entirely main's row and is current. #255 touched `scrapex/webui/app.py`;
`extension/app.js` is a different file, so `3514` was never exposed to it.

`release-engine.yml:352` is the one that moved, and it moved **because of this PR** —
the header edit added eight comment lines above it. That is `LESSONS.md` §7's exact
mechanism catching its own author inside the pull request that added it. Re-derived
rather than nudged, and the `PINNED` row records why it is not 344.

### Structural checks

* **`LESSONS.md` sections are 1–12, each exactly once.** No duplicate §12 — this
  branch adds no section, only entries inside the existing §7 and §8.
* **No `RESERVED` row here.** All four dicts are empty and neither `OP-44` nor
  `REQ-33` is named in this branch, so there is no reservation to retire.
* **Registers:** `OP` max 44, `REQ` max 33 — matching main. `Q-16` is the only id
  this branch holds, and `Q-1…Q-18` is contiguous with it.

### One correction worth recording about `Q`

`Q` **is not a tracked register.** `REGISTERS` in
`tests/test_the_registers_cannot_collide.py` covers `R`, `REQ`, `OP` and `DEC` only,
and its pattern is heading-anchored (`^## Q-…`) while every `Q` entry is written as
bold text. So main's hole at `Q-16` could not have failed anything and `RESERVED` was
never in play. The guard is green because it does not look.

Deliberately **not** fixed here: the `C4` struck-and-kept form (`~~Q-17~~` beside
`**Q-17 · …**`) would read as a duplicate under the existing pattern, so extending it
needs its own decision rather than a drive-by in this PR.

### Mutation, re-run on this base

Ten mutations: **eight killed, two controls stayed green**, every restore
hash-verified, `__pycache__` purged between runs, post-control green.

The two controls are the point, not padding. `M2` injects a *correct* instruction and
must stay green; `M3` moves `VERSION` and must **also** stay green, because with no
release owed there is nothing to compare. That is why the v1 set was discarded rather
than re-run — "move `VERSION` and watch it fail" cannot kill anything once every
instruction has correctly become history.

`M3` earned itself earlier: it caught the §7 paragraph quoting the bad sentence
verbatim, so the entry explaining the trap was holding an instruction. Found by
mutation, not by reading.

### The `cut` rule is now where prose gets written

`§7` carries it as a set-off block: a release that has happened is written **tagged /
published / released / went out**, never `cut`, and never with `git tag …` spelled
beside it — those are the two shapes the guard reads as work still owed. Stated with
the reason it is four words rather than a cleverer regex: the wrong word costs one
rewording, the wrong silence costs a release instruction the workflow refuses at its
first step.

### Not merged

`R-42`. Yours to take — this branch is first in the queue.

PR #278, opened 2026-08-27
Four of the five muqawil blockers were ruled, and the live plan never said so

muhammadbayoumi:
`R-58` makes `2026-08-26-what-remains-of-muqawil.md` the one live plan, and its status table
marked steps 1–5 as **⛔ his ruling**. He ruled four of them the next day as `R-53`–`R-56`, and
the table never said so — **so the first document a session is sent to told it to stop on four
decisions that already existed.**

| step | now |
|---|---|
| 1 · the poisoned profile schema | **`R-53`** — re-approve all 17,371 rows onto a clean 27-field v4 |
| 2 · the State column | **`R-54`** — the root first: a confirming pass moves `last_seen_at`, then state is computed against the RUN |
| 3 · the 263 stranded rows | **`R-56`** — a fresh listing crawl, 58 min at concurrency 4, and **not** a re-approve |
| 5 · the placeholder pair | **`R-55`** — one ruling for both, store absence |
| **4 · the 148 + 81 gap** | **⛔ still his** (`REQ-41`), now marked as **the only one** |

**Every mapping was checked against the step's own GATE text rather than inferred from the
numbers.** Step 2 is the clearest: its gate reads *"a confirming pass moves `last_seen_at` on
the record"*, which is `R-54`'s choice word for word.

The headline section is **struck rather than deleted** (`C4`) — its reasoning about why those
five were the blocking ones is what got them ruled, and that survives being wrong about the
count.

Four steps are buildable now; one waits on him.


muhammadbayoumi:
Closing this on the owner's decision of 2026-08-27, and recording the measurement so it is not
re-opened by guesswork.

**Its substance is already on `main`.** `R-68`, `R-69`, the plan's status table with zero
blocked rows, and the `REQUESTS.md` entries all reached `main` through later pull requests —
`#279` carried the plan file among them. Nothing here is lost by closing it.

**And merging it now would regress eight citations.** The only difference left between this
branch and `main` is the line numbers into `scrapex/extract/service.py`. A twenty-line comment
added in `#279` moved them, and they were repointed with `difflib` in that same pull request.
This branch still holds the pre-`#279` numbers — line 895 where `main` says 915, and seven
more at the same twenty-line offset.

That is the harmful case rather than a cosmetic one: `tests/test_the_documents_cite_what_they_claim.py`
would either fail on them or, worse, accept a number that resolves to the wrong line and send
a reader somewhere that no longer says what the document claims. Three citations had already
drifted silently before that guard existed, which is why it exists.

**The branch stays on the remote**, so this is reversible with one click if any of it turns out
to be needed. Recorded in `docs/STATE.md`'s "Open pull requests" banner and in `OP-89`.

PR #319, opened 2026-09-03
The panel tells him to open a terminal, and promises an updater that nothing starts (OP-124, R-81)

muhammadbayoumi:
Two sentences on his only interface that are not true, the guard that let one of
them stand, and the thing that guard could never have caught.

He does not use a terminal ([R-81](docs/RULINGS.md)). The panel is his only
interface. So a screen that prints a shell command is not an instruction, and a
screen that describes a capability nothing provides is worse than silence.

## 1 · The banner names a terminal command, and it is written twice

```
scrapex/webui/app.py   "fix": "python -m scrapex.cli init-db"     the engine sends it
extension/app.js       "Fix: " + (lag.fix || "python -m ...")     the panel hardcodes
                                                                  it AGAIN, as a fallback
extension/app.js       POST /api/databases/upgrade                the button that works
```

**A button that performs it already exists on the Settings screen.** The
duplication is why this is `OP-124` and not a one-line fix: one string in two
files with nothing comparing them, so correcting the engine alone would have
left the panel printing it from its own copy.

**This banner had never appeared before `#293`.** `OP-113` is why — `schema_lag`
was published, rendered, and never carried — so the first thing that repair did
was put a terminal command in front of him. **A fix that makes a surface visible
inherits everything wrong with it**, and nothing in the change that carried the
field would have said so.

**Checked before pointing the banner anywhere:** `POST /api/databases/upgrade`
is mounted only `if databases is not None`, so the remedy is itself absent on a
`--db` start. The banner names the screen and the button rather than a route,
which is true on every start. That check is `OP-119` applied inside the fix for
something else.

## 2 · And the installer note promises an updater that nothing has ever started

The note beside the download's SHA-256 says, in `#246`'s words and still on
`main`:

> *"From here on the Engine downloads and checks its own updates against this
> number before installing."*

**It is false, and it was false the day it was written.** Measured across
`origin/main`:

| what exists | what calls it |
|---|---|
| `create_update_router()` — mounted unconditionally, `GET` and `POST` | **nothing** |
| `update.fetch_and_verify()`, `plan_swap()`, `swap_is_possible()` | only `update_api` |
| any `api/update` string in `extension/**/*.js`, `extension/**/*.html`, or the engine's templates | **none, in the whole history** — `git log -S` over those paths returns no commit |
| a timer or startup task that checks by itself | **none** — nothing imports `update` except the router |

**So the updater is built, mounted, and doorless**, and the only sentence
describing it claims it is running. `R-81` says a capability with no control in
the panel has no control; this is that one step further on, because a screen
asserts the control exists.

**The doorless updater is recorded, not repaired** — `OP-124` names it and says
why: the fix is either a caller or a deletion, and that is the owner's call.
This PR only stops the screen claiming otherwise.

## 3 · And the guard over that sentence could not tell a promise from its denial

`test_the_checksum_on_screen_says_who_checks_it` asserted `"Engine" in text`.
Its docstring states the rule exactly right — *"a number nothing verifies is
worse than none, it reads as a promise"* — and then checked for a **word**.

> *"The Engine checks its own updates"* and *"the Engine does not fetch or
> install its own updates"* both contain it.

**It stayed green across the entire period its subject was false**, and it would
have gone red on the honest sentence had the honest sentence omitted the word.

**Now the claim is tied to the code**: the note may say the engine verifies
updates **only while something actually calls `/api/update`**, and it fails in
**both** directions.

| mutation | result |
|---|---|
| restore the promise | **red** — *claims engine-side verification while nothing calls /api/update* |
| build the caller | **red** — *the panel now calls it, so the note may say so, and should* |

A guard that can only fail in one direction is half a guard.

## `LESSONS` §29 goes from eleven rows to thirteen

Rows 12 and 13 are two kinds the taxonomy did not have, both from the primary
session's `OP-126`:

- **`assert 400 <= status < 500`** cannot separate *"wrong source key"* from
  *"not a RunMode"*, and a status-code range is not a sloppy assertion — it is a
  deliberately general one. **When an assertion is imprecise by design, the
  design is the defect.**
- **A guard that asserts a capability is MISSING** holds the missing-ness in
  place: it converts a defect into a requirement, and the only change that can
  turn it red is the fix. Three instances now, so it is a shape.

Plus a seventh true-number-about-the-wrong-subject: **a line count reported as a
row count**, which happened inside the verification of this section's own
repair. Its author caught and corrected it unprompted; the words are theirs.

## Register

`OP-124` is declared, and its reservation on `main` retired — the **seventh**
this session has retired, and every one was named by
`test_a_reserved_number_is_not_also_declared` before the session noticed.

## Gates

Rebased onto `main` across sixteen merges. Every citation the rebase moved was
re-derived **by content**, not by a delta — fourteen of them, by running the
guard until it went silent — and one was corrected a second time because the
loop had taken the **first** of two matches when the entry described the second.

**Local gates were still running when this was opened, at the owner's
instruction. CI is the verdict here; I will post the settled local reads as a
comment rather than claim a green I have not seen.**

### DIFF
--- a/docs/APPROACHES.md
+++ b/docs/APPROACHES.md
@@ -162,7 +162,7 @@ standing next to — do not tidy it, do not restyle it, mention dead code rather
 than deleting it. P1 governs *the thing you are building*.
 
 The live example is B2 step 2: the panel already has Choose-Columns
-(`extension/app.js:1578`, `:1617`), and the instruction is to **extract it into a
+(`extension/app.js:1583`, `:1617`), and the instruction is to **extract it into a
 shared module**, not to write a second one. Writing a duplicate in the name of
 staying surgical is precisely the failure the migration plan warns about — "or
 the two surfaces will disagree about how a column is saved."
--- a/docs/BACKLOG.md
+++ b/docs/BACKLOG.md
@@ -556,8 +556,8 @@ And only one of the **two** `worker_alive` computations was fixed:
 
 | where | what it calls | verdict |
 |---|---|---|
-| `scrapex/webui/app.py:1726` — `/api/health` | the new two-heartbeat `worker` verdict | correct; the panel reads this one (`extension/engine.js:38`) |
-| `scrapex/webui/app.py:2796` — `_about` | `worker_is_alive(conn)`, single heartbeat | **the function the fix superseded** |
+| `scrapex/webui/app.py:1737` — `/api/health` | the new two-heartbeat `worker` verdict | correct; the panel reads this one (`extension/engine.js:38`) |
+| `scrapex/webui/app.py:2807` — `_about` | `worker_is_alive(conn)`, single heartbeat | **the function the fix superseded** |
 
 `_about` renders the engine's own `/settings` page
 (`scrapex/webui/templates/settings.html:162-167`), so **the engine still shows
@@ -565,7 +565,7 @@ And only one of the **two** `worker_alive` computations was fixed:
 engine is started at all.
 
 **Next action:** three separate things — the second `worker_alive` at
-`app.py:2796`; the heartbeat's behaviour under a held write lock; and the 409 on
+`app.py:2807`; the heartbeat's behaviour under a held write lock; and the 409 on
 `/api/storage/restore` with a mirror of
 `test_start_fresh_is_refused_while_a_crawl_runs`.
 
@@ -1415,7 +1415,7 @@ exactly one place — an unmerged worktree branch that had already been run agai
 warehouse — and `R-24` forbids the obvious shortcut of replacing the database.
 
 **What is NOT closed is the panel's sentence.** An engine that refuses to start never binds a
-port, so `extension/app.js:3400` reports **"Not detected"** for a schema fault, a permissions
+port, so `extension/app.js:3405` reports **"Not detected"** for a schema fault, a permissions
 fault and an absent engine alike — false, and it sends the reader to reinstall what is already
 installed. That half is `OP-38`.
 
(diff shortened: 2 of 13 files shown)


muhammadbayoumi:
**Settled local gate reads, as promised when this was opened. Three green, one red — and the red is `main`'s, proven, not this branch's.**

| gate | result |
|---|---|
| `pytest -q` | `exit=0 on 64292f20`, 0 failed |
| `node --test` | `exit=0 on 64292f20`, 421/421 |
| `ruff check scrapex/` | `exit=0 on 64292f20` |
| `SCRAPEX_FULL_MIGRATIONS=1 pytest -q` | **`exit=1 on 64292f20`, 1 failed** |

The failure is `tests/test_the_engine_survives_being_killed.py::test_a_killed_engine_does_not_leave_a_job_claiming_to_run` — *"after a crash and a restart the job still says 'running'"*.

**It is not this branch's, and it was measured rather than assumed.** Run against `origin/main` in a throwaway worktree with `PYTHONPATH` pinned to that worktree, so the import resolves to main's own source and not to the editable install in another checkout:

```
origin/main, SCRAPEX_FULL_MIGRATIONS=1   FAILED   (same test, same assertion)
origin/main, default mode                 2 passed
```

**So `main` is red in the full-migrations mode right now, and green in the default one.** This branch touches the schema-lag banner string, the installer note, one guard, and documents; nothing it changes is read by that test.

## Why nobody saw it

`.github/workflows/ci.yml` gates the full-migrations job on `needs.scope.outputs.scope == 'full'`, and scope is computed from the diff: `docs` when every changed path is documentation, `extension` when every path is `extension/` or documentation, `full` otherwise.

**So a defect can land through a narrow-scope pull request and stay invisible until the next wide one runs the gate — where it appears as that pull request's failure.** `#308` and `#311` were documentation-only and never ran this mode. `#319` touches `scrapex/webui/app.py`, so it does, and it inherits the red.

That is not an argument against the scope gate, which was measured and is worth its saving. It is an argument for knowing that **the first full-scope PR after a narrow one carries someone else's verdict**, and for reading it that way rather than blaming the branch in front of you.

I have not fixed it here. It is a separate defect in a file this branch does not touch, and folding it in would put an unrelated repair under this title.

PR #326, opened 2026-09-04
The panel's source list is the price manifest, so muqawil cannot be in it (OP-145)

muhammadbayoumi:
> «لماذا مصدرمقاول لا يظهر ضمن المصادر المتاح زحفها!»

**The first answer I gave him was wrong**, and the measurement corrected it. I read it as a consequence of the warehouse created half an hour earlier — the row would have been in the old one. It is not that.

## `/api/sources` never looks at the warehouse for *which* sources exist

```
scrapex/webui/app.py:1836    for entry in app.state.manifest.sources:
                                 s = summaries.get(entry.source_key)   ← the DB only decorates
```

`sources.yaml` declares **twelve** — `MADAR`, `ALSWEED`, `ELBUROJ`, `ADVANCEDCASTLE`, `ELSEWEDYSHOP`, `MASDAR`, `SIKAEGSHOP`, `HEIDELBERG_EG`, `SAMEHGABRIEL`, `GPP_ENERGY`, `ARAMCO_FUEL_SA`, `SPARK_ESHOP` — and muqawil is not one of them. **The code already says so**, in the module written for this very question:

```
scrapex/sourceboard.py:13   source_site, declared nowhere else   muqawil_org  <- not in sources.yaml at all
```

Measured against his running 0.4.8: `/api/sources` answers **12** rows, `muqawil in the list: False`. So it could never have appeared, on any warehouse.

## And the module that answers it has one caller, the CLI

`sourceboard` opens by saying it exists because he asked *«اى الجديد واى الى خلص»* and no command could answer. It spans both registries, carries the four-state vocabulary (`registered` / `built` / `active` / `paused`), and works with no database at all.

```
scrapex/cli.py:527          from . import sourceboard
```

**That is its only caller.** Nothing in `webui/app.py` mentions it. `R-81`: a capability with no control in the panel has no control — and this is **the day's fourth instance of one shape**: a retention policy with no caller (`OP-136`), an updater with no door (`OP-124`), a remedy behind a page the engine cannot serve (`OP-144`), and now an answer with no route.

## A second, independent gap

Even with a route, a fresh installation shows nothing: the squashed baseline's seed carries `retention_policy` (1 row) and `scrapex_meta` (2) and **no sources**. The muqawil row is created by `catalog.register_site` from inside the collector's own first run (`scrapex/contractors.py:287`) — a by-product of `scrapex contractors`, a command he does not use. Measured on today's warehouse: `source_site` holds **one** row, `ARAMCO_FUEL_SA`.

## What it would take — not built here

A route over `sourceboard` and the panel's list drawn from it. The module exists, is tested (`tests/test_every_source_appears_on_one_board.py`), and needs no schema change. Declaring muqawil in `sources.yaml` is the wrong half of the fork: that manifest is a *price* declaration (family, currency, tax mode) and a contractor directory has none of those, which is why validation would refuse it. It composes with `REQ-54`.

**This is `R-32` at the level of the interface** — he ruled that price is one category and that filing the tool under it was a mistake; the panel's source list is still that filing. Changing what his only interface lists is his to approve.

## Note on `OP-144`

It is declared as a **reserved hole**: committed on `claude/the-tag-follows-the-version` at `ad8b1d8`, which cannot be pushed until `gh`'s token carries the `workflow` scope. That is ORCHESTRATION.md §3's "invisible and still real" case, and the row says so rather than letting the gap read as a skipped number.

`pytest -m docs`: **376 passed, 2 skipped**.

### DIFF
--- a/docs/BACKLOG.md
+++ b/docs/BACKLOG.md
@@ -5317,6 +5317,79 @@ which un-orphaned them without fixing anything, and the guard said so.
 `def store` two lines above it had moved thirteen lines. The block cited both. Only a
 content check can tell a correct number from a coincidence.
 
+### OP-145 · The panel's source list is the PRICE manifest, so the contractor category cannot appear in it
+
+**Asked by him 2026-09-04** — *«لماذا مصدرمقاول لا يظهر ضمن المصادر المتاح زحفها!»* — half
+an hour after a fresh warehouse was created on his machine.
+
+**THE FIRST ANSWER WAS WRONG AND THE MEASUREMENT CORRECTED IT.** It read as a consequence
+of the new database: the row would have been in the old one. It is not that. `muqawil_org`
+**cannot appear in that list on any warehouse, new or old**, because the list is not built
+from the warehouse at all:
+
+```cited
+scrapex/webui/app.py:1836           for entry in app.state.manifest.sources:
+```
+
+`/api/sources` iterates the MANIFEST and uses the database only to decorate those rows
+(`summaries.get(entry.source_key)`). `sources.yaml` declares **twelve** sources — `MADAR`,
+`ALSWEED`, `ELBUROJ`, `ADVANCEDCASTLE`, `ELSEWEDYSHOP`, `MASDAR`, `SIKAEGSHOP`,
+`HEIDELBERG_EG`, `SAMEHGABRIEL`, `GPP_ENERGY`, `ARAMCO_FUEL_SA`, `SPARK_ESHOP` — and
+muqawil is not one of them. The code already says so:
+
+```cited
+scrapex/sourceboard.py:13       source_site, declared nowhere else       muqawil_org   <- not in sources.yaml at all
+```
+
+Measured against his running engine: `/api/sources` answers **12** rows and
+`muqawil in the list: False`.
+
+**AND THE MODULE WRITTEN FOR EXACTLY THIS QUESTION HAS NO ROUTE.** `sourceboard` opens by
+saying it exists because he asked *«اى الجديد واى الى خلص»* and no command could answer;
+it spans both registries, carries the four-state vocabulary (`registered`, `built`,
+`active`, `paused`), and works with no database at all. Its callers:
+
+```cited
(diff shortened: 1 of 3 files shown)


muhammadbayoumi:
## Correction, and the error was mine

**The headline claim in this PR was false.** I wrote that `/api/sources` walks the price manifest alone, so a warehouse-only source could never appear. The route's **last statement** is:

```
scrapex/webui/app.py:1890        out.extend(_dataset_listing())
```

It appends one card per site holding `dataset_definition` rows, and the comment beside it exists precisely because *"a `source_site` row could never reach the panel however much data it held"*. **His own history disproves my claim twice over**: he complained, with screenshots, that `muqawil.org` appeared **twice** on that screen (`REQ-37`, `R-47`).

**The answer I gave him first was the right one and I talked myself out of it.** It is the fresh warehouse — `_dataset_listing()` groups `dataset_definition`, and the warehouse created at 12:30 today holds **zero** (measured: `generic_record` 0, `source_site` 1, `ARAMCO_FUEL_SA` alone). The old warehouse held two and drew one folded card.

**How it happened, since that is the reusable part:** I measured the route at the loop it *opens* with, got twelve rows and no muqawil, and stopped because the reading fitted. The route is sixty lines long. `LESSONS` §9 names the shape already — *a search for one spelling of a feature is not a measurement of the feature*.

**What survives, and why the number stays.** Both findings under the false headline were independent of it and are unchanged:

1. `sourceboard` — written for his question, spanning both registries, carrying the `registered` state — has **one caller (`cli.py:527`) and no route**. The panel cannot ask the question the module answers.
2. The muqawil **row** is created by `catalog.register_site` inside the collector's first run, so a fresh installation has nothing until `scrapex contractors` has run once — a command he does not use.

`R-32` is re-stated where it actually bites: the list can show a source that has already collected something, and has **no place at all** for one that is registered and waiting its turn — his own «يحفظ فقط فى قائمة مصادر حتى ياتى دوره».

The false claim and the reason for it are kept at the top of the entry under `C5` rather than edited away. `dd87386`.

PR #330, opened 2026-09-04
A source registered in the code is listed before it has data, from one place (P1/Q1)

muhammadbayoumi:
> «المفروض المصادر تظهر بدون بيانات فهى مسجلة ومحفوظة فى الكود»

**This is what brings the muqawil crawl back to the panel.** He reported it gone after today's merges; measured, **none of the six touched a single file of the crawl path** — the control hangs off the dataset card, and the datasets went with the v10 warehouse that `R-84` locked out. So the button did not disappear: the row it lives on stopped existing.

> **Stacked on [#329](https://github.com/muhammadbayoumi/ScrapeX/pull/329)**, whose base is `main`. GitHub retargets this to `main` when that merges.

## What was wrong

`/api/sources` was the price manifest plus `_dataset_listing()`, and that second half groups `dataset_definition` rows — so a directory appeared only **after** its first crawl had stored something, while `directories.BUILDERS` had known how to crawl it all along. `visibleSources()` filters nothing, so **Choose sites was not hiding it: it was never sent.**

**And the row alone would have been a button that cannot work.** Measured: `POST /api/jobs` answered `404 unknown source_key 'muqawil_org'` on a fresh warehouse, because `SourceResolver` asks the manifest and `source_site`, and a directory's `source_site` row is written by `catalog.register_site` **from inside the collector's own first run**. The route now accepts a key `BUILDERS` holds, because the crawl is what writes the row.

## `kind: "directory"`, not `"dataset"`

The difference is one control versus two: the panel gates *"Open the data table"* on the dataset marker, and that reaches `/api/table/{key}` — a 404 for a site with no dataset yet. So this kind is admitted to **the crawl and nothing else**, and the panel's crawl gate now reads `site_key` as its own comment always claimed it did (*"GUARDED ON `site_key` RATHER THAN ON THE KIND"*, while the code also demanded one exact kind).

## He caught the DRY defect before I did

> «هل التعديلات التى قمت بها dry وطبقا للمسجل ؟»

The first version put `BUILDERS` inside `webui/app.py`, so the panel listed muqawil and `sourceboard.board()` — what `scrapex sources` reads — did not. **Two answers to "which sources exist"**, which is `ENGINEERING` P1/Q1. Worse: `OP-145` records *that same morning* that the fix was a route over `sourceboard`, and `sourceboard.py:109` had already recorded the missing state as a finding pointing at `REQ-25`.

So **`sourceboard.from_code()` owns it** and both surfaces read it. `built`, not `registered`, per the module's own vocabulary: a collector exists and nothing is scheduled. The board dedupes against the warehouse, because a crawled directory has a row carrying its real lifecycle — two entries for one source is `REQ-37`'s twice-drawn card in a list instead of on a screen.

## Six guards, two of them because of the mistake

- the panel's key set and the board's must be **identical** on an empty warehouse — the guard against the split returning, and it names the difference if it does;
- a crawled directory is listed **once**, by the warehouse, with its own state (`paused`, not `built`);
- and the key the row carries is a key the job route accepts — the assertion that makes the row more than a label.

Four board tests moved with the third registry, and **two of them had been measuring the manifest by measuring "everything"**, which only worked while there were two.

The last commit repoints **32 citations** my insertions moved, by subject and not by arithmetic, plus nine `PINNED` rows and 19 markdown link targets that had drifted from their own labels. `ruff` and `eslint` clean; docs tier **376 passed**; the panel suite green.


muhammadbayoumi:
Reopened as #593, because this pull request never reached `main`.

Its base was `claude/a-socket-is-not-an-engine`, not `main`. Merging #329 first and this one second squashed it onto a branch that had already been absorbed, so GitHub reports it merged while `48ee301` is not an ancestor of `main` — measured: `grep -c from_code scrapex/sourceboard.py` on main returns 0.

#593 is the same commit cherry-picked onto `main`, with the affected suites green. My mistake, and the reason it was invisible is worth naming: a squash merge onto a stale base reports success in every place a session would look.

Issue #531, opened 2026-09-04
Fix the killed-engine job status on Windows, and let CI see it


muhammadbayoumi:
**From** `docs/plans/2026-08-21-the-tool-itself.md`

**Why now** — The plan's OP-19 entry points at a sweep that exists and works; the live defect is that it does not always win. Issue #461 measured 8 failures in 13 runs after the swept=True marker landed, and issue #401 isolated the same assertion to a clean checkout of main on Windows. ci.yml runs the suite on ubuntu-latest only, so main is green over a red the owner sees on his own machine — and the consequence is the worst failure this product has: a source silently stops being crawled with nothing on any screen saying why.

**Evidence** — tests/test_the_engine_survives_being_killed.py:163-188, :266; scrapex/jobs.py:1149-1195, :1399; .github/workflows/ci.yml:154; issues #461 and #401

---
Measured on 2026-09-04 by reading the plan and then verifying every commitment in it against the code and the git history. The plan's own status markers were not trusted; neither was `docs/plans/README.md`, which is wrong about several of them.

muhammadbayoumi:
**Closed: it dies with #401, which closed on the same measurement.**

This is the plan-audit wrapper around the same assertion. Both halves are answered:

**The flakiness half.** `8df368d` (#938) rewrote both failure modes out of
`tests/test_the_engine_survives_being_killed.py` — the kill now lands mid-flight by
construction, and the instantaneous read became a bounded 60-second poll. Measured on Windows
today: **6 consecutive green runs** of that file. Against this issue's reported 8-failures-in-13
that is p ≈ 0.3%.

**The CI half is confirmed as filed, and it stays true:** CI is ubuntu-only — all four
`runs-on:` in `.github/workflows/ci.yml` (`:73`, `:108`, `:213`, `:553`) are `ubuntu-latest`,
and the only `windows-latest` in the repository is the **build** job at
`release-engine.yml:83`.

That last fact is not this issue's finding, but it is worth someone's eye: the platform whose
race this issue was about is the one CI never runs. If that should change, it wants its own
issue and a number from him — Windows runners are billed differently.

*Verified before closing: the commit is an ancestor of `origin/main`, and the code claim re-grepped today.*

Issue #461, opened 2026-09-04
OP-19 · The chaos test races the startup sweep it is checking — STILL LIVE, re-measured 2026-08-26


muhammadbayoumi:
> **RE-MEASURED 2026-08-20, and "MOSTLY fixed" is too generous.** Thirteen runs of
> `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` across two
> checkouts: **8 failed, 5 passed.** The `swept=True` marker did not close the race.
>
> It fails on **unmodified `cab69b1`** — 1 of 5 there — so it is not caused by any
> branch in flight; a separate worktree carrying only documentation changes failed
> 7 of 8. That asymmetry is this entry's own thesis rather than a second bug: the
> conditions differed, and **two other sessions were editing this repository at the
> time**, which is precisely the "loaded Windows machine" named below. The failure
> is always the same assertion, `tests/test_the_engine_survives_being_killed.py:266`
> — after crash and restart the job still reads `running`.
>
> The trap described below was walked into again on the way here, and this entry is
> what stopped it: three consecutive failures in the modified worktree looked like
> proof the branch had caused them. Four passes on the unmodified checkout refuted
> that in one command. **Never conclude from runs on one side only.**
>
> **Re-measured again on the evening of 2026-08-20, and this time the load was
> deliberate.** The partitioned muqawil crawl was running — ~1,964 requests against a
> live site — which is the loaded machine this entry names, rather than a coincidence
> of other sessions. Under it: **3 of 3 failures on unmodified `main` at `2366d6d`**,
> same assertion at `tests/test_the_engine_survives_being_killed.py:266`, and the same
> failure inside the branch. Both sides, as this entry demands. So a crawl in progress
> is now a known way to reproduce it fairly reliably — which is worth more to whoever
> fixes it than another run of the flake in quiet conditions.
>
> Consequence unchanged and worth restating: `_source_is_busy` reads that status,

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L7363](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L7363). `OP-19` is kept in the title because the code cites it.

muhammadbayoumi:
**Closed: the race is gone, and the code says so in its own words.**

`8df368d` (#938) replaced the instantaneous read with a bounded wait.
[...]

Issue #401, opened 2026-09-04
OP-98 · A killed engine leaves its job saying `running` — on Windows, and CI cannot see it


muhammadbayoumi:
**Found 2026-08-28 by running the full suite on the owner's machine. NOT this branch's
defect — proven.**

`tests/test_the_engine_survives_being_killed.py::test_a_killed_engine_does_not_leave_a_job_claiming_to_run`
starts a real engine, kills it, restarts over the same database with `swept=True`, and
asserts the sweep has reset the job. It reads `running`.

**Isolated to the platform, not to the change.** Re-run on a clean checkout of `main` at
`b836de3` in a second worktree, with none of this branch's commits present: **identical
failure, same assertion, same value.** `main`'s CI is green, and
[the CI workflow](../.github/workflows/ci.yml) runs on ubuntu only — so this is a
**Windows-only red that CI is structurally unable to report.**

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L3476](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L3476). `OP-98` is kept in the title because the code cites it.

muhammadbayoumi:
## One file, two reds, opposite meanings — and only one of them is this issue

Reproduced this defect on Windows today during PR 640's verification, and in separating it from what two other sessions hit the same day it turned out **`tests/test_the_engine_survives_being_killed.py` produces two different failures that both read as "the chaos test failed again".** Three sessions each characterised it separately and none noticed they were describing different things.

| line | assertion | what a red there means |
|---|---|---|
| **266** | `after a crash and a restart the job still says 'running'` | **This issue.** The crash was staged, the restart happened, the sweep did not clear the job. A genuine Windows defect. |
| **254** | `the job settled before the kill landed, so this run proved nothing about a crash` | **The test's vacuity guard working.** The crawl won the race under load, so the test refused to report a pass it had not earned. Not a defect in the engine. |

Line 254 is deliberate, and the file says so at `SLOW_DETAIL_S = 0.6`:

> Enough products, answered slowly enough, that the crawl is still in flight when the kill lands. **A shop that answers instantly would finish before this test could stage the crash, and the test would pass without ever testing it.**

So a red at 254 is the test protecting the suite from a false pass. A red at 266 is this issue. **Check the line number before filing or attributing either.**

## The reproduction, for the record

[...]

PR #608, opened 2026-09-05
The updater gets a door — the button half

muhammadbayoumi:
## What this is now

**The button half only.** This pull request was split; the reading half is
#680, and it merged. The diff below is the pressing and the watching, and
nothing else.

Base is `main` again, rebased onto the squash of #680 (`66caca5d`).

## Why it was split

The second critical review of the unsplit change returned **four must-fix and
nine should-fix — more than the first pass had.** That is not converging, and
`CLAUDE.md` says a change that does not converge is too big to review.

Every must-fix was in this half: a lifecycle defect, not a rendering one. The
read half could only paint a wrong sentence; this half can leave a control that
lies about work in flight. Holding both at once is why neither got reviewed
properly.

The findings against this half stand and are unaddressed here — they are the
work of the next passes, now against a diff a third the size.

## The change

- **`startEngineUpdate` POSTs and reports what the engine SAID about the
  asking.** A refusal comes back named (`already running`, `no release`,
  `already the published version`, `no installer attached`), so the reason is
  shown rather than a status code — `OP-126` is the row about a guard that
  accepted any 4xx and therefore recorded a reason it never measured.

  **A timeout is not a refusal.** The engine sets `running` and starts its
  worker thread *before* this request returns, so a deadline that expires says
  nothing about whether the download began — and it usually did. Printing "the
  engine refused" over a download that is running is a visible wrong record,
  which is worse than a silent one.

- **`pollEngineUpdate` watches, one watcher at a time, and only while ScrapeX's
  own screen is up.** 310 attempts at 3 s is 930 s, above the engine's own
  `DOWNLOAD_TIMEOUT_S = 900.0`, so the engine is the side that gives up first
  and can say why (`R-48` rule 2) — the same arithmetic as
  `RESTART_CONFIRM_ATTEMPTS`, 125 over the engine's 122. Running out is not a
  failure here: the panel stops watching, the engine keeps downloading.

- **The primary action becomes the engine's when the engine can take it.**
  `chrome.downloads` remains the first install and only the first (`R-36`): the
  panel can never be the installer.

  **`staged` disables the button.** `update_available` and `can_self_update` are
  recomputed from the *running* version, which staging does not change, so both
  stay true after a successful download and a second press would fetch ~70 MB
  again on top of a file already there and already verified. "A button that
  cannot work is worse than no button" — this is the other half: one that works
  and should not.

- **The installer note gains its new wording here**, in the half that earns it.
  It stops exactly where `plan_swap` stops: the engine fetches and checks;
  replacing the running executable is still his step.

## Two corrections carried in from the review

**`READERS` names `pollEngineUpdateLoop`, not the wrapper.** The previous commit
split the poll into a wrapper plus a loop and left the marker on the wrapper, so
the contract guard went blind on half its subject with every test still green —
in the file written about tests that agree with the wrong thing. The floor on
the slice length now scales with the number of readers, so an entry added
without a matching function fails loudly instead of silently shrinking the
window.

**The comment above the poll guard claimed `engine-detail` is not in `VIEWS`.
It is** — `extension/app.js:139`. The claim came from reading eleven lines of a
longer list, and it reached a pushed commit message as well. The guard itself
was never wrong; its stated reason was.

## Gates

`python -m pytest -m "extension or docs"` — exit 0 on `338753ff`.


muhammadbayoumi:
## Critical review: **NOT CLEAN**. Three must-fix, six should-fix.

The four-dimension panel and adversary that `main`'s merge bar requires have run.
**This is not mergeable, and the reason is in the middle of the feature rather than
at its edges.**

### MUST FIX 1 — the caller cannot read the answer it receives

`create_update_router()` returns the phase, the progress and the failure detail
**nested under `progress_state`**, and the refusal reason as
**`self_update_blocked_because`**. This branch reads `report.phase`,
`report.progress` and `report.detail` — three keys that do not exist at the top
level. `grep -rn "progress_state\|self_update_blocked_because" extension/` returns
**zero**.

Measured against the router's real bodies:

| engine state | what it says | what this branch prints |
|---|---|---|
| source checkout | *"This is a source checkout… Update it with git rather than from here."* | generic *"This engine cannot replace itself…"* |
| release with no SHA-256 | *"This release attaches no SHA-256, so its download cannot be proved whole."* | **the identical generic sentence** |
| downloading, 50 % | 50 % | *(empty — line hidden)* |
| staged | downloaded, digest matched | *(empty)* |
| failed | the typed reason | *(empty)* |

**The two opposite causes this feature exists to tell apart collapse into one
sentence**, which is the exact thing `update_api.py`'s own comment forbids. `busy`
is permanently false, the poll exits on its first iteration, and the constants
derived from `DOWNLOAD_TIMEOUT_S` never execute.

**Repair:** normalise once in `engineUpdateState` —
`{...body, ...(body.progress_state || {}), blocked: body.self_update_blocked_because}`
— and read `blocked` for the refusal. Keep it off `detail`: a refusal reason and a
failure reason are different facts.

### MUST FIX 2 — the five new tests are green over a body the engine cannot produce

`_report()` in the tests invents a **flattened** shape, and `_STUB` replaces
`window.fetch` wholesale, so no test has ever seen the real router. **The GET
body has never been asserted by anything, on either side of the seam.**

So the mutation check — five of five fail with the caller removed — was true and
worthless: the code and the tests share the same wrong assumption, so neither can
expose the other.

**Repair:** rebuild `_report()` from the router's real key layout, and add one
Python test that calls `create_update_router()` through a `TestClient` and asserts
its top-level key set against the keys `app.js` reads. **Without it this class of
defect re-enters silently.**

### MUST FIX 3 — the new line renders one engine's sentence on another engine's screen

`#engine-update-state` is a sibling of `#engine-detail-actions` and is **not in the
hide list** `renderEngineDetail` applies for a candidate engine. Open ScrapeX
Engine, go back, open Scrapy: the paragraph is still there — on the one screen
whose own comment says it "promises nothing to press". No race needed.

**Repair:** add it beside the other hides, and gate the un-hide with
`theInstalledEngineIsOnScreen()` exactly as the sibling node already does.

### SHOULD FIX 4–9

4. `/api/update` has no `LOCAL_POLICIES` row, so it takes the generic 5 s deadline
   against a handler whose whole cost is a 4 s third-party fetch — and the timeout
   prints *"The engine stopped answering"* about an engine that is downloading.
5. A POST timeout renders as *"The engine refused"* about a download that started.
6. The poll neither stops when the screen is left nor restarts when it is re-entered.
7. At `staged`, the only prominent control re-downloads the ~70 MB file that is
   already staged and verified.
8. The panel's own row and the engine's sentence can contradict each other, and
   every new test runs in that state.
9. Three false claims in the prose shipped with this diff.

### The deciding question, answered

*Does this make the owner's screen state anything false in a reachable state?*

**Yes — five reachable states are false and five more are silently omissive**,
including both states the feature was written to distinguish. The worst is the
common one: he presses the new control, 70 MB downloads, **and the screen returns
to exactly what it was**. The only way to learn a download is running is to press
again and read *"An update is already running."*

The engine's capability is real, and the caller genuinely reaches it — `POST` is
correct, which is why hand-clicking looks like it works. What is wrong is
everything the panel does with the reply.

---

**Status: held, not abandoned.** The branch is `feat/the-updater-gets-a-door` at
`f1dee213`, based on `c6269a2a` and now one behind `main`. It needs a rebase onto
`68030ab4` — which touched `extension/app.js` and `extension/app.html`, so the
overlap is real — then the repairs above, then the panel again.

muhammadbayoumi:
**Returning this to its session: two required checks are failing, so this is not mergeable and the merge session is not going to diagnose it.**

Read at `f7359709`, after the rebase:

```
mergeable     MERGEABLE
mergeState    BLOCKED
rows          10 pass · 4 fail
```

The four failures are two checks, each appearing twice because a PR carries both a `push` and a `pull_request` run:

| check | runs |
|---|---|
| `test` | [33974483134](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/33974483134/job/101328689912) · [33974485910](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/33974485910/job/101328698423) |
| `migration-authority` | [33974483134](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/33974483134/job/101328689985) · [33974485910](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/33974485910/job/101328698413) |

Both are in the required set under the branch protection now on `main`, so GitHub will refuse the merge regardless of anything else.

**The rebase itself succeeded** — the branch does now contain `68030ab4`, and `mergeable` is `MERGEABLE`, so there is no textual conflict. The failure is behavioural, which is exactly what the rebase was supposed to surface and did.

**One thing worth knowing before you diagnose:** `main` has moved again since this run. It is now `7a64810d` ("The crawl's own report names the action, not a command (#610) (#641)"), which changed `scrapex/contractors.py` and one test file. So this branch is behind once more and will need another rebase after the fix. Do the fix against current `main` rather than against `68030ab4`, so you are not diagnosing on a base that has already been replaced.

**A caution about `migration-authority` specifically:** that job sets `SCRAPEX_FULL_MIGRATIONS=1`, and there is a known intermittent that fires under it — `test_a_killed_engine_does_not_leave_a_job_claiming_to_run`, filed as `#401` and `#531`, with `#461` on the race in the same file. On Linux it normally passes, so a failure there is more likely real than intermittent — **but read the failure body before assuming either way.** A mass of failures in one family is an instrument problem until proven otherwise; a specific named assertion is a defect.

**Also still owed on this branch, independent of the red:** its critical review has not been run. It is a code change, so `#595`'s bar applies — the four dimensions, an adversarial pass over what they find, then a re-review after the fixes. The merge session will ask what KIND of review ran before it asks for completion figures, because a review with no agents reports "0 errored" truthfully.

Post the settled read when both are green: rows with zero pending, and confirmation that the `test` row was the engine tier rather than a documentation-tier pass.

muhammadbayoumi:
**Your red is fixed and this was ready — `#640` merged ahead of it, so it needs one rebase. I have simulated it: there is no conflict.**

```
origin/main   7a64810d -> fcca90b4
              Two rulings share one number in shipped code, and the design system cites the wrong one (#640)
mergedAt      2026-09-06T07:03:26Z
```

This PR now reads `BEHIND` by 1. Its 14 rows are all `pass` and its `test` job ran the real tier — I confirmed the step `Tests (Python engine + Python-vs-frozen contract)` reported **success**, with the documents-only and extension-only tiers correctly skipped. So the work is verified; only the base moved.

### What I checked before merging `#640` ahead of you

You share a file with it — `tests/test_panel_dom.py` — so this was not a disjoint case and I did not treat it as one.

| | change to `tests/test_panel_dom.py` |
|---|---|
| `#640` | 7 lines changed, scattered across 388–694 (ruling-number renames) |
| this PR | 223 lines, pure additions appended at 6183 |

The hunks do not overlap, and `git merge-tree` on the two heads merged cleanly. Simulated again against the merged `main`: **no conflict**, so the rebase is mechanical.

I also checked for content coupling rather than stopping at the text, because the documented failure here is two PRs that share no lines and break each other anyway:

- **`#640` adds a new guard**, `tests/test_the_design_system_cites_its_own_rulings.py`, which pins every ruling number cited on the design surface to the heading it still carries in the frozen archive. Your diff adds `R-36`, `R-48` and `R-80` citations in `extension/app.js`. **That file is not in the guard's `DESIGN_SURFACE`** — the tuple is exactly the six design assets (`tokens.css` and `appearance.js` in `design/`, `extension/` and `scrapex/webui/static/`). So your citations are out of its scope and it will not fire on them.
- Your diff also adds **`#246`** in a comment. That is the shape the colour-literal guard reads as a three-digit hex value, and it has caught several sessions today. Yours is in `tests/test_panel_dom.py`, which that guard does not scan, so it is safe — but if any of that text ever moves into a UI source file, write it as `PR 246`.

### What to do

Rebase onto `fcca90b4` and re-run. Then the check that proves the rebase moved the base and nothing else:

```bash
```

**Still owed and unchanged: the critical review.** This is a code change, so the bar applies — the four dimensions, an adversarial pass over what they find, then a re-review after any fixes. When you report it, say what KIND of review ran before its completion figures: a review with no agents reports "0 errored" truthfully, and the merge session now asks in that order.

Post the settled read and it will be merged.

muhammadbayoumi:

```
```

## What actually happened

```
```

## What to do

```bash
```

---

muhammadbayoumi:
**`main` moved again and this one overlaps your branch in files, so your rebase is no longer only about the base.**

```
origin/main   74caea5c -> 25e63535
              A Database page of its own, above Engine, and the upgrade moved to it (#679)
mergedAt      2026-09-06T12:12:32Z
```

`#679` changed seven files, and these are the ones it shares with your branch:

```
extension/app.html
extension/app.js
tests/test_panel_dom.py     (this one only if your branch touches it)
```

It moved the upgrade control out of Engine and onto a Database page of its own, so the panel's structure around those controls has changed — not just the lines' positions.

**What that means for you:** the earlier rebases today were mechanical because the changes were disjoint. This one may not be. Re-derive rather than assume, and if you cite a line number in `extension/app.js` anywhere, recompute it against `25e63535` rather than carrying it.

**The check that proves your rebase moved the base and nothing else:**

```bash
```

Identical before and after is a stronger statement than "no conflict". If it differs, the rebase changed your content and you want to know why before you report a green.

## Two things to confirm before calling it ready, both learned today

**One — measure the base, do not read it.** GitHub reported `mergeable_state: clean` on a PR that was two commits behind with a required check missing. `mergeStateStatus` describes whether git can merge the branch, never whether the branch is correct, and it is computed asynchronously so it lags a merge by minutes. Measure it:

```bash
```

**Two — confirm the row set is complete, not just green.** A PR can show all-pass while a required context never reported at all. Expect nine names:

```bash
gh pr checks <n> | awk -F'\t' '{print $1}' | sort -u
```

And confirm the `test` row was the engine tier rather than a documentation-tier pass wearing the same name — check the run's step conclusions, preferring the `pull_request` run.

muhammadbayoumi:
**Split.** The second critical review returned four must-fix and nine should-fix — more than the first — so this was cut in two rather than pushed at a third time.

- #680 — the read half: the panel asks the engine and renders the sentence. Merges on its own, against `main`.
- This PR — the button half, stacked on #680. It retargets to `main` when that merges.

The head was force-pushed (`37dc4151` → `338753ff`); the pre-split tree is kept at the tag `split-safety-37dc4151` locally and every commit that was here is contained in the two branches.

The thirteen findings against the unsplit change are not addressed by the split itself. Every **must-fix** lives in this half — they are lifecycle defects, not rendering ones — and they are the work of the next passes, now against a diff a third the size.

Splitting also surfaced a finding the unsplit change could not: `test_the_checksum_on_screen_says_who_checks_it` keyed the installer note's claim on `"/api/update" in app.js`, so a **read-only** caller made it *demand* the note promise engine-side verification — the exact false promise `OP-124` removed, re-required by the test written to prevent it. Fixed in #680, where the condition first becomes wrong.

muhammadbayoumi:
**The three must-fix are fixed** (`2d158ff1`). The fourth, M1, is [#681](https://github.com/muhammadbayoumi/ScrapeX/issues/681) — an engine defect that predates this change.

**M2 and M3 were one missing call.** `renderEngineDetail` painted the rows and the banner and asked the engine about nothing, so returning to ScrapeX's screen left the sentence blank — the candidate branch had hidden it and nothing rewrote it — and left the button frozen at whatever it said on the way out. Open while idle, leave, come back mid-download, and the button still *offered* a ~70 MB download that was already running. "Reopen this screen to see where it got to" was a false instruction printed by this feature's own poll.

The installed branch now calls `renderEngines()`. Not a smaller call: the sentence, the latest version, the verdict badge and the button all come from one report, and asking for a subset is how they end up disagreeing. Not awaited, so `renderEngineDetail` stays synchronous and `showView` and the title focus are not put behind a round trip.

**M4 was a guard asked only on the way into an eight-second wait**, and the repair is that the guard moved into the one writer. Four callers put text on that line — the renderer, the button, each turn of the poll, and the poll running out — and when the check was repeated at the call sites it was right at three of them and absent at the fourth. `writeEngineUpdateLine` asks both questions and returns whether it wrote, so the poll drops the rest of its work for a screen that has gone rather than fetching a release feed for it.

**The harness had to change to make these falsifiable.** The stub reads `window.__update` instead of a closed-over value — a screen that is never re-rendered and one that is look identical unless the truth moves while the reader is away — and `__delay_ms` holds a report in flight, because one that arrives while he is still on the screen paints correctly under either guard.

Each fix was removed on purpose and its test watched to fail. `pytest -m "extension or docs"` exit 0.

Still open: the nine **should fix**, and the deferred one — the panel's release row and the engine's sentence can contradict each other, which needs a decision on which source wins.

muhammadbayoumi:
**Seven should-fix defects fixed** (`8e6efc7f`). One honest note first: the second review's list of nine was never written into this PR — it existed only in that session. So these are re-derived by measurement against the current, smaller diff rather than cited by number, and two of the original nine (the blind contract guard, the fixture shape) were the split's own repairs and are already in.

| | measured | now |
|---|---|---|
| **A deadline measured for one method was binding the other** | `POST /api/update` → **8000**, was `localMutation` **10000** | a policy row may name a method; this one names `GET`. POST → 10000 |
| **The renderer awaited ahead of every paint it already had data for** | `await engineUpdateState()` was the first statement — the latest version, the release detail, the button label and `#engine-row-verdict`, *which is on the catalogue screen*, all waited up to 8 s for a report none of them read | the await moved below them |
| **The DOM fixture's `latest` block carried 3 keys of 9** | `detail` among the six missing — read by `engineUpdateSentence` on the unreachable-feed path | the contract guard now looks inside `latest`, and `test_panel_dom.py` asserts its own fixture against the live router |
| **`staged` named a step and withheld what taking it needs** | `staged_path` and `staged_version` in the report, read by **nothing** | the sentence names the file; the test demands it |
| **The poll running out left a dead button** | busy disabled it; the exhausted branch never re-enabled it | re-enabled beside the sentence |
| **"`git log -S \"api/update\"` returned no commit in the whole history"** | **false** — it returns `c233a219`, the commit that wrote the comment saying the route had zero callers. `git log -S` searches text, so a note *about* a route counts as an occurrence of it | replaced with the claim that is true and repeatable in one line: `grep -rn "api/update" extension/` found exactly one hit, inside an HTML comment |
| **"310 attempts is 930 s"** | a floor, not the value: each turn is the 3 s sleep **plus** a request bounded at 8 s, so 930 s–3410 s | stated as a floor; the conclusion strengthens, since every one of those is above the engine's 900 |

The deadline rule and the staged sentence were each broken on purpose and their tests watched to fail. `pytest -m "extension or docs"` exit 0 · `node --test extension/tests/*.test.mjs` 428 pass · contract parity pass.

**Not fixed, deliberately, and each recorded rather than argued in a file:**

- [#686](https://github.com/muhammadbayoumi/ScrapeX/issues/686) — `GET /api/update/plan` is the second doorless route in this same feature. Drawing the plan is a new control, not a repair, and it is blocked on [#614](https://github.com/muhammadbayoumi/ScrapeX/issues/614).
- [#687](https://github.com/muhammadbayoumi/ScrapeX/issues/687) — `renderEngines` now has two un-awaited entry points and its release-feed guard is check-then-act across an await. **Optional**: costs one duplicate request pair on a narrow race, renders nothing wrong.
- [#681](https://github.com/muhammadbayoumi/ScrapeX/issues/681) — the engine's own collapsed refusal sentence (was must-fix 1), predates both halves.

**And one needs a decision, not a fix:** the panel's release row and the engine's sentence can contradict each other, because two sources answer the same question — the panel's own manifest fetch (which must stay, it is the only answer when there is no engine) and the engine's report. Which one wins when they disagree is his call, so it is not being decided in a diff.

muhammadbayoumi:
**The engine wins** (`6147b2aa`). His ruling, on the one finding the review deferred because a diff could not settle it.

When the engine has answered, its answer replaces the release row — the version, the detail, the verdict badge **on both screens**, the button label, and the installer behind it — and `installed` with them, which the engine knows first-hand and the panel only remembers.

**Two boundaries, each with its own test, each broken on purpose:**

- **The panel's own fetch stays.** Deleting it is the obvious wrong reading: `renderEngines` runs with no engine at all — the first-install case — and `update_api` reports the installer URL *"so the panel can still hand the file to chrome.downloads for a FIRST install"*. The engine wins when it speaks; it cannot win when it is not there.
- **It wins with an answer, not with a failure to get one.** A report whose own manifest fetch failed carries `latest.state` of `offline` or `unreadable` and no version. Taking that would blank a row the panel had filled correctly — the ruling costing the row the trust it was meant to buy. The state is checked, not merely the presence.

Making the panel win again fails the first test; letting an unreadable feed win fails the second.

`paintEngineRelease` is extracted because the row is painted twice now, from two sources — painting it twice from two copies of the same code is how those two would come to disagree in a third way. The **first** paint still happens before the await, so `#engine-row-verdict`, which lives on the catalogue screen, does not sit blank for eight seconds.

**One thing measured while writing the test, recorded because it looks like a defect and is not:** with `can_self_update` true the button branch returns before the tail, so `#engine-download-checksum` keeps the panel's digest while the row shows the engine's version. It is inside `#engine-install-steps`, which that same branch hides, and nothing can un-hide it without re-running the tail that rewrites it. The test asserts the state where the number is readable, and opens the `<details>` first so it asserts about the screen rather than about the DOM.

**And the ruling found one more doorless thing** — [#690](https://github.com/muhammadbayoumi/ScrapeX/issues/690): both release paths publish `minimum_extension`, parsed and tested, read by **nothing**, under a comment saying it exists so the panel can refuse an incompatible engine *before* installing rather than after. Not fixed here — refusing needs a decision about whether the button is disabled or the reader is warned.

`pytest -m "extension or docs"` exit 0 · `node --test extension/tests/*.test.mjs` 428 pass.

muhammadbayoumi:
# HANDOVER — stop point, and what the next machine picks up

**Head `d753cafa`, based on `origin/main` `2a046ae0`. Tree clean, everything pushed, gates green.** Nothing exists only on the machine this was written on.

`pytest -m "extension or docs"` exit 0 · `node --test extension/tests/*.test.mjs` 451 pass.

## What the four-dimension panel returned

Run on `499503b6`. **Fourteen findings: three must-fix reached independently by three reviewers, plus the tests reviewer's own four.**

| dimension | verdict |
|---|---|
| architecture | 1 must-fix · 3 should-fix · 1 optional |
| code quality | 1 must-fix · 3 should-fix · 1 optional |
| performance | 1 should-fix · 2 optional |
| **tests** | **4 must-fix · 5 should-fix · 1 optional** — 23 mutations run, **14 survived** |

## DONE in `d753cafa` — the code half

- **The must-fix all three reviewers found.** One lost poll turn rendered "the engine stopped answering" and ended the watch, about an engine still downloading. The deadline is a **tie, not a margin**: `CHECK_TIMEOUT_S = 4.0` is applied by httpx **per phase**, so the engine's worst legitimate answer is connect 4 s + read 4 s = exactly the 8000 ms the panel allows. The attempt budget owns the giving up now, and **the button is not touched in either terminal branch** — re-enabling it produced a control that was enabled, labelled "Downloading…", with `onclick = null`: pressable and inert.
- `installerDownloadRunning`, cleared on **both** exits.
- `staged` now means *this* release is staged (`staged_version === latest.version`).
- The engine's `verifiable` verdict is honoured — `=== false`, never `!verifiable`.
- `READERS` gains `updateEngineReleaseUI`; the slice floor is **per slice**, not on the sum.
- The checksum note is qualified; the poll's derivation corrected a second time.

## NOT DONE — the whole tests dimension. Start here.

**Four must-fix, measured by mutation, all still open:**

1. **`pollEngineUpdateLoop` has no behavioural coverage.** Three mutations — gutting the loop, removing the reopen restart, removing the watcher after a successful POST — left **all 17 tests green**. The only test reaching the loop asserts an *absence*, so deleting the watcher satisfies it for the opposite reason to the one it names. Needs one progression test: stub 20%, flip `window.__update` to 50%, watch the line follow, flip to `staged`, assert the loop stopped.
2. **Every error path in the write half is deletable with everything green.** The whole POST `catch`, the poll's lost-turn branch, and the "Asking the engine…" line each survived. `_STUB` answers every POST with 200 and **has no failure mode at all**; `isTimeoutError` is never exercised. The timeout-versus-refusal distinction the longest comment in this diff exists to protect is unverified.
3. **The `staged` button branch has no assertion.** Making the staged button live, and relabelling it, both survived.
4. **The contract guard.** `READERS` and the per-slice floor are fixed in `d753cafa`, but the reviewer's two mutations (`report.installed` → `installed_version`; split the loop into a wrapper and rename `report.phase` inside) should be **re-run to confirm they now fail**. They were not re-run after the fix.

**Five should-fix, all open:**

5. `report.installed` cannot be asserted by construction — `_HEALTH["version"]` and `_report()["installed"]` are both `"0.4.6"`, so the half of the ruling about `installed` is invisible to every test. Change the fixture to `"0.4.7"`.
6. "Leaving the screen stops the watching" — the **view half** of the guard and the pre-await guard both survive mutation; only the ownership half is covered.
7. **`phase: "failed"` has no test at all** — the state the owner sees on a digest mismatch, the branch carrying the whole security value of verify-before-stage.
8. `test_the_panel_reads_the_refusal_reason_at_all` greps the whole file, and the literal also appears in **two comments** — renaming the read keeps it green. Assert on `_normaliser_pairs()` instead. *(Partly addressed in `d753cafa`; re-run the mutation to confirm.)*
9. Three DOM tests hand-write `latest={...}` inline and bypass the new fixture guard — stripping six of nine keys stays green. Give `_report()` a `latest_over=` kwarg.

## Filed as issues, not to be fixed here

- **#761** — `GET /api/update` refetches the manifest every poll turn: 39–311 third-party round trips per download, for one integer. Engine-side.
- **#762** — Restart discards a download in flight and orphans the `.part`; `clear_staging` has no production caller.
- **#763** — a dead branch, an unreachable ternary, and a percentage that can freeze at 0% for a whole download.
- **#681** — extended with the measurement: `(update_available=True, can_self_update=False, blocked="")` is reachable in **no case**, so the panel's no-installer fallback is dead against the real engine. The panel half is fixed here; the engine half is that issue.
- **#686**, **#687**, **#690** — from the earlier passes, unchanged.

## The bar this has not yet cleared

`CLAUDE.md`: a change merges only when a critical review returns **nothing**. This review returned fourteen findings and roughly half are addressed. **After the tests work: re-run the four dimensions, then the adversary, then merge only on a clean pass.**

Also note `main` moves several times a day — rebase before the next gate run rather than after.

PR #640, opened 2026-09-05
Two rulings share one number in shipped code, and the design system cites the wrong one

muhammadbayoumi:
Closes #617.

`R-84` in `docs/archive/RULINGS.md:3659` is *"The base changes now — and at publication no migration is ever deleted again"*. `scrapex/dbupgrade.py`, `storage.py`, `version.py`, `databases/domain.py`, `contractstamp.py` and `db/engine/schema.sql` cite it for exactly that, correctly.

**Fifty-three other citations use the same number for the design system**, which is `R-85` at `docs/archive/RULINGS.md:3757`. The lines read `R-84: ONE COLOUR CHOICE`, `R-84 deleted device colours, so there is one colour mode and no attribute to carry a second`, `THERE IS NO DEVICE COLOUR PATH, AND THAT IS R-84 RATHER THAN AN OMISSION`, and «احذف الثلاثة وابق supabase وحده» — every one of them `R-85`'s subject, none of them `R-84`'s.

## Why this is a defect and not stale prose

`CLAUDE.md` makes `git log --grep=R-84`, then `docs/archive/`, the supported way to learn why a line exists, and names the code's citations as the **only** reason that frozen archive is kept at all. So a wrong number is not old prose in a dead file — it is a broken lookup in the mechanism that replaced the registers. And it fails the worst way a lookup can: the reader lands on a real ruling, in another domain, with nothing to suggest they are in the wrong place.

## Measured, not swept

Every `R-84` outside `docs/archive/` was read and classified by what the surrounding sentences are about.

| | count | disposition |
|---|---|---|
| migration, correct | 41 | untouched |
| design system, wrong | 53 | renamed |
| `CLAUDE.md`'s `git log --grep=R-84` | 1 | left — it is the syntax, not a claim about the ruling |

**A blanket rename would have corrupted 41 correct citations to repair 53 wrong ones.** Twelve could not be classified by subject-matter matching and were read individually: four design (`test_panel_dom.py`'s *"supabase is the baseline colour choice under R-84"*, *"R-84 PUT TWO OF THESE BELOW THE FLOOR THEY ASK FOR"*, and `test_a_palette`'s *"UNTIL R-84, WHICH DELETED TWO OF THE THREE"*), six migration, one neither.

The 53 break down as 30 across the three synced copies of `tokens.css` and `appearance.js`, 2 in `scrapex/webui/app.py`, and 21 across six test files — including `tests/test_vendor.py`, the file that enforces the design system's own colour rules.

[...]

PR #669, opened 2026-09-06
A finished crawl's pages become rows, from the panel (#668)

### LINKED ISSUE #668: A finished crawl's 6,713 pages cannot be turned into rows from the panel — the interpret stage has no door
**His crawl finished at 2026-09-06T05:01:44Z with 56 of 56 cells and zero errors. The panel shows the same 17,304 rows it showed before it started, and says the source was last collected on 21 August.**

Nothing is broken. The pages are stored and safe. **They simply cannot be turned into rows from the only interface he uses.**

## Measured

```
pages this crawl stored       6,713
of those, interpreted             0
newest generic_ingestion      2026-09-01T12:06:39Z     (five days before the crawl ended)
newest page captured          2026-09-06T05:01:24Z
dataset rows, before & after     17,304
```

## Why the card says 21 August

`_dataset_freshness` reads `extract_service.last_evidence_captured_at`, and that function's own docstring says what it reads and why:

> *"`generic_ingestion` … is which pages this dataset was BUILT FROM — both already written on every crawl — so the freshness is a read, not a column."*

It reports when the dataset was last **fed** a page, not when a page was last **fetched**. Both are honest questions; the card asks the first. With no ingestion since 1 September, `2026-08-21T17:56:31Z` is the correct answer to the question being asked — and it is the wrong question to put on a card beside a crawl that finished five minutes ago.

## Where the capability lives, counted

The step that turns stored evidence into rows is `extract/service.py:527 approve_candidate`. Where can it be started?

| | |
|---|---|
| command line — `scrapex contractors --approve` | **14 occurrences**, help text: *"interpret the stored pages of --run-ref into rows. Re-fetches nothing"* |
| API route matching approve / interpret / ingest | **0** |
| job kind in `jobs.SPECIALISED_RUNNERS` | **0** — only `organization_enrichment` and `directory_crawl` |
| control in the panel | **0** |

**This is `R-81` at its largest scale so far.** A fourteen-hour crawl of 6,713 pages produced evidence that only a terminal can convert, and he does not use a terminal. Every earlier instance of this rule was a sentence offering a command; this is an entire stage of the pipeline with no door.

## It is also why the previous findings compound

`#606` (a resume that reads as a stall) and `#613` (no Resume control, no discard warning) are about the panel misdescribing a crawl. This is about the crawl's **output** never arriving. Together they mean the surface he watches can be wrong about the past, the present and the result.

## Options

| | effort | risk | what it costs |
|---|---|---|---|
| **do nothing** | none | every crawl's harvest waits for someone at a keyboard | the data he crawls is not the data he sees |
| **a route only** | small | a route with no control is `R-81` again, one layer in | nothing reaches the panel |
| **a job kind + a control** (recommended) | medium | it is a third specialised runner, and the pattern is established twice already | the crawl's result arrives where he can see it |
| run it automatically at the end of a directory crawl | small | **rejected as the whole answer** — interpretation can fail on its own terms, and folding it into the crawl means one failure reports as the other. It is a reasonable *default* once the job kind exists, never a substitute for it |

**Recommendation: a job kind, reached the way the other two are, plus a control on the dataset card.** `organization_enrichment` and `directory_crawl` both did exactly this, so the shape is proven and the registry entry is one line.

**And the card should say which question it is answering.** *"Last collected"* over an ingestion date, beside a crawl that ended today, is a true sentence that misleads. Either say *"last interpreted"*, or show both.

## One thing to decide before building

`approve_candidate` short-circuits on `(snapshot, locator)` plus `schema_hash`, which was recorded as a defect years-of-sessions ago: a corrected parser returns `recovered=True` for every page and writes nothing. **Whether re-interpretation of already-ingested pages is in scope for this control is his call** — for the 6,713 pages here it does not arise, because none of them has ever been ingested.

muhammadbayoumi:
## The stage that had no door

His muqawil listing crawl finished **2026-09-06T05:01:44Z with 56 of 56 cells, zero errors and 6,713 stored pages**. Zero of them were interpreted. The newest `generic_ingestion` row was five days *older* than the crawl, and the card went on showing the 17,304 rows it showed before the crawl started.

`contractors.approve` is what turns evidence into rows. It could be reached from `scrapex contractors --approve` — fourteen occurrences — and from nowhere else:

| | |
|---|---|
| command line | **14** |
| API route | **0** |
| job kind | **0** |
| control in the panel | **0** |

`R-81` says the panel is his only interface, so a fourteen-hour crawl produced a harvest only a terminal could convert. Every earlier instance of that rule was a *sentence* offering a command; this was an entire stage of the pipeline with no door.

Closes #668.

## Measured on a copy of his real warehouse before any of it shipped

2.08 GB taken through the SQLite backup API rather than a file copy — consistent against a live writer, and it does not need the `-wal` that a plain copy can miss committed pages without. Migrated, then interpreted:

```
909 page pairs in 21 minutes
908 ingested, 1 refused with its reason named
   ("No contractor cards on this page")
contractors  17,304 -> 17,738
coverage       96.8% -> 99.4%
```

**434 contractors that were already on disk.** His own file was never opened for writing.

## Three decisions, each with its reason

**A job kind, not a stage of the crawl.** Folding interpretation into the crawl was the cheaper answer and the wrong one: interpretation fails on its own terms — a parser that cannot read a page, a schema that moved — and a failure reported as the crawl's sends the next session looking at the network. It is also worth doing *without* a crawl, because the evidence of a run that finished days ago is still on disk.

**A page boundary is safe where a cell boundary was not.** The crawl stops between cells because a cell's completeness proof spans many pages, so a cell cut in half has fetched pages and proved nothing. Interpretation has no such span: `between_pages` may stop in front of any pair, because every earlier one is written and the next has cost nothing.

**It holds no politeness reservation**, and that is the one place it differs from `directoryjob` on purpose. `_CrawlAdmission` stops two jobs asking one site for pages at once; this job asks nobody for anything, and holding a host lane would block a real crawl of that site for the length of an interpretation that could not have collided with it.

## `0018` widens one CHECK, and the baseline stays at 17

`db.py` refuses a migration numbered at or below the baseline's declared version. The first draft bumped `schema.sql` to 18 as well and broke the whole suite **at collection** — the baseline is a snapshot and migrations run above it.

Measured on the live shape before writing: `crawl_job` is **157 rows**, four tables carry `REFERENCES crawl_job` (`change_event`, `crawl_run`, `job_log_entry`, `organization_enrichment_job`), and `ix_crawl_job_status` has to be put back.

## Two guards that were lying, and one piece of my own dead code

**The migration harness lied first.** Run as a plain `executescript`, `0018` leaves all four referencing tables pointing at `crawl_job_old` — which it then drops. Run the way `domain.py` runs one (isolation owned, `foreign_keys` off around it, `BEGIN IMMEDIATE` in front) they survive. **His live warehouse settled which reading was right:** it carries `0017`'s rebuild with every reference intact. The test now applies it the production way, and asserts the log rows and the index survive.

**The card-action guard could not see two of nine offers.** `declared_actions()` read the `SOURCE_ACTIONS` array and split on the literal `"{action:"` — and the two inline offers, the crawl and this one, open the brace and break the line. So that whole suite passed while never looking at either. It reads both regions now and de-duplicates on the action **and its proof**, because `update` legitimately appears twice with two.

**And a mutation found dead code of mine:** a `HAVING pages > 0` that changed no verdict, because the INNER JOIN beside it already excludes a crawl that stored nothing. Deleted, with the real mechanism named in its place.

## Fifteen mutations, fifteen caught

Including three on the panel itself, because **a text guard cannot see whether the panel *offers* an action** — hiding it behind `false ?` leaves the literal in place and the parsing suite stays green. So the offer is proved by pressing it in a real browser and reading what was posted: that the row is there, that the key is the **site's** and not the dataset's, and that the request **names the kind** — without which the route infers a crawl and goes back to the site for pages already on disk.

## Gates

On `76b88c20`, fingerprint `DIRTY=0`, `TREE=389e2338`:

```
ruff scrapex/          clean
pytest -m extension    exit 0
mutations              15 written, 15 caught
```

**Not yet settled, and I will post them here rather than claim them:** the two full-suite runs (template mode and `SCRAPEX_FULL_MIGRATIONS=1`) are in flight on that tree. **This branch is also one commit behind `main`** and needs a rebase before it is eligible under the strict protection — I have not rebased because the suite is reading this working tree, and a green measured on a tree that no longer exists is not a green.


muhammadbayoumi:
**Returning this to its session: four required checks are failing, and the branch is now two commits behind. The merge session does not diagnose reds.**

```
mergeable     UNKNOWN
mergeState    BEHIND        (behind by 2)
rows          10 pass · 4 fail
```

The four failures are two checks — `test` and `migration-authority` — each appearing twice, because a PR carries both a `push` run and a `pull_request` run. That duplication is normal here and is not itself a symptom.

### `main` moved twice since this branch was cut

```
7a64810d   The crawl's own report names the action, not a command (#610) (#641)
fcca90b4   Two rulings share one number in shipped code, and the design system cites the wrong one (#640)
```

**Diagnose against `fcca90b4`, not against the base you were cut from.** Fixing on a superseded base costs the work twice, and branch protection on `main` now has `strict` enabled, so GitHub will refuse the merge until this branch contains `fcca90b4` regardless of the checks.

### Two things worth knowing before you read the failures

**`#640` changed what some guards check.** It renamed 53 wrong `R-84` citations to `R-85` across the six design assets, and added `tests/test_the_design_system_cites_its_own_rulings.py`, which pins each ruling number cited on the design surface to the heading it still carries in the frozen archive. Its `DESIGN_SURFACE` is exactly `tokens.css` and `appearance.js` in `design/`, `extension/` and `scrapex/webui/static/`. If this branch touches any of those six files and cites an `R-` number, that guard is new since your last green.

**`migration-authority` sets `SCRAPEX_FULL_MIGRATIONS=1`,** and there is a known intermittent that fires under it — `test_a_killed_engine_does_not_leave_a_job_claiming_to_run`, filed as `#401` and `#531`, with `#461` on the race in the same file. On Linux it normally passes, so a failure there is more likely real than intermittent. **Read the failure body before concluding either way**: a mass of failures in one family is an instrument problem until proven otherwise, while a specific named assertion is a defect.

### What is needed

1. Read the actual failure bodies, not the counts.
2. Fix against `fcca90b4`.
3. Rebase, re-run, and confirm the `test` row was the engine tier rather than a documentation-tier pass wearing the same name — a row named `test` can be either.
4. The critical review, if this is a code change and it has not been run.

Post the settled read when the rows are green with zero pending, and it will be merged.

muhammadbayoumi:
## Gates settled — everything green, and the version is his

On `9f3f0255` (the tree this branch carried before the rebase), fingerprint identical at both ends, `DIRTY=0`, `TREE=110c5331`:

```
ruff scrapex/                        clean
pytest -m extension                  exit 0
pytest -q                            exit 0
SCRAPEX_FULL_MIGRATIONS=1 pytest -q  exit 0
mutations                            17 written, 17 caught
```

Now rebased onto `fcca90b4` (head `b57fdb93`) and the same gate is re-running on that tree — a green measured on a tree that no longer exists is not a green, and #641 merged into `scrapex/contractors.py` in the meantime, which this branch also touches.

**VERSION 0.4.9 is his word**, given after the contract stamp reported the schema stream moving. `export-version` regenerated `contract-baseline.json` (2 migrations, 96 routes, protocol 1), `capability-baseline.json`, `version-vectors.json` and `CHANGELOG.md`; `pyproject.toml` followed, because `test_the_installer_carries_the_same_number` says bump both or neither. **The protocol did not move** — a new job kind travels through `POST /api/jobs`, which the panel already speaks.

## The finding this change did not contain, and would never have been found without it

`0018` is the first migration to land above `R-84`'s squash, and it detonated a class of assumptions. **One of them is severe and it is not a test.**

`_went_through_the_collapsed_chain` recognises a ledger stamped by the pre-squash baseline and accepts it — that is how a warehouse built before the squash still opens. One of its four conditions demanded the database be sitting **exactly** at the baseline's version. But `_migrate` applies its migrations **first** and stamps afterwards, so the moment anything exists above the squash the database is one version further on by the time the recognition runs. The condition fails, and the upgrade dies with `schema.sql checksum changed` about a file nobody touched.

**Measured on a copy of his own 2.08 GB warehouse:** `initialize()` applied 0018 and then raised.

So **every pre-squash warehouse was permanently un-upgradeable through the panel's own Upgrade button**, and would have been by whatever migration came first. It was not a defect in this change; it was a defect waiting for this change.

`< migration.number` rather than `!=`. Below the baseline is still refused, and that half carries the proof: a database under it cannot have applied the chain the baseline absorbed, so it may not borrow the record's word for it.

**After the fix, on a fresh copy of his warehouse:**

```
before   v17   "Needs upgrade"
initialize() -> applied [18] in 2.9s
after    v18   Healthy
integrity ok · 158 jobs kept · accepts dataset_interpret
```

The new guard takes no tool: `test_a_warehouse_that_went_through_the_chain_can_still_be_upgraded` builds a database through the recovered chain and upgrades it with the shipped build the way the button does. Restoring `!=` reddens it, and reddens the carry test too.

## The other five were the same conflation, corrected rather than relaxed

`baseline` and `latest` were the same number while the stream was one file:

- the **squash record** describes the BASELINE, which is its name — it was compared against `latest_schema_version`, which moves when a migration lands on top;
- the **drift guard**'s short-stream allowance was written for `count == 1` in the week the squash left exactly one file. A stop point needs `2 <= P < count`, so a stream of two has no room either — `count < 3`, with the message saying why the range is empty, and the frozen record still checked rather than assumed;
- **"below the baseline" meant `latest - 1`** in three tests. With a migration above the baseline that *is* the baseline, and the honest verdict there is "Needs upgrade" — precisely the branch `domain.py`'s own comment said three tests believed they were exercising and were not;
- the **carry tool** reaches what this build can reach, not the baseline. A warehouse left at the baseline while the build expects more is still one the owner would be told to upgrade.

## Seventeen mutations, seventeen caught

Two of them are worth naming because they found defects in my own work rather than confirming it:

- a `HAVING pages > 0` that changed no verdict, because the INNER JOIN beside it already excludes a crawl that stored nothing — dead code reading as the guard;
- a mutation of mine that **passed** because it named a function that does speak, so it tested nothing. Replaced with a genuinely silent one, which went red.

Issue #674, opened 2026-09-06
main is red: the baseline refusal test selects every migration when the chain is two long


muhammadbayoumi:
`main` at `b50bc1e9` fails two tests. Measured in a throwaway worktree checked out from `origin/main` with nothing else in it, so no branch, no working-tree edit and no environment of mine contributes:

```
FAILED tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py
       ::test_a_database_below_the_baseline_is_refused_and_not_replayed
FAILED tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py
       ::test_the_refusal_says_what_to_do_and_names_no_terminal_command
```

**Branch protection requires `test`, so this blocks every merge, not one.**

## The cause is arithmetic that the squash invalidated

The test builds a database deliberately *behind* the baseline and asserts it is refused:

```python
whole = tuple(Migration(n, p) for n, p in dbmod._migration_files())
stop = max(2, len(whole) // 2)
behind._migrations = whole[:stop]
...
with pytest.raises(DatabaseMigrationError, match="no upgrade path"):
```

Measured on `main` right now:

```
migrations on disk    1
_migration_files()    2
declared baseline    17
```

So `stop = max(2, 2 // 2) = 2`, and `whole[:2]` is **every migration there is**. The database is therefore fully migrated rather than behind, nothing refuses it, and `pytest.raises` reports DID NOT RAISE.

`len(whole) // 2` assumes a long chain. Collapsing the chain to a baseline plus one migration left it too short for its own halving, and `b50bc1e9` adding `018_a_job_may_interpret_what_a_crawl_stored.sql` moved the count from 1 to 2 — which is exactly where `max(2, …)` stops protecting it.

## Why it looked green earlier

Full suites run against branches based on `68030ab4` and `fcca90b4` reported 0 failures several times today. Those trees carried a different migration count; the failure appears when the count reaches 2. The test has been fragile since the squash and only now had the input that exposes it.

## What would fix it

Not a number. The test needs a database that is genuinely below the baseline whatever the chain length — build it from the baseline's own `user_version` minus one, or skip when the chain is too short to have a midpoint and say so out loud rather than silently selecting everything. A guard whose subject can vanish while it still passes is the failure mode this repository has spent the day removing.

Filed rather than fixed here because it belongs to whoever owns the squashed baseline, and because three branches are waiting behind it that do not touch this file.

Issue #671, opened 2026-09-06
Seventeen guards that pass against broken code, each proved by mutation


muhammadbayoumi:
A three-lens review of the tests dimension returned **25 findings, 17 of which survived an adversary**. Every one was proved by mutation — the reviewer changed the code, ran the suite, and recorded that it stayed green.

This is the first review of that dimension that actually ran. The first panel generated nine tests findings and attacked none (six agents died on a session limit, and a silent `slice(0, 6)` had already dropped three). The second lost its tests reviewer entirely and still reported `COMPLETE: true`. Both reports were indistinguishable from a clean dimension.

    lenses asked 3 · reviewers returned 3 · dead after retry 0
    findings 25 · judged 25 · never attacked 0 · agents 28, 0 errors

---

## Already fixed — `claude/one-request-path-that-owns-the-status`

**`request()` had no behavioural test** (the only *must fix*). Every guard on it was a source-text check: it could see `throwOnHttpError: false` inside `raw()`, and could not see whether `request()` still honoured the flag. Measured: `const throwOnHttpError = true` left the whole repository green while `raw()` began throwing on the 404s its callers exist to interpret — so an engine too old for `/api/databases/upgrade` would be reported as a generic HTTP error instead of "old engine".

Closed on that branch with six tests that drive the real functions against a stubbed fetch. Three mutations, all caught.

---

## `claude/the-archive-travels-in-pieces`

**The short-read guard has no test** — raised independently by all three lenses. `if (chunk.size !== wanted)` with its own `kind` and its own sentence is one of the two things the chunked upload rests on, and deleting it leaves the suite fully green. Worse, `upload-holds-one-chunk.test.mjs` *fabricates* the error it is meant to provoke: it hands `upload` a source that throws a hand-made `{kind: "short-read"}`. That proves `upload` propagates an error object and nothing about whether `range()` ever produces one.

**The Drive fake ignores its `fields` query** — raised by two lenses. `metadata()`'s correctness rests entirely on `fields: "id,name,size,..."`, because Drive returns only what is asked for. The fake returns whatever the test wrote, so dropping `size` from that query stays green — while `verifyLatest` then rejects **every healthy backup**: "Drive is holding undefined bytes and the backup was recorded as 541531989." That is precisely what the comment beside it says must never happen.

[...]

Issue #657, opened 2026-09-05
HANDOVER — the state of the backup path, and how to continue from here on another machine


muhammadbayoumi:
**Read this first. It is written for a machine that knows nothing about the session that produced it, and nothing in it is a claim that was not measured.**

## THE GOAL IS NOT MET, and this says so plainly

The session's goal was **a first successful backup, taken from a button on Manage account**. The button is built, tested and pushed. **No real backup has ever succeeded** — not on this machine, not anywhere. The engine's folder holds a bundle built today and never uploaded:

```
~/.scrapex/engine/scrapex-bundle-20260906-142150.zip     572.1 MB   built, never uploaded
```

The last step needs a signed-in browser and belongs to the owner. It is three actions, below.

## Where everything is

`main` is `b934450a`. Everything from this session is merged except one open PR.

| | | |
|---|---|---|
| **#735** | the backup button on Manage account | **open**, `9f1854d4`, green locally, 5 files |
| #675 #676 #677 #678 | the helper, the request path, the chunked archive, the red-main fix | merged |

```bash
git fetch origin && git checkout claude/a-backup-button-where-the-backups-are
```

Local and `origin` are identical and the tree is clean. Measured on this commit: **lint 0 errors · node 451/451 · python 3,756 passed, 12 skipped, 0 failed** (python run with `SCRAPEX_FULL_MIGRATIONS=1`, which is how the `migration-authority` check runs it). The branch adds exactly **3** `def test_` over `main` — 3,472 to 3,475, all in `test_panel_dom.py`.

## The half that failed in production is now PROVEN on the real archive

The 2026-09-03 failure was the read, not the upload: *"the engine built an archive of 541531989 bytes and this panel read 0."* That half was measured today against the live engine and the real file, with no Google involved and nothing written:

```
probe        206 · Content-Range: bytes 0-0/599938937 · ETag present
archive      599,938,937 bytes (572.1 MB) — identical to the file on disk
read         144 chunks of 4 MB · every response 206 · 3.4 s
sha256       byte-identical to the .zip on disk
```

**Run through `extension/backend.js`'s own `sourceFor()` and `chunk()`**, not a re-implementation — the exact code that runs when the button is pressed. No short read, no 200, so `If-Range` held across all 144 requests.

[...]

PR #675, opened 2026-09-06
The helper Chrome starts, and it answers inside the five seconds Chrome allows

muhammadbayoumi:
Chrome launches a native messaging host with the calling extension's origin as a positional argument and, on Windows, `--parent-window=<handle>`. `scrapex native-host` accepted neither, so **argparse exited 2 before the host read one byte of stdin**. Measured on 2026-09-03, launching it exactly as Chrome does:

```
scrapex: error: unrecognized arguments:
  chrome-extension://ekcgggphcfdbjgfkcmjagehfjhijeang/
exit: 2
```

The panel said *"Native helper unavailable — restarting through the engine"* and fell back to HTTP, which is why the fallback looked like the feature.

**The registration was never the problem**, and checking it first is what made the real cause findable: the manifest, the `HKCU` pointer and the `.bat` were all correct, and the endpoint refused its own arguments.

## And starting is only half of it

Accepting the arguments meant `_cmd_native_host` reached `_engine_path` for the first time — which calls `DatabaseRegistry.verify()` → `health()` with its default `integrity=True` → `PRAGMA quick_check(1)` plus `pragma_foreign_key_check`. `health`'s own docstring records both as O(FILE SIZE). Measured on the owner's 1,982 MB warehouse, with a crawl running:

| | |
|---|---|
| `registry.engine.path` | **0.0 ms** — everything the host needs |
| `health(integrity=False)` | 17.3 ms |
| `health()` — what it called | **38,157.7 ms** |
| Chrome's budget | **5,000 ms** |

Chrome spawns a fresh host per `sendNativeMessage`, so PING, AUTOSTART_STATUS, SET_AUTOSTART, CHECK_STARTUP and UPGRADE_DATABASE each paid it, and the panel reported *"the helper did not answer in time"* about a helper that was fine. **So this change turned an instant crash into a 38-second one** until path resolution was separated from the health verdict. Now **1.2 ms**.

Nothing is lost: `serve()` opens the warehouse per command and turns a bad one into a framed `engine_unavailable` reply the panel can render. An exit before the first frame gives the panel nothing to say.

## The guards

Both are behavioural. The launch-budget guard makes `health` itself explode and asserts the launch path never reaches it — patching the method rather than one route to it, so any spelling fails. Mutations run and caught: restoring `_engine_path` fails two tests; removing `%*` from the generated launcher fails another.

The launcher test also stopped asserting on an install artifact under the current user's home, which **skipped on any machine that had never installed one** — it now asserts on `scrapex/nativehost.py`'s generator, which this repository owns.

## A citation that had already rotted again

`docs/archive/LESSONS.md` cited `cli.py:1360` for the line that prints the error verbatim. That was true when it was written and the 29 lines this work added above `main()` moved it to **1387**. Re-derived rather than carried, twice — including once more after the rebase, because this exact number went stale twice in one day. `tests/test_the_documents_cite_what_they_claim.py` excludes `docs/archive/` deliberately, which is why a citation inside one can rot in silence.

## Review and state

Split out of `claude/the-panel-read-zero-and-could-not-say-why` on the owner's ruling after a third review pass found more than the second. **This half touches no extension file.**

Measured on this branch alone: **lint 0 errors · node 421/421 · python 3,730 passed**, with two failures that are `main`'s and not this branch's — see #674, where the same two are reproduced on `origin/main` in a clean throwaway worktree. Two review findings against `scrapex/cli.py` are recorded in #671 rather than fixed here.

Part of milestone 15. The handover is #657.


muhammadbayoumi:

```
```

## What actually happened

```
```

## What to do

```bash
```

---

PR #676, opened 2026-09-06
One door to the engine, and it reports what the engine said

muhammadbayoumi:
He pressed **Back up to Drive** on 2026-09-03 and got: *"The engine built an archive of 541531989 bytes and this panel read 0."* The engine had built it, the file was complete on disk with no `.part` beside it, and the panel could say nothing about why — **because the line that read it threw the status away.**

`api()` was the only entry to the engine and it ends in `res.json()`, so two callers that needed a zip used a bare `fetch(...).blob()`. That is the shape of the defect: the one request in the product that moves half a gigabyte was the one request that could not report its own failure.

So `request()` owns the request and `api()` / `bytes()` / `raw()` own the parse. A third caller cannot acquire a bare fetch by needing a different body.

`raw()` exists because two callers need a **refusal to be ordinary**: a 404 from `/api/native/status` and `/api/databases/upgrade` means "this engine is too old", not a failure, and the restart poll asks `/api/health` repeatedly.

## The rationale was wrong the first time I wrote it

The docstrings said those bare fetches had *"no deadline, no abort signal"*. **They had both**, from this file's own `window.fetch` override: `localApiPath` matches any URL under the active backend beginning `/api/`, and every removed call site qualified. What they lacked was the **status check**.

In a repository where these narratives are the durable record, naming the wrong absence sends the next reader to the wrong layer — someone told the override does nothing for `/api/` calls is free to delete it. Corrected here in all three places it was repeated.

## The path is now run, not only read

This closes the one *must fix* a completed review panel returned. Every other guard on this file is a source-text check: it can see that `throwOnHttpError: false` appears inside `raw()`, and **it cannot see whether `request()` still honours the flag**.

Measured before writing the test: replacing the flag's read with `const throwOnHttpError = true` left the entire repository green — 444 node tests and every python guard — while `raw()` began throwing on the 404s its callers exist to interpret. An engine too old for `/api/databases/upgrade` would then be reported as a generic HTTP error instead of "old engine", and the owner told nothing he can act on.

Six tests now drive the real functions against a stubbed fetch. Three mutations, all caught:

| mutation | result |
|---|---|
| the opt-out ignored | caught |
| the engine's `detail` dropped | caught |
| a 2xx that is not 200 treated as an error | caught |

**The third one missed at first**, because the test used `raw()` — which opts out of the status check, so it could not have failed whatever `request()` did. Found by running the mutation rather than by reading the test, and rewritten to go through `bytes()`. The reason is recorded in the test.

## Review and state

Split out of `claude/the-panel-read-zero-and-could-not-say-why` on the owner's ruling, **as the foundation half**. `range()` and `sourceFor()` are parsers over this `request()` and follow in their own branch on top of this one.

One consequence worth naming: `raw()`'s comment says the opt-out is *"here and nowhere else"*. On the unsplit branch that was **false** — a finding the second review panel raised. On this branch it is true, because the two other users are not here.

Measured on this branch alone: **lint 0 errors · node 427/427 · python 3,727 passed**, with two failures that are `main`'s and not this branch's — reproduced on `origin/main` in a clean worktree in #674.

Part of milestone 15. The handover is #657; the review findings are #671.


muhammadbayoumi:

```
```

## What actually happened

```
```

## What to do

```bash
```

---

Issue #619, opened 2026-09-05
fetchPanelPack has no bundle_format gate, and repeats the refusal the other two now share


muhammadbayoumi:
Measured on `claude/the-panel-read-zero-and-could-not-say-why` while collapsing a different duplication.

`extension/drive.js` had the same bundle-format refusal written twice, in `verifyLatest` and `fetchLatest` — identical condition, identical `"wrong-format"` kind, differing only by a trailing `" Nothing was restored."`. That pair is collapsed into one `readableLatest` helper on that branch.

**`fetchPanelPack` is the third case and it is left alone deliberately**, because fixing it is a behaviour change rather than a tidy-up:

- It duplicates the `"No backup has been uploaded from any device yet."` refusal verbatim (`extension/drive.js`, in `fetchPanelPack`).
- It reads `bundle_format` **zero times** — measured, not assumed. So unlike the other two it has no format gate at all.

Routing it through `readableLatest` would silently ADD a format gate it does not have today. That is a real improvement and it is exactly what Slice 2 of the read plan (#597) proposes, so it belongs there where someone is thinking about it, not smuggled in through a de-duplication.

**Note for whoever picks up Slice 2:** by the time the chunked-upload branch lands there will be ONE gate in `drive.js`, not two. A third copy inside `fetchPanelPack` would be the change amplification CLAUDE.md's DRY rule names — same knowledge, same reason to change, and a `BUNDLE_FORMAT` bump that misses one leaves a path silently accepting what the others refuse.

Issue #597, opened 2026-09-04
The plan: the panel reads the pack the engine already wrote


muhammadbayoumi:
## The verdict, and the measurement that forced it

Buildable — but not as "read the database". Buildable as "read the export the engine already writes", and that inversion is forced by one measurement I took on the live file rather than read about.

THE MEASUREMENT THAT DECIDES IT: the warehouse is a WAL database and a browser cannot apply a WAL. `C:\Users\Muhammad\.scrapex\engine\scrapex-engine.db` is 15,499,264 bytes, header bytes 18/19 read `2 2` (WAL, not rollback journal), and a `-wal` of 4,260,112 bytes sits beside it right now. `scrapex/db.py:93` sets `PRAGMA journal_mode = WAL` on every connection, so this is the steady state, not an accident. `spikes/opfs-sqlite/FINDINGS.md:33` records that no OPFS VFS implements `xShmMap` — and `FINDINGS.md:117-126` shows the SAH pool silently rewriting bytes 18/19 to `1,1` on import. So a browser handed that file reads a database missing 4.26 MB of committed transactions, with no error and no way to know. The only thing that can produce a WAL-free consistent copy is `sqlite3`'s backup API, which is the engine — the broken thing. In BOTH of his scenarios there is nothing on the machine that can make the live file readable.

This is not the concurrency constraint and he is right that concurrency does not apply — neither scenario has a writer. It is a different constraint, it is not in the spike's four, and it kills the local-`.db` route on its own. The spike's speed verdict (1.4–1.6x, `FINDINGS.md:315`) is true and irrelevant: an engine that runs the query fast against the wrong bytes is worse than no engine.

[...]

PR #677, opened 2026-09-06
The archive travels in pieces, and the check asks Drive instead of downloading it

muhammadbayoumi:
> **Based on #676, not on `main`.** `range()` and `sourceFor()` are parsers over the `request()` that PR adds, so this cannot stand alone and is not pretending to. GitHub will retarget it to `main` when #676 merges.

## Holding the archive whole is what failed

A Chrome side panel asked to keep **541,531,989 bytes** in one Blob read **0**. A 378,655,878-byte build had worked four days earlier, so the ceiling sits between them and moves with the browser rather than with this code.

`upload` was already chunked and resumable — it sliced its 4 MB chunks out of a fully buffered Blob. **So the fix is not a new upload path: it is that the thing being sliced no longer has to exist all at once.**

**The size comes from the engine, not the manifest**, and that is the whole care in `sourceFor`. `expectSize` compares what the engine *described* against what *arrived*, and on 2026-08-30 those resolved to different builds and a 0-byte archive reached Drive under a pointer carrying the real one's digest. Sizing the source from the manifest would make both sides of that comparison the same number and the guard would pass on anything.

## And the archive could be spliced from two builds

`/api/bundle/archive` re-resolves "the newest zip on disk" on **every** request, and ~130 requests over minutes leave a window: a second panel window taking a backup finishes a new build, and every later chunk comes from a different file.

**No guard could see it.** When the new archive is *longer*, every chunk is still exactly the size asked for, the total still matches the pointer, and the check button certifies it complete. It would be found at restore, months later, when the central directory of one build does not match the offsets of another.

Closed with `If-Range` plus a `Content-Range` total comparison — both measured against the live engine rather than assumed. A one-byte probe answers:

```
206 Partial Content
Accept-Ranges: bytes
ETag: "c9d60b0b9a6ebd8b8d6e89b2612898fd"
Content-Range: bytes 0-0/541531989
```

## The check button never restored anything

Its own comment said so, and it **downloaded** the archive to look at `archive.size`. Asking Drive is the cheaper check and the **stronger** one: a downloaded size catches a truncated *download*, and the failure that actually happened was a truncated *upload* — a 0-byte archive reached Drive with a pointer describing it as whole. The label stops promising a restore it has never performed.

## One format gate, not two

`verifyLatest` and `fetchLatest` each carried their own copy of the same refusal — identical condition, identical `"wrong-format"` kind, differing by one trailing sentence that is context rather than knowledge. `readableLatest` holds it once and absorbs a duplicated no-backup refusal beside it.

`fetchPanelPack` is **deliberately untouched**: it has no format gate at all, so routing it through would be a behaviour change belonging to the read plan (#597), not a de-duplication. Recorded as #619.

## A failing engine is not a range problem

`sourceFor` briefly mapped every non-206 to "this engine will not serve byte ranges", and `app.js` answers that by reading the archive whole — so a transient 500, or the 416 Starlette returns for `bytes=0-0` on an empty file, **would have reinstated the half-gigabyte read this work exists to remove.**

The fix was a deletion. Measured in node: `Response.ok` is true for 200 and 206, false for 416 and 500 — so neither parser ever needed to read the status itself, and taking the opt-out is what caused it. Both now let `request()` throw, with the engine's own `detail`.

## Nothing in this repository had ever sent a Range header

Three engine-side tests now send one, reassemble the archive from its chunks, and prove the `If-Range` behaviour the swap detection rests on. Breaking `FileResponse` fails five of them.

## What is NOT fixed here, and why

A completed review panel left **sixteen should-fix findings against this half** — the short-read guard has no test, the Drive fake ignores its `fields` query so dropping `size` from that query stays green while rejecting every healthy backup, the check button's call site is unpinned. Each is proved by a mutation and recorded in **#671**.

They are not fixed here on purpose: **the reason this branch exists is that one change carrying all of it was too big to review**, and fixing them into it rebuilds what the split took apart.

## State

Measured on this branch alone: **lint 0 errors · node 450/450 · python 3,730 passed**, with two failures that are `main`'s and not this branch's — reproduced on `origin/main` in a clean throwaway worktree in #674, which blocks every merge until it is fixed.

Part of milestone 15. The handover is #657.


muhammadbayoumi:

```
```

## What actually happened

```
```

## What to do

```bash
```

---

muhammadbayoumi:
**`#676` is merged, so the commit this branch is stacked on is now on `main` under a different SHA. Your rebase needs to drop your copy, and that is worth knowing before you run it.**

```
origin/main   d7ecedc3 -> 74caea5c
              One door to the engine, and it reports what the engine said (#676)
mergedAt      2026-09-06T11:48:21Z
```

## Why this rebase is not the ordinary kind

This branch was stacked: `git merge-base --is-ancestor 4524eb9d 128c79da` was true, so it carried **two** commits ahead of `main` — `#676`'s and its own.

`#676` was **squash-merged**, which is this repository's convention. A squash produces a *new* commit with the same content and a different SHA. So `74caea5c` on `main` and `4524eb9d` on this branch are the same change and git cannot tell that from the hashes.

**What that means in practice:** on rebase, your copy of `4524eb9d` should drop out as already-applied. If it does not, you may see conflicts in the five files the two slices share:

```
eslint.config.mjs
extension/app.js
extension/backend.js
extension/tests/the-request-path-owns-the-status.test.mjs
tests/test_no_request_to_the_engine_escapes_the_request_path.py
```

Those are not real conflicts between two ideas — they are one change meeting itself. Resolve toward what is already on `main` for anything that belongs to `#676`, and keep only what is genuinely this slice's.

**Afterwards, confirm the branch is only its own work:**

```bash
git log --oneline origin/main..HEAD          # expect ONE commit, not two
```

If two commits remain after the rebase, the squashed one did not drop and the branch is carrying a duplicate.

## One thing I could not verify, and it is yours to check

At the last read this PR showed **10 check rows** where `#675` and `#676` each showed 14. The difference is the CodeQL family — `CodeQL` plus the three `Analyze` jobs were absent from its row set, leaving only the five `ci.yml` checks across two runs.

`CodeQL` is a **required context** under the branch protection now on `main`. GitHub reported `CLEAN` anyway, which I did not act on: a required check that never reports and a `CLEAN` verdict cannot both be right, and the merge session does not merge on a status it cannot reconcile.

It may simply be that the CodeQL run had not attached yet when I read it. **After the rebase, confirm all nine check names are present** before calling it ready:

```bash
gh pr checks 677 | awk -F'\t' '{print $1}' | sort -u
```

Expect: `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`, `CodeQL`, `contract-parity`, `lint`, `migration-authority`, `scope`, `test`.

## And the earlier red was not yours

Confirmed and now closed: the `test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py` failures that hit all four open PRs were `main`'s, from a fixture that had been passing vacuously since the squash. `#678` fixed it and `main` is green at the job level.

Post the settled read when the rebase is done and the row set is complete.

muhammadbayoumi:
**GitHub is reporting this PR as `CLEAN` and it is not. I measured it rather than read it, and it fails three ways. Not merging.**

Measured against `origin/main` at `74caea5c`:

```
head                    128c79da   (unchanged — the rebase has not happened)
contains origin/main    NO
commits ahead of main   2          (expected 1; the squashed duplicate is still here)
check names present     5 of 9     (CodeQL and the three Analyze jobs absent)
```

And GitHub says:

```
mergeable        true
mergeable_state  clean
```

Those cannot both be right. Branch protection is live and correct — I re-checked it while investigating: `protected: true`, `strict: true`, and `CodeQL` is in the required contexts. So a branch that is behind `main` and missing a required check should not read `clean`.

The most likely explanation is that `mergeable_state` is computed asynchronously and has not recomputed since `#676` merged minutes ago. **Either way the merge session does not act on it**, because `mergeStateStatus` describes whether git can merge the branch, never whether the branch is correct — and here it is not even accurate about the base.

## What still needs doing

**The rebase has not happened yet.** The head is the same SHA it was before `#676` landed. Everything in my previous comment still applies:

- `#676` was squash-merged, so your copy of `4524eb9d` is the same content as `74caea5c` under a different SHA. On rebase it should drop as already-applied.
- After rebasing, `git log --oneline origin/main..HEAD` must show **one** commit. Two means the duplicate survived.
- Then confirm the row set is complete before calling it ready:

```bash
gh pr checks 677 | awk -F'\t' '{print $1}' | sort -u
```

## This also fixed a defect in my own instrument, which is worth saying

My watch reported this PR as ready. It was reading GitHub's `mergeable` and `mergeStateStatus` and trusting them — the exact mistake I have been flagging to other sessions all day, sitting inside the tool I use to catch it.

It now **measures** two things itself instead of asking:

- `git merge-base --is-ancestor origin/main <head>` — is the branch genuinely on top of `main`, rather than merely reported as clean.
- whether every required context actually appears in the row set, rather than trusting the aggregate verdict.

I ran it against this PR as a control before trusting it, and it correctly refuses on both counts. A guard is untrusted until the defect it names makes it red.

muhammadbayoumi:
**`main` moved again and this one overlaps your branch in files, so your rebase is no longer only about the base.**

```
origin/main   74caea5c -> 25e63535
              A Database page of its own, above Engine, and the upgrade moved to it (#679)
mergedAt      2026-09-06T12:12:32Z
```

`#679` changed seven files, and these are the ones it shares with your branch:

```
extension/app.html
extension/app.js
tests/test_panel_dom.py     (this one only if your branch touches it)
```

It moved the upgrade control out of Engine and onto a Database page of its own, so the panel's structure around those controls has changed — not just the lines' positions.

**What that means for you:** the earlier rebases today were mechanical because the changes were disjoint. This one may not be. Re-derive rather than assume, and if you cite a line number in `extension/app.js` anywhere, recompute it against `25e63535` rather than carrying it.

**The check that proves your rebase moved the base and nothing else:**

```bash
```

Identical before and after is a stronger statement than "no conflict". If it differs, the rebase changed your content and you want to know why before you report a green.

## Two things to confirm before calling it ready, both learned today

**One — measure the base, do not read it.** GitHub reported `mergeable_state: clean` on a PR that was two commits behind with a required check missing. `mergeStateStatus` describes whether git can merge the branch, never whether the branch is correct, and it is computed asynchronously so it lags a merge by minutes. Measure it:

```bash
```

**Two — confirm the row set is complete, not just green.** A PR can show all-pass while a required context never reported at all. Expect nine names:

```bash
gh pr checks <n> | awk -F'\t' '{print $1}' | sort -u
```

And confirm the `test` row was the engine tier rather than a documentation-tier pass wearing the same name — check the run's step conclusions, preferring the `pull_request` run.

muhammadbayoumi:
**Retracting one line from my earlier comment: I told you to expect ONE commit after the rebase and that two would mean the duplicate survived. That is no longer true, and by my own measurement your branch is correct.**

Measured at `8bd7f76c` against `origin/main` at `1b34ed50`:

```
8bd7f76c  The bare-fetch detector was reading prose, and it failed on somebody else's comment
e2a9fb7c  The archive travels in pieces, and the check asks Drive instead of downloading it
```

Two commits, and **neither is `#676`'s**. The squashed duplicate dropped exactly as it should have, and you have since added a second commit of your own. So "two commits" now means "two of your own commits", not "the duplicate survived". My rule was written for the state the branch was in an hour ago and I should have scoped it to the SHA rather than to a count.

**And the false `CLEAN` is gone.** Your PR now reads honestly:

```
mergeStateStatus  BEHIND      (was CLEAN while 3 commits behind)
behind            1
CodeQL            present     (was absent)
Analyze           3 rows      (were absent)
rows              14 pass, 0 fail, 0 pending
```

That confirms the mechanism I filed as #682 from the other direction: the same PR flipped from a false `CLEAN` to a correct `BEHIND` at the moment the CodeQL rows attached. Your branch was never the problem — the status was.

## What is left

One rebase. You are one commit behind because `#675` merged after your re-run:

```
1b34ed50  The helper Chrome starts, and it answers inside the five seconds Chrome allows (#675)
```

It touched `scrapex/cli.py`, `tests/test_the_native_host_starts_the_way_chrome_starts_it.py` and one citation line in `docs/archive/LESSONS.md` — no overlap with your files, so this rebase should be mechanical.

After it, the two checks worth running before you call it ready:

```bash
gh pr checks 677 | awk -F'\t' '{print $1}' | sort -u   # expect nine names
```

Nine: `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)`, `CodeQL`, `contract-parity`, `lint`, `migration-authority`, `scope`, `test`.

Your second commit — a bare-fetch detector that was reading prose and failing on someone else's comment — is the same shape as three other things found today: a guard that fires on correct code. Worth saying so in the PR body if it is not already there.

PR #680, opened 2026-09-06
The panel says what the engine says about its own update

muhammadbayoumi:
## What this is

The read half of the doorless updater, split out of #608.

`GET /api/update` has been mounted, complete and tested on the engine side since
PR 246. `git log -S "api/update"` over `extension/` returned **no commit in the
whole history** — the route had never been called by anything. This calls it and
puts the sentence it answers with on screen.

Nothing here starts an update. Pressing the button is #608, stacked on top of
this.

## Why it is split

#608 carried the reading and the pressing together. Its second critical review
returned **four must-fix and nine should-fix — more than its first**, which is
what "does not converge" looks like, and `CLAUDE.md` says to split at that
point rather than push a third pass at the same diff.

The two halves fail differently. This one can only paint a wrong sentence. The
other can leave a control that lies about work in flight — a disabled button
over a download nobody is watching, a poll that keeps asking for a screen the
reader left. Every must-fix in that review was in the second kind, and a
reviewer holding both at once reviewed neither.

## The change

- **`engineUpdateState` normalises at the seam and nowhere else.** The engine
  nests `phase`, `progress` and `detail` under `progress_state` and names the
  refusal `self_update_blocked_because`. The flatten happens where the response
  arrives, so no call site reads the nested shape and no two copies can drift.
- **`engineUpdateSentence` renders the reason the engine gave**, never a bare
  "no update available". `can_self_update` is false for causes that ask opposite
  things of him — a source checkout is not a fault and its remedy is git; a
  release with no digest is the publisher's and there is nothing for him to do —
  and `update_api` keeps the field separate saying the panel "must be able to say
  WHICH".
- **`#engine-update-state` is hidden for a candidate engine.** `#view-engine-detail`
  is shared by all seven, so an unguarded un-hide paints ScrapeX's sentence onto
  Scrapy's page. Written only while ScrapeX's own screen is the one on show.
- **`updateReport: 8000`** bounds `GET /api/update` above the engine's own 4 s
  manifest fetch, so the engine is always the side that gives up first and can
  say why (`R-48` rule 2).

## The test that had to exist

`tests/test_the_panel_reads_the_keys_the_updater_sends.py` asserts the seam
against the **real router**, not a fixture — because a fixture is exactly what
went wrong. Five panel tests were green against a body the engine cannot
produce, and a five-of-five mutation check proved only that the two halves
matched each other, not that either matched the engine.

It derives what the normaliser adds from the normaliser's own source. An earlier
draft listed it as a constant, and both repairs could then be deleted with every
test still green: a hand-written map compared against a hand-written map asserts
that two people agree, not that the code does.

## And one existing guard had to be sharpened — which the split found

`test_the_checksum_on_screen_says_who_checks_it` tied the installer note's claim
to `"/api/update" in app.js`, which is true the moment the panel merely **asks**
the engine what it can do. So a read-only caller made the guard **demand** the
note promise engine-side verification — the exact false promise `OP-124`
removed, re-required by the test written to prevent it.

It now keys on `post("/api/update"`, because only starting an update makes the
note true. The note's own wording is unchanged here and gains its new text in
the half that earns it. Broken on purpose and watched to fail before it was
committed.

The unsplit change never surfaced this: it added the POST in the same diff, so
the guard stayed green over a condition that had quietly become wrong.

## Gates

`python -m pytest -m "extension or docs"` — exit 0 on `f6f881d4`, the same
selection CI runs for an extension-scope change.


muhammadbayoumi:
**Returning this to its session: it conflicts with `main`. GitHub reports `mergeable: CONFLICTING`, and the merge session does not resolve conflicts on someone else's branch.**

```
mergeable     CONFLICTING
mergeState    DIRTY
head          f6f881d4
origin/main   25e63535
```

**The conflicting file:**

```
tests/test_panel_dom.py
```

That is not surprising and it is worth knowing why. `#679` merged a few minutes ago — *"A Database page of its own, above Engine, and the upgrade moved to it"* — and it changed `extension/app.html`, `extension/app.js` and `tests/test_panel_dom.py`. It **moved the upgrade control out of the Engine page onto a Database page of its own**, so the panel's structure around those controls changed, not merely the line positions.

Your PR is titled *"The panel says what the engine says about its own update"*, which is the same surface. So this is a genuine content collision between two changes to the same area, not an accident of formatting.

**Rebase onto `25e63535` and resolve it against what `#679` actually built** — the upgrade lives on the Database page now, so a test asserting it on the Engine page is asserting the old structure.

## Before you call it ready

**Measure the base, do not read it.** GitHub's `mergeStateStatus` is computed asynchronously and lags a merge by minutes. Today it reported `clean` on a PR that was two commits behind with a required check missing:

```bash
```

**Confirm the row set is complete, not merely green.** Expect nine names — yours currently has all nine, which is good and worth keeping an eye on after the rebase:

```bash
gh pr checks 680 | awk -F'\t' '{print $1}' | sort -u
```

**Confirm the `test` row was the engine tier**, not a documentation-tier pass wearing the same name — read the step conclusions of the `pull_request` run.

**And after rebasing:**

```bash
```

A different hash before and after means the rebase changed your content, which is exactly what you want to know before reporting a green.

Issue #685, opened 2026-09-06
A running interpret job was entered three times, and its progress bar was reset twice mid-pass


muhammadbayoumi:
One `dataset_interpret` job on his live warehouse logged its **entry preamble three
times** while a single pass was running. Read from `job_log_entry`, job 158, in id order:

| log id | at | line |
|---|---|---|
| 9972 | 14:01:15 | `interpreting the stored pages of job-job_6eb28381bf56 — 6,713 page(s) on disk` |
| 9973 | 14:01:21 | `approve job-job_6eb28381bf56: 909 page(s) on disk` |
| **9974** | **14:06:51** | `interpreting the stored pages of …` **again** |
| **9975** | **14:07:05** | `interpreting the stored pages of …` **again** |
| 9976 | 14:07:16 | `approved 908 page(s): …` |
| 9980 | 14:07:16 | `interpretation finished in 6.0 min, 909 page pair(s) read from disk` |

That line is written once per entry to `run_dataset_interpret_job_once`, immediately
after the call sets the job `preparing` with `progress_done=0`. So the job was entered
three times, and the two extra entries landed **five and six minutes into a pass that
had announced itself at 14:01:21 and did not report until 14:07:16**.

`approve` announced itself and reported **once**, and the outcome is correct — 908 pages
ingested, 464 contractors created, nothing running afterwards (no writes after
14:07:16, no non-terminal interpret job). **No data was harmed this time.**

## Why it still matters

Each entry runs, before any work:

```python
jobs._update(conn, job["job_id"], status=JobStatus.PREPARING.value,
             stage=JobStage.PREPARING.value, progress_done=0, …)
```

So **his progress bar was reset to zero twice, five minutes into a six-minute run** —
the panel's Run screen showed a job going backwards. And an entry that got past the
preamble on a longer run would start a **second concurrent `approve` pass over the same
909 page pairs**, on a second connection, against the same rows.

## The suspect, named as a suspect

`jobs.reclaim_orphaned_jobs` (`scrapex/jobs.py:1150`) flips **every** `preparing` /
`running` / `resuming` job to `queued` unconditionally — no heartbeat test, no age test.
Its docstring justifies exactly that with a startup premise:

> *"Only ONE worker ever executes, so at startup any in-flight job is ours and nobody
> else's — nothing can be legitimately running."*

True at startup. But `_loop` also calls it **mid-life**, on the reopen path:

```python
if conn is None:
[...]

Issue #684, opened 2026-09-06
One interpret job logs 6,713 pages then 909, and neither line says what it counted


muhammadbayoumi:
Two lines from one interpret job, six seconds apart, on his live warehouse:

```
14:01:15  Saudi Contractors Authority: interpreting the stored pages of
          job-job_6eb28381bf56 — 6,713 page(s) on disk. Nothing is fetched
14:01:21  approve job-job_6eb28381bf56: 909 page(s) on disk
```

**Same label, same run ref, two numbers, and the second is 13% of the first.** Read in
order it says 5,804 pages were dropped between one line and the next. Nothing was
dropped — the two lines count different things and neither says which.

## What each number actually is

Measured read-only on the live warehouse, 2026-09-06:

| | |
|---|---|
| snapshot **rows** under that ref | **6,713** |
| distinct `source_url` | **1,818** |
| English rows / Arabic rows | 3,357 / 3,356 |
| **page keys after collapsing the locale pair** | **909** |

`1,818 ÷ 2 = 909`, so the second number is right and it is the one that describes the
work: `contractors.approve` reads an en/ar **pair** as one page.

And the 6,713 is inflated by the sweep's own passes — the same URLs stored once per
pass:

| pass | rows stored |
|---|---|
| a1 | 1,818 |
| a2 | 1,546 |
| a3 | 1,260 |
| a4 | 1,138 |
| a5 | 761 |
| a6 | 190 |

That is by design: the listing reorders, so each pass re-reads it, and `approve`'s "LAST
ONE WINS" comment already handles the duplicates. The defect is only that
`latest_crawl_run_ref` reports `count(s.page_snapshot_id)` and calls it "page(s) on
disk".

## Where

- `scrapex/datasetjob.py` — `latest_crawl_run_ref` returns the row count, and
  `run_dataset_interpret_job_once` prints it as `— {pages:,} page(s) on disk`.
- `scrapex/contractors.py` — `approve` prints its own count of collapsed keys with the
  same words.

## Options

1. **Name what each counts.** `6,713 stored reading(s) across 6 passes` and
   `909 page(s) to interpret`. One line each, no arithmetic changes. Effort: small.
2. **Report only the number that describes the work.** Have `latest_crawl_run_ref`
   return the collapsed count so both lines agree. More honest, and it costs a second
   query over the same rows — the count is per source and the collapsing rule lives in
   `contractors`, so this moves knowledge out of the module that owns it.
[...]

Issue #682, opened 2026-09-06
A required check that never attaches does not block the merge — GitHub reports the PR clean


muhammadbayoumi:
**A required status check that never attaches to a pull request does not block it — GitHub reports the PR as `clean`. Measured today on `#677`, which is the only open PR missing `CodeQL` and the only one GitHub calls clean.**

Recorded by the merge session. Not fixed here — it is a repository-settings question and the owner's call.

## The measurement

Branch protection on `main` was enabled today with six required contexts: `lint`, `scope`, `test`, `migration-authority`, `contract-parity`, **`CodeQL`**. Verified live: `protected: true`, `strict: true`.

Across the four open pull requests at the time of writing:

| PR | `mergeStateStatus` | `CodeQL` row | `Analyze` rows | on top of `main` |
|---|---|---|---|---|
| #680 | `DIRTY` | 1 | 3 | no |
| #675 | `BEHIND` | 1 | 3 | no |
| #608 | `UNSTABLE` | 1 | 3 | no |
| **#677** | **`CLEAN`** | **0** | **0** | **no** |

`#677` is the only PR whose row set is missing the required `CodeQL` context, and it is the only one reported as `CLEAN`. It is also **two commits behind `main`** while being called clean, which no `strict` protection should allow.

```
gh api repos/muhammadbayoumi/ScrapeX/pulls/677
  mergeable        true
  mergeable_state  clean
```

## Why this is the dangerous direction

Every false-green in this repository has one shape: **the check's input disappeared and the absence read as a pass.** A suite killed at 40% prints no failures. `grep -c FAILED` on an empty file returns 0. A `jq` that was never installed returns empty fields.

This is that shape applied to branch protection itself. The requirement is not "CodeQL passed" — in practice it resolves to "no CodeQL row says otherwise". A check that never runs cannot say otherwise.

**A protection that can be satisfied by a check's absence protects nothing in exactly the case where you most want it to.**

## What is not yet established

I do not know *why* `CodeQL` never attached to `#677` specifically, and I want to be explicit about that rather than guess:

- CodeQL here is **default setup**, not a workflow file: `state=configured`, languages `actions, javascript, javascript-typescript, python, typescript`, `schedule=weekly`.
- It attaches correctly to the other three open PRs, so it is not broadly broken.
- `#677` is a **stacked** PR whose head (`128c79da`) has not moved since before two merges landed, so one plausible explanation is that its analysis was superseded and its rows aged out rather than never having run.

[...]

PR #688, opened 2026-09-06
The corruption scan is asked for, not drawn on every page open

### LINKED ISSUE #683: The storage route runs a 2 GB integrity scan to draw a status card, and misses its deadline every time
`GET /api/storage` cannot answer inside the deadline the panel gives it, on his warehouse, and the new Database page shows the consequence on every load: **Schema: unreadable**, empty Health, empty Size on disk, empty Backups, and `The request exceeded its 5000 ms deadline.`

The page degraded exactly as designed — the Upgrade control stayed enabled and named the native host — but it degrades *always*, so the page states nothing.

## Measured, 2026-09-06, on the live warehouse (2,080,395,264 bytes, read-only)

| | |
|---|---|
| `GET /api/storage` end to end | **7.81 s** |
| the deadline it is given (`STARTUP_DEADLINES.destinationData`) | **5,000 ms** |
| `storage.health()` alone | **5.73 s** |
| everything else — path, sizes, schema, backups | **0.02 s** |
| `_schema_facts` (added by #679) | **0.004 s** |

So one call is the whole cost, and the rest of the page is instant. Inside `health()`
(`scrapex/storage.py:648-650`), measured separately on the same file:

- `PRAGMA quick_check` — **2.89 s**
- `PRAGMA foreign_key_check` — **1.50 s**

Both are full scans of a 2 GB file. `/api/health` by comparison answers in **0.36 s**.

## Why this is a shape problem, not a number problem

An integrity scan is something you *ask for*, not something a page does on open. And the
cost grows with the warehouse: it was inside 5 s when the file was smaller, so raising
the deadline buys time and not a fix — the same reservation `STARTUP_DEADLINES` already
records beside `bundleBuild`, which is the one deadline there derived from a measurement
rather than chosen.

Note the panel is not the only caller: the Settings screen's `loadStorage` reads the same
route, so it has been failing on his machine too — quietly, four rows down a settings
screen.

## Options

1. **Split the integrity verdict out of the route.** `/api/storage` answers with the
   cheap facts (0.02 s); health becomes its own request, or a control the page offers.
   The page then states the version, the sizes and the backups immediately and says when
   integrity was last checked. Effort: moderate. Blast radius: two callers of `health`.
2. **Cache the verdict with its timestamp.** The page shows `checked N minutes ago` plus
   a Re-check control. Cheaper than 1, and it keeps one route. Risk: a cached "healthy"
   outliving the fact.
3. **Raise `destinationData`.** One number. Makes every load of two screens wait ~8 s,
   and expires again as the warehouse grows.
4. **Do nothing.** The page keeps saying `unreadable` on his machine, and so does
   Settings.

Which one it is decides what a screen offers, so it is his call.

muhammadbayoumi:
Closes #683.

His Database page showed **`unreadable` over three empty cards, on every open** — and it
was right to, because `GET /api/storage` could not answer inside the deadline the panel
gives that path. Screenshot in #683.

## Measured on the live warehouse, 2,080,395,264 bytes

| | |
|---|---|
| `GET /api/storage` end to end | **7.81 s** (10.26 s on a second reading) |
| the deadline that path gets (`STARTUP_DEADLINES.destinationData`) | **5,000 ms** |
| `storage.health()` alone | **5.73 s** |
| → `PRAGMA quick_check` | 2.89 s |
| → `PRAGMA foreign_key_check` | 1.50 s |
| everything else — path, sizes, schema, backups | **0.048 s** |
| **after this change** | **0.060 s** |

So the route was not slow — it **failed**. And the Settings screen's four rows had been
failing the same way before the page existed, quietly, four rows down a settings screen.
Both pragmas are O(file size), so raising the deadline buys time and not a fix.

## The split is a rule this codebase already stated

`EngineDatabase.health` took the same keyword, with the same name, off the same kind of
measurement at 1,067 MB. And `storage._warehouse_identity` says in terms: *"Integrity
and identity are deliberately separate checks."*

`storage.health` asked **both** questions under one name and charged a full file scan for
it. **This is the caller that was missed, not a new idea.**

## Two claims, never confused

`integrity_checked` rides on **every** verdict — all five exits, not the happy one —
because "healthy" after no scan is a **narrower claim** than "healthy" after one. A
caller branching on it would get `None` from any path that forgot to stamp it, and
`None` is neither claim. The sentence narrows with it:

> Readable, at the expected version, and the right kind of file. **Corruption has not
> been checked.**

**And a finding nobody cleared still counts.** The routine verdict stopped looking for
corruption, so on its own it would report a file the owner was told last week was
damaged as ready — "Enabled" on the same screen that said "damaged". The last **wide**
verdict is the last thing anybody actually learned about the file, and it holds until
another check replaces it. A clean check releases it. Both directions are tested,
because a rule that only ever says no is not a rule.

## `POST /api/storage/integrity`

**POST because it records what it found.** Without a recorded moment, a card showing no
problems is indistinguishable from a card that never asked — so the page can say *when*
corruption was last looked for. New state key `storage_integrity`, separate from
`storage_last`, which is the last storage **action**; conflating them is the same class
of error as the Backups row below.

**The scan does not hold the write lock**, deliberately rather than incidentally.
`health` opens its own read connection, so 5.9 s under the lock would stall a running
crawl's writes for the sake of a status card. The lock is taken for the one small write
at the end.

**The deadline is derived, and its position in the table is part of it.** The `storage`
rule ends `(?:[/?]|$)`, so `/api/storage/integrity` matches it too and
`deadlineForLocalRequest` returns the **first** match — written one line lower it would
collect the 5,000 ms this change exists to escape, and the panel would abort a scan the
engine goes on to finish. That is the failure `/api/bundle` already paid for once in
this same table. `integrityScan: 120000` covers about 40 GB at the measured 2.8 s a
gigabyte, and a test case names the ordering.

## The control is the opposite of the Upgrade button beside it

The difference is real rather than a matter of taste:

- `upgradeDatabaseFromPanel` falls back to the native host, so it survives a dead engine
  and **stays enabled**;
- the scan has **no fallback** — the native host does not read the warehouse — so it is
  **disabled** when the engine is silent, and the note says why.

A button that cannot work is worse than no button, in both directions.

## Two defects of mine in the page, found on the way

- **The Backups row read `s.last`** — `storage_last`, the last storage *operation*, which
  a compaction and a repair write too. Measured: it named a manual backup from
  2026-08-30 as the newest while the newest copy on disk was the pre-upgrade one taken
  2026-09-06 — six days later, and the one whose existence answers *"is it safe to press
  Upgrade"*. The size printed beside it was `backup_bytes`, the **total** across all four
  copies, so one 2 GB file was labelled 6.9 GB. It reads `backups[0]` now, documented
  newest first.
- **`wireRuntimeRepair`'s comment still said `#runtime-note` "stays in Settings".** I
  moved it to the Database page with the button and left the sentence describing where it
  used to live.

## Three lies in the harness, all of which would have let a guard pass

- It declared `backup_count: 2` and served `backups: []` — every DOM test would have
  asserted against "no backup" while the card claimed two.
- It had no `integrity` field, so the whole Corruption row would have rendered its empty
  state in every test.
- **`POST /api/storage/integrity` was answered with `/api/storage`'s payload**, because
  the route lookup took the first key that was a *prefix* rather than the most specific
  one — so a caller read a plausible `status` for a route the stub never declared. Any
  nested pair of paths had the same hole; it sorts by specificity now.

## Tests

**Ten mutations, ten caught** — including the deadline rule moved one line down, and the
harness's prefix lookup put back.

- **9 engine-side** (`tests/test_the_integrity_scan_is_asked_for.py`). The strongest
  plants a **real foreign-key violation** and asks one file two questions: the wide call
  says `damaged`, the narrow says `healthy`. If the narrow call had found it too, the
  flag would be skipping nothing and the saving would be zero.
- **4 DOM guards**: the Corruption row says "not checked" *and why it is not run on
  open*; the press really sends the POST and dates the answer; the control is withdrawn
  when the engine is silent while the upgrade beside it is not; the Backups row names
  the newest **copy** and its own size.
- **3 node cases** for the deadline table, including one that exists only to hold the
  rule's position.

Local gate on `a1287998`: `ruff check scrapex/` clean, `eslint extension/` clean,
`node --test extension/tests/*.test.mjs` **428 passed**. The full Python suite in both
template and `SCRAPEX_FULL_MIGRATIONS=1` modes is running as this opens; the result goes
in a comment.

Filed while measuring this and **not** fixed here: #684 (one interpret job logs 6,713
pages then 909, and neither line says what it counted) and #685 (a running interpret job
was entered three times and its progress bar was reset twice mid-pass).


muhammadbayoumi:
**Not mergeable, for exactly one reason: the branch is behind `main`. The work itself is complete — nothing here needs fixing, diagnosing or reviewing again.**

Measured by the merge session at `a99880fa` against `origin/main` at `2a046ae0`:

```
mergeable                 MERGEABLE
mergeState                BEHIND
on top of main (measured) NO — behind 2
rows                      14 pass · 0 fail · 0 pending
required contexts         all nine present, CodeQL included
```

Branch protection on `main` has `strict` enabled, so **GitHub itself will refuse the merge** until this branch contains `2a046ae0`. That is not a judgement call the merge session is making.

**Worth noting in your favour:** this PR carries its full row set. Two other PRs today read `mergeStateStatus: CLEAN` while being behind `main` *and* missing the required `CodeQL` context — a state where nothing visible blocks a merge that should have been blocked twice over. That is filed as #682. Yours is honestly reported, which is why the only thing standing in the way is the base.

## This rebase is NOT mechanical — measured, not assumed

`main` moved twice under you:

```
1b34ed50  The helper Chrome starts, and it answers inside the five seconds Chrome allows (#675)
2a046ae0  The archive travels in pieces, and the check asks Drive instead of downloading it (#677)
```

**Two of your sixteen files were changed by those merges:**

```
extension/app.html
extension/app.js
```

So this is not the disjoint case. `#677` in particular reworked the request path in `extension/app.js` and `extension/backend.js` — one `request()` now owns the request while `api()`/`bytes()`/`raw()` own the parse, and several bare `fetch(...)` calls gained a deadline and an abort signal. If your change touches how the corruption scan is requested, it may now be sitting on top of a different request contract than the one you wrote against.

**Re-derive rather than assume.** If you cite a line number in `extension/app.js` anywhere, recompute it against `2a046ae0`.

## What to do

1. Rebase onto `2a046ae0` and re-run.
2. Confirm the rebase moved the base and nothing else:

```bash
```

Identical before and after is a stronger statement than "no conflict". **A different hash means the rebase changed your content**, which is exactly what you want to know before reporting a green — especially given the two-file overlap above.

3. Before calling it ready, run the two checks that GitHub's own status does not answer:

```bash
gh pr checks 688 | awk -F'\t' '{print $1}' | sort -u
```

4. And confirm the `test` row was the engine tier rather than a documentation-tier pass wearing the same name — a row named `test` can be either:

```bash
gh run view <pull_request-run-id> --json jobs \
  --jq '.jobs[] | select(.name=="test") | .steps[] | "\(.conclusion) \(.name)"'
```

Prefer the `pull_request` run: a `push` run tiers against the branch's previous tip, a `pull_request` run against the merge base.

## Why the merge session will not rebase it for you

Branch protection offers an "Update branch" control, and using it would make this PR mergeable immediately without waiting. It has deliberately not been used.

Updating the tree from underneath you **voids the green that was measured on it, and with it any critical review that was run on it**. That would make the PR eligible in form while destroying the thing that makes it eligible in substance. The branch is yours; the merge is the merge session's.

**The moment this branch is genuinely on top of `main` — measured, not read from a status field — it will be merged.** It is being watched.

---

*Posted by the merge session. Nothing on this branch was touched. Eleven pull requests merged today; `main` is green at `2a046ae0`.*

muhammadbayoumi:
**Gate result, as promised in the body — and the first run was RED. Three failures, one of them a real defect in the page.**

Fixed in `a99880fa`, and the gate is clean on that head with identical fingerprints at both ends (`HEAD=a99880fa DIRTY=0 TREE=c3183a3d`):

| | |
|---|---|
| `ruff check scrapex/` | 0 |
| `eslint extension/` | 0 |
| `node --test extension/tests/*.test.mjs` | 0 — 428 passed |
| full suite, template mode | 0 |
| full suite, `SCRAPEX_FULL_MIGRATIONS=1` | 0 |

**What was red, and why the targeted runs did not show it.**

`test_pressing_it_asks_the_engine_and_dates_the_answer` failed only in the full-file run. `checkIntegrityFromPanel` refreshes the page after a successful check, because the stored verdict feeds `ready`. That refresh re-renders the Corruption row from `GET /api/storage` — which, until the engine's record is visible to the next read, carries no verdict at all. So the row wrote **"not checked" over the answer that had just arrived**: press Check, read `healthy`, watch it flip back.

Run alone, the test observed the text *before* the reload landed and passed. **A guard that depends on losing a race is not a guard**, so it now calls `loadDatabase()` itself and asserts the verdict survived. `heldIntegrity` is the fix: a re-read replaces the held verdict only with a **newer** one. Two more mutations, both caught — dropping the hold, and letting the read always win.

**And the contract moved, so the version moved with it** (`R-35`). `POST /api/storage/integrity` is a new endpoint; `contractstamp.differences` reported it against the recorded baseline. 0.4.9 → 0.4.10 in `scrapex/version.py` **and** `pyproject.toml`, with the four generated contracts and the CHANGELOG regenerated by `python -m scrapex.cli export-version` in the same commit.

Twelve mutations now, twelve caught.

muhammadbayoumi:
## Update — the overlap has grown from two files to five, and it will keep growing

`main` has moved again since the last comment. This branch is unchanged at `a99880fa` and is now **three commits behind**:

```
1b34ed50  The helper Chrome starts, and it answers inside the five seconds Chrome allows (#675)
2a046ae0  The archive travels in pieces, and the check asks Drive instead of downloading it (#677)
```

**Of your sixteen files, these five have been changed on `main` since your branch was cut:**

```
extension/app.html
extension/app.js
extension/startup.js
extension/tests/startup.test.mjs
tests/test_panel_dom.py
```

When this was first flagged the overlap was two files — `extension/app.html` and `extension/app.js`. `#608` added the other three: it reworked `extension/startup.js` and its test, and touched `tests/test_panel_dom.py`.

**Nothing is wrong with this PR.** It still reads 14 rows passing, zero failures, zero pending, with all nine required contexts present including `CodeQL`. Its work is done and its review is done. The only thing standing between it and `main` is the base — and the cost of that gap is now compounding rather than static.

## What has actually changed under you, so the rebase is informed rather than blind

- **`#677`** reworked the request path in `extension/app.js` and `extension/backend.js`. One `request()` now owns the request while `api()`/`bytes()`/`raw()` own the parse, and several bare `fetch(...)` calls gained a deadline and an abort signal they never had. **If the corruption scan is requested through any of those, the contract underneath it has changed.**
- **`#608`** rewrote parts of `extension/startup.js` and added `extension/tests/startup.test.mjs`. It also changed `tests/test_panel_dom.py`. **If your change alters what the panel draws on page open — which the title suggests it does — the startup path is now different code.**
- **`#675`** touched `scrapex/cli.py` and the native-host test; no overlap with your files.

`#688`'s subject — *the corruption scan is asked for, not drawn on every page open* — sits directly on the page-open path that `#608` just modified. That is the one to read first.

## Unchanged

```bash
```

A different hash before and after the rebase means the rebase changed your content. With five overlapping files that is now more likely than not, and it is worth knowing before a green is reported.

Then the two checks GitHub's status does not answer:

```bash
gh pr checks 688 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

And confirm the `test` row was the engine tier, not a documentation-tier pass wearing the same name.

**The merge session is still watching this and will merge it the moment it is genuinely on top of `main` — measured, not read.** It has not touched the branch and will not; using "Update branch" would void the green and the review that were measured on this tree.

Twelve pull requests merged today. `main` is green at `984e5359`. This is the only one left.

muhammadbayoumi:
**Rebased onto `984e5359`. Every check you named, measured rather than read — and the content hash you told me to watch DID change, so here is why.**

Head is now `d8fb4077` (`73f75124` + `d8fb4077` on top of `984e5359`).

```
git merge-base --is-ancestor origin/main HEAD   ->  on top of main
git diff origin/main...HEAD | git hash-object    ->  BEFORE f1e74ef8   AFTER 19a87fd7
```

## The hash changed. It is context, not content — and that is proven, not assumed

You predicted this and said it was worth knowing before reporting a green. It changed, so I did not stop at "no conflict". For each of the five overlapping files I stripped position and context from both diffs and compared **only the changed lines**:

```
git diff -U0 <old-base> a99880fa -- <f> | grep '^[+-]' | sort   vs
git diff -U0 origin/main HEAD      -- <f> | grep '^[+-]' | sort
```

| file | changed lines |
|---|---|
| `extension/app.js` | **identical** |
| `extension/startup.js` | **identical** |
| `extension/tests/startup.test.mjs` | **identical** |
| `tests/test_panel_dom.py` | **identical** |
| `extension/app.html` | **identical** |

So the rebase moved my change onto new neighbours without altering one line of it. The raw hash differs because hunk headers and surrounding context differ, which is what a hash over a whole diff measures.

## But identical lines are not identical MEANING, and your #608 note was the right one to read first

`#608` did not only move code near mine — **it changed the shape of the rule I added.** `LOCAL_POLICIES` entries can now be three-tuples with a method, and `deadlineForLocalRequest` gained `if (method && method !== normalizedMethod) continue;`. So a rule that matches the path but not the method now **falls through to later rules**, which it never did when I wrote mine. `/api/update` is now GET-only.

My entry is a two-tuple, so `method` is `undefined`, falsy, and the rule still applies to every method. Asked of the built module rather than reasoned about:

```
deadline for /api/storage/integrity  ->  120000   (integrityScan)
deadline for /api/storage            ->    5000   (destinationData)
```

Both correct, and the ordering survived: my rule still sits above the `storage|…` rule that would otherwise swallow it with the 5,000 ms this PR exists to escape. `extension/tests/startup.test.mjs` carries a case whose only job is to hold that position.

## Local gate on `d8fb4077`, fingerprints identical at both ends

```
BEFORE HEAD=d8fb4077 DIRTY=0 TREE=73e1dad0
  ruff check scrapex/                        0
  eslint extension/                          0
  node --test extension/tests/*.test.mjs     0
  full suite, template mode                  0
  full suite, SCRAPEX_FULL_MIGRATIONS=1      0
AFTER  HEAD=d8fb4077 DIRTY=0 TREE=73e1dad0
```

## The nine contexts, and the `test` tier

```
Analyze (actions)                 pass
Analyze (javascript-typescript)   pass
Analyze (python)                  pass
CodeQL                            pass
contract-parity                   pass
lint                              pass
migration-authority               pass
scope                             pass
test                              pass
```

And the `test` row is the **engine tier**, not a documentation-tier pass wearing the name. From the `pull_request` run (`34043486768`), that job's steps:

```
success  Tests (Python engine + Python-vs-frozen contract)
skipped  Documents only (this change touched nothing a document does not own)
skipped  Extension gate only (this change touched nothing else)
```

The engine step ran; the two narrower tiers skipped.

## One thing you should know before merging, because it is not in this PR

**#734 also claims 0.4.10.** It carries migration `0019` and bumped `scrapex/version.py` and `pyproject.toml` from 0.4.9 on a base that did not yet have this PR. Whichever of the two lands first, the other needs a rebase and a re-run of `python -m scrapex.cli export-version`. The contract check will say so rather than letting it pass silently — this is the class #694 already tracks.

---

*Reported by the session that owns this branch. It has not been merged and will not be by me.*

PR #734, opened 2026-09-06
A door for the missing profiles, and the card says when a press is owed

### LINKED ISSUE #689: The profile crawl has no door: no job kind, no route, no control — and 410 contractors have no profile
He named four things he cannot do from the panel. **Two of them he can**, and two have no
control, no route and no job kind — the same hole `approve` had until #669.

## Measured on the live warehouse, 2026-09-06

| dataset | rows |
|---|---|
| `contractors` | **17,768** |
| `contractor_profiles` | **17,393** |
| `contractor_enrichment` | 17,304 |

**410 contractors have no profile row.** The newest profile was written **2026-09-01**;
nothing has fetched a profile since. And the registered scope is `full_then_listing`, so
**nothing but the missing door stands in the way** — under `listing_only` the collector
would refuse, and it would be right to.

## What already works, and where

| he asked for | control | evidence |
|---|---|---|
| crawl the muqawil source | **Update now**, in the contractors card's menu on the Data page | it is what ran job 157: 6,713 pages |
| an update | the same control — `POST /api/jobs` infers `directory_crawl` from `muqawil_org` being in `directories.BUILDERS` | `app.py:3778` |

Worth noting for discoverability: there is **no card named muqawil**. `R-47` folded the
site's datasets into one card called *contractors*, so the control that crawls
muqawil.org is behind a card with a different name.

## What does not exist at all

`contractors.details` (`scrapex/contractors.py:630`) is reachable from `scrapex
contractors --details` **and from nowhere else**:

- not in `SPECIALISED_RUNNERS` (`scrapex/jobs.py`) — the four kinds are `crawl`,
  `organization_enrichment`, `directory_crawl`, `dataset_interpret`;
- not in `crawl_job`'s `job_kind` CHECK (`db/engine/schema.sql:1073`, plus `0018`);
- no route, and nothing in `sourceActions`.

So under `R-81` — the panel is the only surface — **the profile crawl does not exist for
the one person the tool is for**, and neither does completing the 410.

## What the door needs

The collector is already the right shape, which is why this is a door and not a feature:

- `details(conn, directory, fetch, fetcher, run_ref, ceiling=0, *, workers=1, connect=None, ids=())`
- the frontier is **built from stored listing pages, with no network** (`listing_pages`,
  `contractors.py:493`), so it survives a restart and knows which profiles are missing;
- `ids=` already **replaces** the frontier, which is exactly "complete these 410";
- `ceiling=` already exists so a first run can stop;
- `workers=`/`connect=` already exist, with the politeness proof: `HttpFetcher._throttle`
  holds its lock across the sleep, so concurrency buys overlap and never a higher
  request rate.

So the work is: a `profile_crawl` job kind, a migration widening the CHECK, a runner in
the shape of `datasetjob.run_dataset_interpret_job_once`, and one action in
`sourceActions`.

## Cost of the run itself

Measured previously at **9.03 s a page single-threaded** (the Dammam run). The 410
missing profiles are 820 pages across both locales, so **about 2 hours** as it stands.
`details` already takes `workers`, and the listing crawl measured 1.14 s a page at six —
the real figure for profiles needs measuring rather than asserting, and the politeness
floor (one request an interval) is what bounds it either way.

A full re-fetch of all 17,768 is a different question: ~35,500 pages, **about 87 hours**
single-threaded. `R-39`'s recorded 11.1 h is wrong — it was measured with six workers and
written against a single-threaded command.

## Options

1. **Build the door for both, with the missing set as the default frontier.** One job
   kind, one migration, one runner, one control — and the control's default is "fetch the
   profiles we do not have", with the whole frontier available too. Effort: the same
   shape as #668/#669, which took one session. Risk: low; the collector is unchanged.
2. **Door for the whole frontier only**, no missing-set default. Smaller diff; he then
   waits 87 hours or sets a ceiling to get the 410, which is the wrong instrument.
3. **Do nothing.** The profile half of muqawil stays terminal-only, which is the state he
   is objecting to.

Recommendation: **1.** The 410 is the live question and the collector already answers it
through `ids=`; a door that could only ask the 87-hour question would be a control he
cannot safely press.

**His call: whether to build it now, and whether the control's default is the missing set
or the whole frontier.**

muhammadbayoumi:
Closes #689.

His words: **«لا استطيع عبر الواجهة … عمل crawl ل profiles او حتى استكمال profiles الناقصة»** — and he was right. `contractors.details` was reachable from `scrapex contractors --details` and **from nowhere else**: no route, no job kind, no control. `R-81` says the panel is his only interface, so the profile half of muqawil did not exist for the one person the tool is for.

This is the same hole `datasetjob` closed for `approve`, one phase earlier in the pipeline. Two of the three verbs over that key had a door; this is the third.

## Measured on the live warehouse, joined on `contractor_id` out of `data_json`

| | |
|---|---|
| contractors sighted | 17,848 |
| profiles stored | 17,379 |
| **sighted ids with no profile** | **469** ← the frontier this door builds |
| the registered scope's whole frontier | ~35,700 pages, **~87 hours** |
| **the missing set** | **938 pages, ~2.4 hours** |

`missing_profile_ids` answers in **0.094 s** on his 2.08 GB warehouse.

**And the first count I reported was wrong, which is why a guard asserts the join.** I joined `generic_record.record_key` across the two datasets and said 410. Those keys are digests of the *parsed row* and the two datasets hash different things — `source_locator` is `div.section-card::row(N)` for a listing card and `div.info-box::row(1)` for a profile — so the join was meaningless and 410 was a coincidence of the right order of magnitude. `test_it_joins_on_the_contractor_id_and_not_on_the_record_key` stores a profile whose `record_key` matches nothing and whose `contractor_id` matches a sighted contractor, so the wrong join cannot come back quietly.

## The default is the missing set, by his ruling

A control whose only question takes 87 hours is one he cannot safely press. `details` already accepted named ids and they **replace** the frontier rather than filter it, so the narrow question is what the button asks by construction.

The route carries `ids`, `whole_frontier` and `ceiling` for a caller that wants another — **reachable so they are not dead branches**, and none of them drawn as a button. Asking for `ids` **and** `whole_frontier` is refused rather than resolved: named ids replace a frontier, so both together cannot mean anything.

**An empty frontier is a refusal, not an empty success.** `details` with `ids=()` falls through to the *scope's* frontier — the 87-hour sweep — so "nothing is missing" has to return before that. That is the exact four-round hole `--ids` records in `contractors.run`.

## `details` gained the boundary it had no way to offer

A 938-page sweep is two and a half hours the owner could not interrupt: no progress hook, no pause hook. `between_pages(index, total)` is the same hook `approve` took —

- asked **before** the fetch, so a stop costs no request;
- **refused above one worker**, because the pool calls back from worker threads where the job's connection cannot be used, and `ThreadPoolExecutor.__exit__` runs every queued task after a stop is decided. Accepting it there would be a pause control that does nothing on the path he would press it on.

**And a stopped sweep now closes `partial`.** The condition read only `len(todo) < len(frontier)` — the ceiling — so a sweep paused after nine of 938 pages closed as `SUCCESS`, and a later reader would take that as *"the site was fully read"*. That distinction is the whole worth of the State column.

## The job runs sequentially, deliberately

`details` takes `workers` and `connect`, and using them here needs `directoryjob.for_writing`'s connection-by-thread rule and `partitioncrawl`'s `stopping` event. That is a separate change with its own measurement; this one is a door.

The politeness reservation **is** taken — this job asks muqawil for pages, unlike the interpreter — and it wraps the fetching alone, so the frontier query and the closing report do not hold a host against another job.

## The card says what is waiting

The other half he asked for: **«اريد الظهور على الكارت انه يحتاج لعمل interpret store pages عند الحاجة لذلك حتى لا انتظر شى يحتاج اكشن منى»**. Measured the day the interpret door shipped: a crawl had finished with 6,713 stored pages, the card went on showing the 17,304 rows it showed before, and nothing said a press was owed.

- the line **names the action** in the menu row's own words, so what he reads and what he presses cannot be two different things;
- the profile line carries a **count**, because the engine knows the set exactly;
- the interpret line carries a **date and no count**, because counting uninterpreted pages means running the interpreter's frontier over every stored page — and #684 is what a guessed number costs: two counts of one run shared a word and read as 5,804 lost pages;
- it is a **comparison of finish times**: a crawl that completed after the last interpretation. Exactly the state needing a press, claiming no number.

**Measured cost:** `/api/sources` answered in 0.35 s before this and the three queries add about **0.09 s**, against the 5,000 ms that path is given. `/api/storage` is the cautionary tale beside it — a full-file scan on every page open, failing its deadline every time (#683).

**`badge off` is the amber badge and it is the only amber one there is.** I wrote `badge warn` first, which is the identical mistake `renderEngineDetail` already records making: an undefined variant renders as the plain grey badge, so a line whose whole job is to draw the eye would have drawn none. The DOM guard now **compares the rendered colour** against a probe `.badge` rather than trusting the class name.

## The drift check is live for the first time since the squash — and it passes

Two migrations have landed on the baseline, so `PREVIOUS_RELEASE` has a valid stop point again: the moment its own note predicted and told the next reader to re-choose at. It is a **slice index** into the stream, not a version, and **2** is the only value in range — an owner at v18, upgraded to v19 by this build, compared against a warehouse built fresh.

That is *not* the lowering the file's header warns about (`1` would compare a fresh build with itself), the note says so, and it says to **re-choose rather than increment** when 0020 lands. Passing means `0019`'s upgrade produces a schema identical to a fresh build.

## Tests

**Twelve mutations, twelve caught.**

- **15 engine-side** (`tests/test_a_door_for_the_missing_profiles.py`): the frontier; the `record_key` trap; the refusal when nothing is missing; the wrong job kind; a non-directory key; the kind registered *and* the CHECK accepting it; the page boundary; the hook refused above one worker; `partial` on a stop; and five route guards including that the default sends **no** frontier.
- **3 DOM guards**: both waiting lines with the amber colour measured, no line on a card with nothing waiting, and the button sending `profile_crawl` with no frontier named.
- **1 recipe** in `test_a_dataset_card_offers_what_works.py`, because that suite refuses an action whose route is not measured against a real engine — it went red on this change, correctly.

**Version:** 0.4.9 → 0.4.10, contracts regenerated by `python -m scrapex.cli export-version` (migration `0019` moved the contract). **#688 also claims 0.4.10** — if it lands first, this needs a rebase and a re-run of `export-version`, and the contract check will say so.

Local gate on `b499ccab`: `ruff check scrapex/` clean, `eslint extension/` clean, `node --test extension/tests/*.test.mjs` clean. The full Python suite in both template and `SCRAPEX_FULL_MIGRATIONS=1` modes is running as this opens; the result goes in a comment.


muhammadbayoumi:
**Green and complete — the only thing in the way is that `main` moved. One rebase and it merges.**

Measured against `origin/main` at `b934450a`:

```
mergeable                 MERGEABLE
mergeState                BEHIND
rows                      14 pass · 0 fail · 0 pending
required contexts         all nine present, CodeQL included
```

Nothing here needs fixing or reviewing again. `strict` is enabled on `main`, so GitHub itself refuses a merge from a branch that is behind — that is not a judgement the merge session is making.

**Two merges landed after your last run**, and both touch the panel:

```
b934450a  The corruption scan is asked for, not drawn on every page open (#688)
```

`#688` changed when the corruption scan is requested rather than drawn on page open, and `#608` reworked `extension/startup.js` and `tests/test_panel_dom.py`. Your subject — *a door for the missing profiles, and the card says when a press is owed* — is card-and-panel work, so check whether either of those moved what your card attaches to before assuming the rebase is mechanical.

After rebasing:

```bash
```

Identical before and after proves the rebase moved the base and nothing else — a stronger statement than "no conflict", and no merge-state field can express it.

Then the two checks GitHub's own status does not answer:

```bash
gh pr checks 734 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

A `CLEAN` verdict is not proof on its own: two PRs today read `CLEAN` while behind `main` *and* missing the required `CodeQL` row — filed as #682. Yours carries its full set, which is why the base is the only thing left.

**It is being watched and will be merged the moment it is genuinely on top of `main` — measured, not read.** The branch has not been touched and will not be: using "Update branch" would void the green and the review measured on this tree.

Thirteen pull requests merged today.

Issue #740, opened 2026-09-06
DECISION FOR THE OWNER (1 of 2): the Supabase product name is rendered as this product's own UI copy on...


muhammadbayoumi:
**Priority P1** · from the 2026-09-06 audit against the authoritative Supabase sources (area: The licence and trademark boundary, and the code-organisation conventions (Apache-2.0 §4/§6, supabase/AGENTS.md, apps/studio/AGENTS.md)).

## Here today
Two strings in the palette registry are user-visible chrome, not attribution: design/appearance.js:36 `label: "Supabase"` and design/appearance.js:37 `description: "Supabase console green, flat and bordered"`. They reach the screen at FOUR render points, all in design/appearance.js: the tile's <strong> (line 275), the tile's <small> (line 277), the tile's aria-label `${palette.label}: ${palette.description}` (line 261), and the appearance summary line `${scheme} · ${color}` where `color` is the label (lines 250-251). The browser is mounted on two surfaces: extension/app.html:870 `<div class="appearance-page-section" data-appearance-palettes>` and scrapex/webui/templates/settings.html:109. Since R-85 the registry holds ONE entry (design/appearance.js:33-41), so the panel draws a one-option radiogroup whose entire visible text is the Supabase mark. Full inventory for context: 175 occurrences of "supabase"/"Supabase" across the 20 live files that carry it (excluding docs/archive/, docs/reviews/, __pycache__, .pytest_cache and the four generated mirrors) = 33 IDENTIFIER + 2 VISIBLE LABEL + 5 test symbol names + 135 ATTRIBUTION (comments, docstrings, the notice, documentation). No Supabase logo, wordmark or brand asset is shipped: design/, extension/icons/ and scrapex/webui/static/ contain only Google's g, the Material sprite and this product's own x-mark.

## What Supabase expects
[...]

PR #764, opened 2026-09-07
The authority list the design system is audited against, kept where every machine reaches it

muhammadbayoumi:
«احتفظ بهذا الملف كمرجع لك ويمكنك تحديثه مع الوقت اذا اكتشفت او عدلنا شى فى desgin system» — 2026-09-06. He supplied a researched list of Supabase's authoritative sources and asked that it be kept and maintained rather than read once.

## It lives in the repository because a reference only one checkout can reach is not one

This work has already been interrupted once by a machine stopping, and the cost was a commit that existed nowhere else. A file in `Downloads` on one laptop is the same exposure.

## And it is a reference, not a register

That is what makes it admissible under CLAUDE.md's rule that work is never recorded in a markdown file. It records **where the truth lives**; it records no outstanding work, no plan and no state. Its companion `docs/DESIGN-SYSTEM.md` describes *this* product's system, and this file describes the system that one is measured against — so when they disagree, this is the outside authority and that is the thing under review.

## What it changed about how the audit was done, and it is more than a link list

The design-system audit that produced milestones 21–27 was run against it, fetching **68** of its URLs. Concretely it added:

- **Three layers, not one** — UI Patterns, Fragment Components, Atom Components. A product that reimplements a fragment which already exists upstream is a finding this repository had no way to detect before, and the audit found nine floating surfaces against one shared, and ten local empty states.
- **Seven official foundations documents, one of which is copywriting.** Voice, action labels, error messages, loading text and confirmations are part of the system being matched. Nothing here had ever been measured against them; that measurement is now milestone 27.
- **A source-conflict precedence order**, so "the docs say X and the code does Y" is recorded as a conflict with links and dates rather than silently resolved to whichever is convenient.
- **Three repositories named as ARCHIVED** — `supabase/ui`, `supabase/design-tokens`, `supabase-ui-web`. Any measurement taken from those is historical, not current.
- **A licence and trademark boundary.** Apache-2.0 permits the derivative work and does not grant use of Supabase's trade names. This product's single palette is registered under the id `supabase`, which is a question this file makes askable — filed as #740, and his to answer.

## Verification

Documentation-only: one added file, 228 lines, no code path touched. CI reports on it like any other change.


muhammadbayoumi:
**Two things stand in the way, and the second is the one nobody has noticed: this pull request has no CI rows at all. Not passing, not failing — none.**

Measured by the merge session at `2a1e3d10` against `origin/main` at `923fc798`:

```
mergeable      MERGEABLE
mergeState     BEHIND
behind         4 commits
ahead          1 commit
check rows     pass=0   fail=0   pending=0      <-- nothing has run
```

## 1 · No checks have run

Every other open pull request carries fourteen rows across nine contexts. This one carries zero. Branch protection on `main` requires six of those contexts — `lint`, `scope`, `test`, `migration-authority`, `contract-parity`, `CodeQL` — so **this cannot merge until CI has actually run and reported**, regardless of anything else.

A push to the branch should trigger it. If a push does not produce rows, something is wrong with the trigger rather than with the change, and that is worth reporting rather than working around.

**Do not read the absence as a pass.** That is the specific failure filed as #682: two pull requests yesterday were reported `CLEAN` by GitHub while a required context had no row at all, because a check that never runs cannot say anything against you. Zero rows is the extreme case of it.

## 2 · Four commits behind

```
b934450a  The corruption scan is asked for, not drawn on every page open (#688)
9fd7e8d6  The design document describes the product that shipped, not the one that was planned (#765)
923fc798  A door for the missing profiles, and the card says when a press is owed (#734)
```

`strict` is enabled, so GitHub itself refuses a merge from a branch behind `main`.

**The good news: your rebase is genuinely mechanical.** You change exactly one file — `docs/DESIGN-SYSTEM-SOURCES.md` — and **none of those four merges touched it.** Measured overlap: empty. So this is the one case today where "no conflict" and "nothing coupled" are both true.

One thing to check anyway: `#765` rewrote `docs/DESIGN-SYSTEM.md` and `docs/UI-KIT.md` so that they describe the product that shipped rather than the one that was planned. If your sources document cross-references either of those, confirm the reference still says what you meant after that rewrite.

## What this PR is, and why it matters more than its size

Your subject — *the authority list the design system is audited against, kept where every machine reaches it* — is the two-machine problem stated as a deliverable. The owner works from two machines and two accounts, and anything recorded outside the repository exists on exactly one of them. A single markdown file is a small diff and a large fix.

It is also **documentation-only**: one file under `docs/`, nothing else. Under the merge bar that exempts it from the critical review requirement — but note that "documentation-only" is decided by what the diff touches, not by the branch name, and the `scope` job decides it independently of anyone's reading.

## What to do

1. Push so CI runs, and confirm rows actually appear:

```bash
gh pr checks 764 | awk -F'\t' '{print $1}' | sort -u
```

2. Rebase onto `923fc798`.
3. Confirm the rebase moved the base and nothing else:

```bash
```

With zero overlap this should be identical before and after. If it is not, something unexpected happened and it is worth reading.

4. Measure the base rather than reading it, because `mergeStateStatus` lags and has been wrong today:

```bash
```

**It will be merged the moment it has real rows and is genuinely on top of `main`.** The branch has not been touched and will not be.

Fifteen pull requests merged; `main` is green at `923fc798`.

muhammadbayoumi:
## Rebased and clean — and the `CodeQL` NEUTRAL I flagged was transient

`main` had moved **four** times, so this read `BEHIND` and branch protection's `strict` would have refused it. Rebased onto `origin/main` without waiting to be asked, since this is the same condition the merge session flagged on PR 640 yesterday.

**The diff is byte-identical before and after the rebase** — `git diff origin/main...HEAD | git hash-object --stdin` returns `a4e1b341` both times. The base moved and the content did not.

| | |
|---|---|
| head | `84a37f56`, 0 behind |
| CI | **14/14 SUCCESS**, `MERGEABLE`, `CLEAN` |
| local | `pytest -m docs` green |
| diff | one added file, 228 lines, no code path touched |

**Correction to something I said mid-flight:** I flagged `CodeQL` returning `NEUTRAL` as a possible gate problem on a documentation-only change. It settled to `SUCCESS` once its three sub-analyses finished — it was reporting an in-progress rollup, not a verdict. Nothing to fix and no issue to file.

Issue #763, opened 2026-09-06
Two panel update states nothing can reach, and one percentage that can freeze at 0% for the whole download


muhammadbayoumi:
From the code-quality dimension of the critical review of #608. **Optional** — recorded rather than fixed, per "optional becomes an issue, never a fix in the same PR".

## Three small things in `extension/app.js`, measured

**1. A dead branch in `engineUpdateSentence`.** `pct` is `null` only when `progress` is absent, and the engines state object never produces that — so the bare "The engine is downloading the update." arm cannot be reached against any engine that has the route.

**2. A percentage that can freeze at 0%.** A manifest whose installer omits `bytes` gives `percent === 0` for the entire download, so the sentence reads "— 0%" for fifteen minutes. `release.py` coerces a missing `bytes` to 0 and `verifiable` does not require it. The panel already has the right idiom ~400 lines below, in `startInstallerDownload`: `total > 0 ? … : "Downloading…"`. The fix is to test `report.progress.total > 0` rather than `typeof … === "number"`.

`.github/workflows/release-engine.yml` always writes `bytes`, so reaching this needs a hand-edited manifest — which is why it is optional rather than a defect.

**3. An unreachable ternary.** In the button branch, `latest.state === "ok" ? latest.version : "the update"` cannot take its second arm: `can_self_update` implies the feed was ok, which implies the engine s `latest` won the row, which implies `state === "ok"`.

## Options

- **do nothing** — one wrong percentage in a case a published release cannot produce, and two arms nobody reaches.
- **fix 2 only** — one condition. It is the only one the owner could ever see.
- **fix all three** — 2 is a condition, 1 and 3 are deletions of unreachable code, which needs care: an arm that is unreachable today is a guard tomorrow if the engine changes.

Recommendation: fix 2, leave 1 and 3. Deleting a defensive arm to remove a line is the wrong direction in a file whose whole subject is states the panel gets wrong.

Issue #762, opened 2026-09-06
Restart engine silently discards a download in flight and orphans the .part file


muhammadbayoumi:
From the architecture dimension of the critical review of #608. **Optional**, recorded rather than fixed.

## The sequence

`#runtime-restart` sits on the engine detail screen, a few rows above `#engine-download`. `scrapex/webui/update_api.py` keeps its state and its `running` flag in a closure in engine-process memory, and the worker is a daemon thread.

Press **Download and check 1.0.2**, then **Restart engine**:

1. The thread dies with the process.
2. The `.part` in `~/.scrapex/updates` is orphaned — the `except`-unlink in `scrapex/update.py` cannot run.
3. The successor process starts with a fresh `idle` state.
4. The panel s next report re-arms the button as if nothing had happened.

`scrapex/update.py`s `clear_staging` has **no production caller**, so nothing removes the orphan. Disk cost is one partial file of up to ~70 MB per occurrence, and it accumulates.

Nothing false is stated on screen and the waste is bounded, so this is optional rather than a defect to fix inside a PR being reviewed down to nothing.

## Options

- **do nothing** — an orphaned partial file per interrupted download, invisible.
- **call `clear_staging` on engine start** — it exists and has no caller. Small, and it makes the successor tidy up after its predecessor.
- **warn before restarting while a download runs** — needs the panel to know, which it does from `progress_state.phase`; more UI for a rare case.

Recommendation: the middle one, on its own. It gives an existing function the caller it was written for.

Issue #761, opened 2026-09-06
GET /api/update refetches the release manifest on every poll turn: 39-311 third-party round trips per download


muhammadbayoumi:
From the performance dimension of the critical review of #608. **Engine-side, and deliberately not fixed in that PR.**

## Counted

`scrapex/webui/update_api.py` calls `release_mod.latest()` on **every** GET, unconditionally. `scrapex/release.py` uses the module-level `httpx.get` — a fresh client, so a new TCP + TLS handshake per call — and sends `Cache-Control: no-cache`, which forces an origin revalidation rather than accepting the CDN copy that the per-minute cache key was designed to share.

The panel polls that route every 3 s while a download runs. The downloading branch reads exactly **one field** of the response: `progress.percent`, a number the engine already holds in memory.

| link | turns per download |
|---|---|
| 25 Mbps | 9 |
| 5 Mbps | 39 |
| 1 Mbps | 188 |
| the poll ceiling | 311 |

So 39–311 handshakes and forced revalidations of `version.json` per download, at 20/min, against a URL whose own design note says everyone asking within the same minute should share one cache entry. Before the panel had a caller the route was hit 1–2 times per Engines visit.

Against 70 MB the added bytes are noise and the download is not measurably slowed — which is why this is not urgent. It matters because it multiplies by the number of installations on a shared IP, and because it gives the transient-failure path its 311 chances to fire.

## Options

- **do nothing** — costs nothing measurable to one user.
- **memoise `latest()` on the minute bucket it already computes** — ~8 lines and a test. A release cut mid-minute becomes visible under 60 s later, which `manifest_url`s docstring already accepts. Drops 310 to at most 16 per 930 s.
- **a progress-only route reading the in-memory state** — zero third-party fetches, but a new API surface with one caller, which is the shape "no helper with one caller" warns against.

Recommendation: the middle one. "Respect the politeness budget" names the cost; the fix is engine-side and belongs in its own change.

Issue #690, opened 2026-09-06
The release feed publishes minimum_extension and published_at, and nothing reads either


muhammadbayoumi:
Found while implementing his ruling that the engine wins the release row (#608).

## The measurement

`extension/releases.js` builds the panel's release object with two fields nothing consumes:

- `minimumExtension` — from `body.minimum_extension_version`, with a comment
  saying it is published beside the release **"so the panel can say `this needs a
  newer extension` BEFORE anything is installed rather than after."**
- `publishedAt` — from `body.published_at`.

`grep -rn "minimumExtension\|publishedAt" extension/` outside `releases.js`
returns exactly one hit, and it is `releases.test.mjs` asserting that the reader
produced them. **No screen reads either.** The engine sends the same two fields
in its own `latest` block (`minimum_extension`, `published_at`) and they are
equally unread.

So the comment describes a capability the panel does not have. That is the shape
`R-81` is about, one level down from a missing button: a fact fetched, parsed,
tested and then dropped.

## Why it matters for `minimum_extension` specifically

It is the only thing that could stop him installing an engine his extension
cannot talk to. Today the panel finds that out after the install, from a
protocol mismatch — which is exactly the "after rather than before" the comment
says this field exists to prevent.

`publishedAt` is smaller: a date beside a version is useful and its absence is
not a defect.

## The options

- **do nothing** — the fields stay parsed and unused. The comment in
  `releases.js` stays false about what the panel does with it.
- **read `minimum_extension` and refuse before the download** — compare against
  the extension's own version and say so on the row rather than after the
  install. Needs a decision about what "refuse" means: a disabled button with a
  reason, or a warning that still lets him proceed.
- **delete both fields and the comment** — honest, smaller, and throws away the
  one fact that could prevent a bad install.

Recommendation: the middle one for `minimum_extension`, and show `published_at`
beside the version while there — both are already in hand on both paths. The
"disabled versus warned" question is his.

Related: #681 (the engine collapses three refusal causes into one sentence) —
this would be a fourth cause worth naming separately if the refusal is built.

PR #766, opened 2026-09-07
The updater answers honestly: the review findings 608 merged without

muhammadbayoumi:
## What this is

**The findings from #608's critical review that #608 merged without.** That PR merged at 15:20 UTC on the squash of `499503b6`; the review had returned fourteen findings and the repairs were pushed an hour later. So they land here instead.

Nothing in this PR changes what the feature does. It corrects what it *says*, and it makes the half nothing was asserting fail when it is broken.

## The must-fix, reached independently by three of the four reviewers

`pollEngineUpdateLoop` rendered a **single** `null` as the definite sentence *"The engine stopped answering while the update was running"* and **returned**, ending the watch — about an engine that was very likely still downloading.

`engineUpdateState` flattens every failure to `null`, the deadline included. And the deadline is a **tie, not a margin** — measured, not read off the comment that claimed one:

```
httpx.Timeout(4.0) -> connect 4.0  read 4.0  write 4.0  pool 4.0
```

`CHECK_TIMEOUT_S = 4.0` is applied **per phase**, so the engine's own worst legitimate answer is connect 4 s + read 4 s = **exactly the 8000 ms** the panel allows, before a redirect hop, the loopback round trip or uvicorn's dispatch. The turn most exposed to it is a poll turn, because those phases run while the same process streams ~70 MB.

The attempt budget owns the giving up now, which is what the restart poll in this same file has always done with a failed probe. This repository has a test named for the distinction — `test_a_health_check_that_times_out_is_not_reported_as_no_engine`: *"a health check that runs out of time is a DIFFERENT answer from an engine that is not there, and the difference is the whole of what to do next."* This loop was making exactly that conflation.

**And the button is not touched in either terminal branch.** Re-enabling it there produced a control that was enabled, labelled `Downloading…`, and carried `onclick = null`: **pressable and inert**. The previous pass's repair swapped a button that could not be pressed for one that can be pressed and does nothing. The way back is `Check again` — a different control, enabled, and now named in the sentence, which also replaces a pointer at *"its own log"*, a surface the panel does not have.

## Six more should-fix

| | |
|---|---|
| **`installerDownloadRunning`** | A browser-side install is state only the panel holds, and nothing could consult it, so any re-render re-armed the button mid-download — reachable with no timing luck through the reopen path, since a first install has no engine and the repaint is microtasks. Pressing again started a second ~70 MB fetch of the same file, landing as `scrapex-engine (1).exe`, with two intervals writing one label. **Cleared on both exits:** a flag released on one is a button that never comes back. |
| **`staged` means *this* release** | `phase` alone cannot tell one release from another. An engine holding 1.0.0 staged stays in that phase until it restarts, so publishing 1.1.0 left the row reading "1.1.0 / Update available" above a disabled button labelled for a **different** version — with no panel path to 1.1.0 and nothing saying so. The engine would have accepted that POST. |
| **The engine's `verifiable` verdict** | False when the host is not in `ALLOWED_INSTALLER_HOSTS` — and the panel was handing that same URL to `chrome.downloads` itself, taking the risk the engine declined, under a sentence blaming a missing SHA-256 that is not missing. **`=== false`, never `!verifiable`:** the panel's own release object has no such field, so a falsy test would delete the button for every first install — the one case with no engine to ask. |
| **The checksum note** | Claimed the engine checks its own updates, and was displayed *only* in the states where that is false: a source checkout, and a release the engine has just refused to verify. |
| **`READERS`** | Gains `updateEngineReleaseUI`, the third reader, added without its marker in the same diff that wrote the rule down. |
| **The poll's derivation** | Corrected a second time — see below. |

## Two claims in the previous prose that do not survive the code

**`DOWNLOAD_TIMEOUT_S` is applied per phase.** For a stream the governing phase is `read` — the longest gap between chunks, not the elapsed time. 70 MB arriving steadily at 20 KB/s takes ~3500 s and never trips it, which is **above this poll's own ceiling**. So the engine is *not* guaranteed to give up first, and the `R-48` arithmetic was comparing an elapsed bound against an idle one. What makes it safe is the sentence — running out says the panel stopped watching and never announces a failure — not the arithmetic.

`download.onclick` is written as a statement rather than a ternary because `test_window_open_survives_only_as_a_fallback` pins that exact line. The ternary preserved the behaviour and broke the guard, which is the wrong trade: the behaviour was never the part at risk.

## The tests dimension: 23 mutations run, **14 survived**

That is the part of this PR that matters most, and it is why the code half above was not enough.

- **The watcher half had no behavioural coverage at all.** Gutting the loop, removing the reopen restart, and removing the watcher after a successful POST each left all seventeen update tests green. The one test reaching the loop asserted an **absence**, so deleting the watcher satisfied it for the opposite reason to the one it named.
- **A correction worth recording:** the new progression test opens the screen with a report already `downloading`, so the *renderer* starts the watcher and the button is never pressed — deleting `await pollEngineUpdate()` **still survived it**. `test_pressing_the_button_starts_the_watching` keeps the engine idle while the screen renders, starting no watcher at all, and asserts the absence of a percentage before pressing, so it cannot attribute what follows to anything else.
- **Every error path in the write half was deletable.** `_STUB` answered every POST with 200 and had **no failure mode at all**, so `isTimeoutError` was never exercised on this path. It now carries `__post_status`, `__post_throws` and `__post_detail`, and one test separates the three outcomes.
- **`phase: "failed"` had no test at all** — the state the owner sees on a digest mismatch, the branch carrying this feature's whole security value.
- **The staged branch's control** was unguarded while its sentence was well covered: making the button live and relabelling it both survived.
- **The contract guard was blind three ways**, one of them its own repeat offence: the length floor was on the **sum** of the slices — about 9000 characters against a threshold of 800 — so any one slice could collapse to nothing, and the wrapper-split move defeated it exactly as it defeated the previous version. And `test_the_panel_reads_the_refusal_reason_at_all` grepped the whole file for a literal that also appears in **two comments**, so renaming the one line that reads it kept the test green.
- **Two fixtures made their own subject unmeasurable.** `_HEALTH["version"]` and `_report()["installed"]` both carried `"0.4.6"`, so the half of his ruling about `installed` was invisible. And three tests passed a whole `latest={...}` block, bypassing the fixture guard added in the same diff — stripping six of nine keys stayed green. `latest_over=` merges instead, and refuses both kwargs at once.

**Measured: 10 of 10 previously-surviving mutations are now caught.**

## Not in this PR

**#761** manifest refetched every poll turn (39–311 third-party round trips per download, for one integer) · **#762** restart discards a download in flight and orphans the `.part` · **#763** a dead branch, an unreachable ternary, and a percentage that can freeze at 0% · **#681** the engine's collapsed refusal sentence, now extended with the measurement that `(update_available=True, can_self_update=False, blocked="")` is reachable in **no case**, so the panel's no-installer fallback is dead against the real engine · **#686**, **#687**, **#690**.

## Gates

`pytest -m "extension or docs"` exit 0 · `node --test extension/tests/*.test.mjs` 451 pass · contract parity pass.

**This has not had its own critical review yet.** It is a code change, so #595's bar applies: four dimensions, an adversary over what they find, then a re-review.


muhammadbayoumi:
**You were ready and were passed over deliberately — `#734` merged first because its rebase is far more expensive than yours. One rebase and this merges.**

```
origin/main   9fd7e8d6 -> 923fc798
              A door for the missing profiles, and the card says when a press is owed (#734)
mergedAt      2026-09-07T06:47:49Z
VERSION       0.4.10 -> 0.4.11
```

Both of you read `CLEAN`, on top of `main`, 14 rows passing, all nine required contexts present. Only one could go first, because under `strict` the second is stale the moment the first lands. The choice was made on rebase cost, not merit:

| | files | carries |
|---|---|---|
| `#734` | 17 | a new migration `0019`, three `contracts/*.json` baselines, a `VERSION` bump, `pyproject.toml` |
| this PR | 4 | `extension/app.html`, `extension/app.js`, and two tests |

Rebasing a migration and its contract baselines is expensive and easy to get wrong; rebasing four files is not. **You were the cheaper one to ask, and that is the only reason.**

## Your actual overlap with what just landed

Two of your four files were touched by `#734`:

```
extension/app.js
tests/test_panel_dom.py
```

`#734` added a door for missing profiles and a card that says when a press is owed — panel work, in the same file you change. So this is not the disjoint case. Read what it did to `extension/app.js` before assuming the rebase is mechanical, and if you cite a line number in either file, recompute it against `923fc798`.

## After rebasing

```bash
```

Identical before and after proves the rebase moved the base and nothing else. **A different hash means it changed your content** — likely here, given the two-file overlap — and that is worth reading before a green is reported.

Then the two things GitHub's status does not answer:

```bash
gh pr checks 766 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

A `CLEAN` verdict is not proof by itself: two PRs yesterday read `CLEAN` while behind `main` *and* missing the required `CodeQL` row, which is filed as #682. Yours carried its full set, so the base is genuinely the only thing left.

## One note on what this PR is

Its subject — *the updater answers honestly: the review findings 608 merged without* — makes it a follow-up to work already on `main`. That is worth keeping visible in the PR body rather than only in the title: a reader six months from now wants to know that `#608` shipped with known findings and that this is where they were answered, not to discover it from a commit subject.

**It is being watched and will be merged the moment it is genuinely on top of `main` — measured, not read.** The branch has not been touched and will not be.

Fifteen pull requests merged; `main` is green at `923fc798`.

muhammadbayoumi:
## Update — `main` moved once more, so you are one behind again

Nothing has changed about the verdict: **green, complete, and only the base is in the way.** 14 pass, 0 fail, 0 pending, all nine required contexts present.

You are now behind by one:

```
923fc798  A door for the missing profiles, and the card says when a press is owed (#734)
```

That is the merge referred to in the previous comment — the reason you were passed over. It is done, so your turn is next in the queue.

**Your overlap with it, measured:**

```
extension/app.js
tests/test_panel_dom.py
```

Two of your four files. `#734` added a door for missing profiles and a card that says when a press is owed, in the same `extension/app.js` you change. Read what it did there before treating the rebase as mechanical, and recompute any line number you cite in either file against `923fc798`.

Unchanged from before:

```bash
git diff origin/main...HEAD | git hash-object --stdin      # different hash = your content changed
gh pr checks 766 | awk -F'\t' '{print $1}' | sort -u       # expect nine names
```

**It will be merged the moment it is genuinely on top of `main` — measured, not read from a status field.** The branch has not been touched.

muhammadbayoumi:
## You are ready and the queue is holding you deliberately — do not keep rebasing

This is not a problem with your PR. Measured: **14 rows passing, zero failures, zero pending, all nine required contexts present.** The work and its review are done.

**The owner has given `#735` merge priority**, because that branch has redone the required work more than once — a conflict resolved, four missing required rows obtained, three rebases against a moving `main`, and then a real failure that turned out to belong to `#608` rather than to itself. Four rounds for one change, none of them its own defect. The queue is being held for it rather than moving past it a fifth time.

**So you will not be merged ahead of it, even when you are the only green one.**

## The practical part: stop rebasing until it lands

You are currently **4 commits behind**. Rebasing now buys nothing:

- It will not make you eligible sooner, because the hold is on order, not on your state.
- Each rebase costs a full CI cycle — and a pull request here runs both a `push` and a `pull_request` workflow, which is measured at roughly 38% of this repository's runner-minutes on its own.
- `#735` will move `main` again when it lands, so any rebase you do now is one you pay for twice.

**Rebase once, after `#735` merges.** You will be told when that happens.

If you want something to do in the meantime, the useful thing is not a rebase: your subject — *the updater answers honestly: the review findings 608 merged without* — makes this a follow-up to work already on `main`. Make sure the PR body says that plainly, so a reader in six months learns that `#608` shipped with known findings and that this is where they were answered, rather than having to reconstruct it from commit subjects.

## For information: what will be under you when your turn comes

Since your last run, `main` has taken:

```
687972c2  Three things a stopped or duplicated job cost him: the pages, the progress bar, and one shared word (#767)
a960711d  The Apache 4(b) statement describes the work it ships, and a guard keeps it that way (#768)
```

`#767` touches `extension/app.js` and `tests/test_panel_dom.py` — two of your four files. `#768` touches the three synced `tokens.css` and `supabase.NOTICE.txt` copies plus `docs/DESIGN-SYSTEM.md`, none of which are yours.

So when you do rebase, `#767` is the one to read, plus whatever `#735` brings.

**Your slot is held, not lost.** Nothing on your branch has been touched.

muhammadbayoumi:
## The hold is released — `#735` merged. Rebase now, once.

```
origin/main   a960711d -> 102e48ff
              Back up from the page that lists the backups, and watch it happen there (#735)
mergedAt      2026-09-07T10:47:45Z
```

You were asked to stop rebasing and wait rather than pay for cycles that would be spent twice. That is over: **you are next, and this is the rebase to do.**

You are **5 commits behind**, which is the cost of the wait — and it was the cheaper cost. Five rebases paid one at a time would have been five full CI cycles for the same result.

## Your overlap, measured against everything you must now rebase onto

Three of your four files:

```
extension/app.html
extension/app.js
tests/test_panel_dom.py
```

The two that matter most for reading, since both landed in that window and both touch what you touch:

- **`#767`** — *three things a stopped or duplicated job cost him: the pages, the progress bar, and one shared word.* It changed `extension/app.js`, `tests/test_panel_dom.py` and `tools/panel_harness.py`. This is the one that fixed the real incident where a cancelled crawl orphaned 3,138 stored pages, so its changes to the panel's job surface are deliberate and should be preserved rather than resolved away.
- **`#735`** — *back up from the page that lists the backups.* It changed `extension/app.css`, `extension/app.html`, `extension/app.js`, `tests/test_panel_dom.py` and `tools/panel_harness.py`. It also had to fix a failure in `test_a_named_refusal_from_the_engine_reaches_the_screen` — the test `#608` added for the update button — because its own `app.js` changes had shadowed that path.

**That last point is the one to read before you start.** Your PR is the follow-up to `#608`'s review findings, so you are working on the update path directly; `#735` just had to repair that exact path after touching the same file. Whatever it did to keep the update button reachable is now on `main`, and your rebase must not undo it.

## After rebasing

```bash
git log --oneline origin/main..HEAD                     # your own commits only
git diff origin/main...HEAD | git hash-object --stdin   # different hash = your content changed
gh pr checks 766 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

With three overlapping files across five commits, expect the diff hash to change and read what changed before reporting a green.

And confirm the `test` row was the engine tier rather than a documentation-tier pass wearing the same name — check the step conclusions of the `pull_request` run.

**Nothing on your branch was touched during the hold.** Nineteen pull requests merged; `main` is green at `102e48ff`.

Issue #670, opened 2026-09-06
The sign-in busy test waits on the button and asserts on the status: a 300 ms window, and a one-line fix


muhammadbayoumi:
`tests/test_signing_in_says_what_happened.py::test_the_google_button_is_busy_while_signing_in` fails intermittently under full-suite load on Windows:

    assert "Signing in" in page.text_content("#signin-status")
    AssertionError: assert 'Signing in' in ''
    tests/test_signing_in_says_what_happened.py:398

It passes **4/4 in isolation** and the whole file passes at file scope. It only appears in a full-suite run.

## The mechanism, and it is not "flaky"

The test waits on **one element** and then asserts on **another**:

```python
page.click("#signin")
page.wait_for_function("""() => {
      const btn = document.getElementById('signin');
      return btn.disabled && btn.getAttribute('aria-busy') === 'true';
    }""", timeout=2000)

assert "Signing in" in page.text_content("#signin-status")
```

The handler at `extension/app.js:6041-6056` sets both, but they do not have the same lifetime:

```js
btn.disabled = true;
btn.setAttribute("aria-busy", "true");   // the wait_for_function is satisfied here
status.textContent = "Signing in…";
try   { await loadAccount({ interactive: true }); }
finally {
  btn.disabled = false;
  status.textContent = "";               // and cleared here
}
```

**The gap inside the synchronous block is not the race** — JavaScript is single-threaded and Playwright cannot poll between two statements of one handler. The reachable window is the other end: the test's stub calls back after **300 ms** (`setTimeout(() => cb("stub-token"), 300)`), and the assertion needs **two separate round-trips** — `wait_for_function` returning, then `text_content` — to land inside that window. Under a loaded full-suite run those round-trips outlast 300 ms, `finally` has already run, and the assertion reads the cleared value.

The observed failure says exactly this: `in ''` is **cleared**, not *never written*.

## Why it belongs in this milestone

It is the same shape as everything else here. **The wait guards one thing and the assertion reads another**, and the two are assumed to be the same state. `aria-busy` on the button is not the status paragraph's text, and they expire at different times.

## The fix is one line

Fold the text into the same in-page predicate, so the whole condition is evaluated atomically in the browser instead of across two round-trips:

```python
page.wait_for_function("""() => {
      const btn = document.getElementById('signin');
      const st  = document.getElementById('signin-status');
      return btn.disabled
          && btn.getAttribute('aria-busy') === 'true'
[...]

Issue #420, opened 2026-09-04
OP-132 · The button's crawl runs one worker where the terminal's runs six


muhammadbayoumi:
**Measured on his live crawl, 2026-09-03, while he asked when it would finish.** Seven of
fifty-six cells closed in 198 minutes:

```
closed 7/56       3,216 contractors · 1,299 listing requests · 198 min
per contractor    0.404 requests
projected total   7,035 requests for 17,417 contractors   ->  5,736 remaining
measured rate     6.6 requests/min          SINGLE WORKER
                  14.6 h left
at six workers    2.4 h left
```

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L4459](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L4459). `OP-132` is kept in the title because the code cites it.

Issue #367, opened 2026-09-04
OP-31 · Six crawl workers starve another process out of the warehouse


muhammadbayoumi:
**Measured 2026-08-21, minutes after `--workers` was built, by using it.** The owner
asked for the Engine to be started while a six-worker crawl was running. It refused:

```
sqlite3.OperationalError: database is locked
  scrapex/jobs.py:975  record_worker_failure
```

Every connection sets `busy_timeout = 5000`, so a writer waits five seconds before
giving up. **Six concurrent writers kept it waiting longer than that** — so the
failure is not a missing timeout, it is a timeout that is too short for the load the
new flag can create.

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L1331](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L1331). `OP-31` is kept in the title because the code cites it.

PR #776, opened 2026-09-07
The notice's last declared departure was never one: it is Supabase's value, resolved

### LINKED ISSUE #775: The notice's last declared departure is on the trunk and is not a departure: --accent-contrast is their resolved value
**Live on `main` at `a960711d`** (PR 768, merged 2026-09-07 09:56Z). That PR corrected four of the five entries in the Apache-2.0 §4(b) statement of changes and left the fifth standing as *"ONE VALUE IS STILL DELIBERATELY REPLACED"*. **The fifth was never a replacement either**, and the guard the same PR added pins the false claim.

Found by the second four-dimension pass of the merge gate, and the adversary strengthened it rather than refuting it.

## The measurement, computed independently rather than taken from the review

Supabase's `--primary-foreground` at the commit the notice pins (`86c813ec`), verbatim from `packages/ui/build/css/source/semantic.css`:

```css
--primary-foreground: oklch(
  min(var(--surface), var(--foreground-lightness)) calc(var(--chroma) * 0.45) var(--primary-hue)
);
```

Evaluated with the scalars their own theme files declare:

| theme | their scalars | resolves to | ships here as |
|---|---|---|---|
| light | `--surface: 0.995`, `--foreground-lightness: 0.1`, `--chroma: 0`, `--hue: 159` | `oklch(0.1 0 159)` = **`#030303`** | `--accent-contrast: #030303` |
| dark | `--surface: 0.19`, `--foreground-lightness: 0.95`, `--chroma: 0.005` | `oklch(0.19 0.00225 159)` = **`#131413`** | `--accent-contrast: #131413` |

**Both reproduce byte for byte.** Two independent scalar sets landing on two shipped values is what validates the arithmetic rather than the conclusion.

So Supabase reaches the same near-black by the same reasoning. This product agreed with them, and the notice records the agreement as a departure. The contrast fact the entry cites — their brand green is 1.99:1 on white — is still true and was never the cause.

## Why it matters more than a wrong sentence

- **It is the direction a licence notice can least afford**, and the same notice self-corrects for exactly this in item 4: *"wrong in the direction a licence notice can least afford, crediting this product with their work."* Item 2 does the mirror image — it credits **this product** with a choice **they** made.
- **Item 2 should now be empty.** All five entries are gone: three restored by `R-85`, one describing a deleted device path, and this one that never was. Every colour value belongs to item 1, *resolved, not copied*.
- **The guard locks it in.** `tests/test_the_notice_describes_the_values_it_ships.py` on `main` pins `OURS = {"--accent-contrast": "#030303"}`, so the false claim is now asserted by a test.

## A second, smaller error in the same document

The sentence PR 768 wrote to replace a stale count says `R-85` moved three values *"into the byte-exact one"*. It moved them into the **re-derived** set: upstream `--border-stronger` and `--ring` are alpha composites and `--warning` is an `oklch()` call, so what ships is this product's evaluation of their expressions. Saying byte-exact over-credits them with three values that are our own output — the mirror of the motion error, and the third time in this work that provenance was stated in the wrong direction.

## And two gaps in the guard, which the same pass named

- It reads **only the first `:root` block**, so a dark value could move with nothing noticing.
- `test_both_easing_curves...` asserts `curve in light` against the **raw file**, and the comment above the tokens contains both curve literals — so **deleting `--ease` entirely would leave the test green**. A guard that passes because of prose is not a guard.

## Acceptance

Item 2 declares no deliberate colour replacement and is kept as an empty statement rather than deleted, so a reader holding an earlier distribution knows the list emptied rather than moved. The comment at `design/tokens.css` beside `--accent-contrast` says the value is theirs resolved. The byte-exact sentence says re-derived. The guard pins an **empty** replaced set, reads both theme blocks, and asserts the curves against the **declarations**. All three synced copies stay byte-identical. Broken on purpose on four axes and red each time.

muhammadbayoumi:
Closes #775.

**PR 768 corrected four of the five entries in the Apache-2.0 §4(b) statement of changes and left the fifth standing.** The fifth was never a replacement either, and the guard that PR added pins the false claim. Found by the second four-dimension pass of the merge gate; the adversary strengthened the finding rather than refuting it.

## The measurement, computed here rather than taken from the review

Supabase's `--primary-foreground` at the commit the notice pins (`86c813ec`), verbatim from `packages/ui/build/css/source/semantic.css`:

```css
--primary-foreground: oklch(
  min(var(--surface), var(--foreground-lightness)) calc(var(--chroma) * 0.45) var(--primary-hue)
);
```

Evaluated with the scalars their own theme files declare:

| theme | their scalars | resolves to | ships here as |
|---|---|---|---|
| light | `--surface: 0.995`, `--foreground-lightness: 0.1`, `--chroma: 0`, `--hue: 159` | `oklch(0.1 0 159)` = **`#030303`** | `--accent-contrast: #030303` |
| dark | `--surface: 0.19`, `--foreground-lightness: 0.95`, `--chroma: 0.005` | `oklch(0.19 0.00225 159)` = **`#131413`** | `--accent-contrast: #131413` |

**Both reproduce byte for byte.** Two independent scalar sets landing on two shipped values is what validates the arithmetic rather than the conclusion.

Supabase reaches the same near-black by the same reasoning. This product agreed with them, and the notice recorded the agreement as a departure. The contrast fact the entry cited — their brand green is 1.99:1 on white — is still true and was never the cause.

**It is the direction a licence notice can least afford**, and the same notice already self-corrects for it in item 4: *"wrong in the direction a licence notice can least afford, crediting this product with their work."* Item 2 did the mirror image — it credited **this product** with a choice **they** made.

## So item 2 is empty

All five entries are gone: three restored by `R-85` on 2026-08-31, a fourth describing a device path `R-85` deleted, and this fifth that never was. Every colour value now belongs to item 1 — *resolved, not copied*.

It is kept as an **empty statement rather than deleted**, because a reader holding an earlier distribution needs to know the list emptied rather than moved.

## A second, smaller error in the same document

The sentence PR 768 wrote to replace a stale count said `R-85` moved three values *"into the byte-exact one"*. It moved them into the **re-derived** set: upstream `--border-stronger` and `--ring` are alpha composites and `--warning` is an `oklch()` call, so what ships is this product's evaluation of their expressions. Saying byte-exact over-credited them with three values that are our own output — the mirror of the motion error the first pass caught, and the third time in this work that provenance was stated in the wrong direction.

## The guard is rebuilt, and rebuilt because it failed

Its set-equality test went red the moment the notice stopped naming `--accent-contrast` as replaced: the document changed under a pin and the pin said so. That is the guard working.

It now pins an **empty** replaced set — an empty set is still a claim — and closes two gaps the second pass named:

- **It reads the dark block as well as the light one.** A dark value could previously move with nothing noticing.
- **The curve assertions read the declarations, not the file.** `curve in light` was satisfied by the comment that describes the tokens, so deleting `--ease` entirely would have left the test green. **A guard that passes because of prose is not a guard.**

Broken four ways on purpose and red each time: a dark value moved; `--ease` replaced with `ease-out` while its comment still named the curve; a departure declared in the notice with no departure in the code; and the device path returned as live code.

## Verification

Rebased onto `origin/main`; `git diff origin/main...HEAD | git hash-object --stdin` returns `cc68a733` before and after, so the base moved and the content did not.

Both suites on the pre-rebase head with `HEAD`, `DIRTY` and `TREE` identical at both ends: `FULL_MIGRATIONS_EXIT=0`, and `TEMPLATE_MODE_EXIT=1` on one known pre-existing failure — `test_the_google_button_is_busy_while_signing_in`, which is #670, diagnosed and filed today: the test waits on the button and asserts on the status text in a second round-trip, outlasting a 300 ms stub. Not in this diff, zero references to this subject, passes 4/4 in isolation.

All three copies of both synced assets hash identically. `scope` resolves to full because the change touches `design/`, so the review gate applies — the third pass is running on this head.


muhammadbayoumi:
**Routed early so this does not sit silent: checks still running, and one commit behind. There is also something specific you should check before rebasing.**

Measured against `origin/main` at `9af120a5`:

```
behind        1
rows          10 pass · 0 fail · 4 pending      <-- still in flight, nothing to read yet
overlap       none with the commit you are behind
```

Nothing is being concluded from those rows while four are pending.

## The thing to check first — you may be colliding with a merge from an hour ago

Your seven files are the three synced `supabase.NOTICE.txt` copies, the three synced `tokens.css` copies, and `tests/test_the_notice_describes_the_values_it_ships.py`.

**`#768` merged earlier today and changed exactly those files**, including that same test file:

```
a960711d  The Apache 4(b) statement describes the work it ships, and a guard keeps it that way (#768)
```

It is already in your history rather than ahead of you — you are only behind `#766`, which is panel work and touches nothing of yours. But it means **you and `#768` are two changes to the same seven files about the same subject**, and your title says the last declared departure "was never one: it is Supabase's value, resolved". So confirm your change composes with the guard `#768` added rather than contradicting it.

**A specific check that matters for synced copies.** When `#768` merged I verified all three NOTICE copies hashed identically — `3d2b776d` — because three copies that must agree are a classic place for silent divergence. Do the same before you call this ready:

```bash
for d in design extension scrapex/webui/static; do
  printf '%-24s %s\n' "$d" "$(git hash-object $d/supabase.NOTICE.txt)"
done
```

Three identical hashes, or the sync is only claimed. Same for `tokens.css`.

**And one thing not to trip on:** the colour-literal guard reads a hash followed by three hex digits as a colour value, so a three-digit PR reference like the hash form of 768 fails it inside any file it scans — which includes all three `tokens.css`. Write `PR 768`. Several sessions paid for that one.

## What to do

1. Let the four pending checks settle, then read the rows.
2. Rebase onto `9af120a5` — mechanical, since `#766` touches nothing of yours.
3. Then:

```bash
gh pr checks 776 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

**Do not read `CLEAN` as proof on its own** — two PRs today read `CLEAN` while behind `main` and missing the required `CodeQL` row. Filed as #682.

4. Because you change `scrapex/webui/static/`, expect the engine tier to run rather than the documents-only tier. A `test` row can be either; check the step conclusions of the `pull_request` run.

**Nothing on the branch will be touched.** Twenty merged; `main` green at `102e48ff`, `9af120a5` settling.

muhammadbayoumi:
## Update — you are 3 behind, the overlap is still empty, and the guard that scans your files just got *looser*

```
head       dc8bb823
behind     3
rows       14 pass · 0 fail · 0 pending
overlap    none
```

You are the only open pull request. Nothing is ahead of you in the queue and nothing is competing for the slot.

The three commits you are behind:

```
9af120a5  The updater answers honestly: the review findings 608 merged without (#766)
a7452182  A schema mismatch says which direction it runs, and the file says who wrote it (#777)
8fc077ee  Three guards: a resume that says so, a number that is not a colour, and a remap that is one hop (#786)
```

**Measured: none of them touches any of your seven files.** So the rebase is mechanical on content.

## But one of them changes a guard that reads your files, and it helps you

`#786` rewrote the colour-literal pattern in `tests/test_vendor.py` — the guard that scans all three `tokens.css` copies, which are three of your seven files. **Your CI ran against the old, stricter pattern; after rebasing you will be measured against the new, looser one.**

I extracted the shipped pattern from `#786` and ran it rather than re-typing it:

| input | old | new |
|---|---|---|
| `#420`, `#367` — a PR or issue number | **failed the guard** | passes |
| `#f0a`, `#abc` — a real 3-digit colour | failed | still fails |
| `#FFFFFF`, `rgba()`, `hsl()`, `light-dark()` | failed | still fails |

The three-digit branch now requires at least one non-decimal hex letter. So the earlier warning in my previous comment — write `PR 768` rather than the hash form — **no longer applies after you rebase**. It still applies before, which is why your current green is honest either way.

**The direction matters: a guard being loosened can only turn a red into a green, never the reverse.** So this rebase carries no new risk from `#786`.

## The sync check still stands, because that one is yours alone

Nothing in `#766`, `#777` or `#786` touches the three `supabase.NOTICE.txt` copies or the three `tokens.css` copies. Which means **nobody but you is keeping them in step**, and three copies that must agree are the classic place for silent divergence:

```bash
for d in design extension scrapex/webui/static; do
  printf '%-24s %s  %s\n' "$d" "$(git hash-object $d/supabase.NOTICE.txt)" "$(git hash-object $d/tokens.css)"
done
```

Three identical NOTICE hashes and three identical `tokens.css` hashes, or the sync is claimed rather than true. When `#768` merged I verified its three NOTICE copies at `3d2b776d`; do the same for yours.

## Then

```bash
gh pr checks 776 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

With an empty overlap the diff hash should be identical before and after. If it is not, something unexpected happened and it is worth reading.

**It will be merged the moment it is genuinely on top of `main` — measured, not read.** Nothing on the branch has been touched.

Twenty-two pull requests merged; `main` is green at `8fc077ee`.

Issue #774, opened 2026-09-07
Several worktrees, one warehouse: any newer checkout migrates it and strands the main one


muhammadbayoumi:
The development-machine half of today's incident, kept separate so it does not inflate the user-facing plan. **A real user with one installed engine cannot reach this state the same way.**

## The measured setup

```
pip install -e  ->  C:\Users\User01\source\repos\ScrapeX          (was 2a046ae0, 0.4.9, migrations to 0018)
worktrees        ->  .claude\worktrees\*                           (full checkouts, several ahead)
warehouse        ->  C:\Users\User01\.scrapex\engine\scrapex-engine.db   (one file, 1,986 MB, v19)
```

`SCRAPEX_DB_PATH` is unset and the pointer names one file, so **every checkout on this machine opens the same warehouse.** Migrations only run forward, so the newest build that starts wins permanently: it stamps v19, and every older checkout is stranded until it catches up.

That is what happened. Some session running a checkout that contained `0019` (added by #734, `923fc798`) started an engine; the main checkout was seven commits behind and could no longer open the file.

It will happen again, because parallel sessions are how this repository is worked and `main` moves several times a day.

## What makes it worse than it sounds

The stranded build is the one `pip install -e` resolves to — so it is the engine the panel starts, and the panel is the only surface. A stranded main checkout means the product is down on his machine while every worktree is fine.

## Options

- **do nothing** — recovery is a `git pull` on the main checkout, which he cannot do from the panel and should not have to. Effort 0. #769 and #770 make the failure self-explaining, which may be enough: the message would name the version to update to, and a session can perform the pull. This is the honest minimum.
- **a per-worktree warehouse**: each worktree exports its own `SCRAPEX_DB_PATH`. Isolation is complete and it is one line per worktree, but it costs the thing worktrees are for — testing against his real 1,986 MB of data. A worktree with an empty warehouse cannot reproduce an append-gate defect, which this repository has already paid for once.
- **the engine refuses to migrate when its own build is a source checkout behind its remote** — turns a silent stranding into a refusal before the damage, and the refusal can name the pull. It needs a network read of the remote at start-up, which is a new dependency on a path that must work offline. It also blocks a deliberate test of an older build.
- **record the newest build the warehouse has seen (#770) and refuse to migrate BACKWARD past it** — migrations never go backward anyway, so this adds nothing today; it would matter only if a rebuild ever lowered the version.

[...]

Issue #771, opened 2026-09-07
The panel offers Update to 0.4.11 on a source checkout, which no download can update


muhammadbayoumi:
Measured from his screen today.

The Engine page showed, at the same time:

```
Installed, not running
Installed version   0.4.9
Build               source · 2a046ae
Latest version      0.4.11   [Update available]
[ Update to 0.4.11 ]  [ Check again ]

1. Run scrapex-engine.exe. It installs for you alone …
2. Windows shows a blue notice the first time …
3. Leave its window open — that window is the engine …
```

**`Build: source` and step 1 says run an .exe.** The engine he runs is `pip install -e` against `C:\Users\User01\source\repos\ScrapeX`; running an installer would put a separate engine on the machine and change nothing about what `scrapex` resolves to. He pressed the update and reported that it did nothing. It could not.

## What the engine already says, and what the panel does with it

`scrapex/update.py`s `swap_is_possible()` reports false for a source checkout, and `update_api.py` puts the reason in `self_update_blocked_because` — *"This is a source checkout, not an installed engine. Update it with git rather than from here."* Since #608 the panel reads and renders it.

So the sentence is right and the CONTROLS beside it are not: `#engine-download` still routes to `chrome.downloads`, and `#engine-install-steps` still un-hides. The tail branch of `updateEngineReleaseUI` is reached whenever `can_self_update` is false — for ANY reason — and a source checkout is one of those reasons. **A button that cannot work is worse than no button**, and this is a button plus three instructions that cannot work.

## And it collides with the honest limit

`R-36`: the panel can never be the installer, and `chrome.downloads` is the FIRST install only. A source checkout has already had its first install — by git. There is nothing for that button to do on this screen.

## Options

- **do nothing** — the update path most likely to be used on a development machine is a dead control under a correct sentence explaining why it is dead. Effort 0.
- **a source checkout gets the sentence and no controls**: hide `#engine-download` and `#engine-install-steps` when the engine reports a source build. ~5 lines plus a DOM test. Needs a field the panel can key on — `latest.installer` is not it, and `self_update_blocked_because` is prose, not a code. Related: #681 proposes returning a code beside the sentence, which this would consume.
- **draw the real remedy** — name the command git would run, without a control. This repository forbids handing him a terminal command as an answer (`R-81`), so it would have to be a control that performs it, which is a new capability and belongs with #614.

[...]

Issue #614, opened 2026-09-05
Decide the engine's self-replace: build the swap plan_swap describes, or state the boundary permanently


muhammadbayoumi:
`plan_swap()` in `scrapex/update.py` **describes** replacing the running executable
and deliberately does not do it. PR #608 builds the caller up to `staged` — the
engine fetches the installer and checks it against the published digest — and the
screen says installing it is still his step.

**This issue is the remaining half: does the swap get built, or does the boundary
stay where it is and get stated permanently?**

## What already exists, measured

```
scrapex/update.py       plan_swap()        names the four steps as DATA, not code
scrapex/relaunch.py     spawn_helper()     already does the detached wait the swap
                                           needs, and already does it for restart
scrapex/update.py       swap_is_possible() answers whether this build could
```

So it is **measurable work rather than unknown**: the hard part — a helper that
survives the engine's exit and renames after it — is built and in use.

## Why it was not built with the caller

The owner's instruction was «النادى أولاً» — the caller first. And the reason to
respect that ordering is not caution in general: **the swap is the one path that
can destroy the engine it is updating.** Download-and-verify failing leaves the
old engine running; a rename failing halfway does not.

Building the caller first means the fetch, the digest check, the progress and
every refusal path are proven on screen before anything replaces a file.

## The decision

- **Build it** — one press updates the engine end to end, and the panel becomes
  the only interface for it, which is `R-81`'s direction.
- **Do not build it** — the boundary stays, and the sentence on screen stays
  honest about it. That is where PR #608 leaves things, and it is stable.

Either is defensible. What is not defensible is a screen that implies the first
while the second is true — which is the defect #439 recorded.

muhammadbayoumi:
**A live instance today, and it is the link that makes "everything upgrades itself" impossible.**

He updated 0.4.9 → 0.4.11 from the panel and the engine did not change, because his build is a source checkout — but the general shape is the one this issue is about, and it appeared in the field rather than in a review:

- The warehouse had been migrated to **schema v19** by a newer build (migration `0019`, added by #734).
- The running engine read **v18** and refused, correctly.
- The refusal says **"Update ScrapeX and retry"** — and for an installed engine there is no mechanical path that does it. `fetch_and_verify` works; `plan_swap` names the steps as data and deliberately does not perform them.

[...]

PR #777, opened 2026-09-07
A schema mismatch says which direction it runs, and the file says who wrote it

### LINKED ISSUE #769: One mismatch, two messages: the open path tells him to initialise a database that is only NEWER than his build
Measured on his machine today, 2026-09-07. **This is the defect that turned a seven-commit lag into a message pointing at a 1,986 MB warehouse.**

## The two copies

`scrapex/databases/domain.py:327-340` — the HEALTH path. It asks which direction the mismatch runs and, when the database is ahead, gives it its own status and its own instruction:

> Needs a newer ScrapeX. This database was written by a later version (schema v19; this build reads v18). Update ScrapeX and retry, and do not downgrade the database.

Correct. And its comment says exactly why it exists: *"Reporting it as Failed and telling the owner to restore a backup sends them to destroy good data over a one-command upgrade, so this case gets its own status and its own instruction."*

`scrapex/databases/domain.py:767-771` — the OPEN/VERIFY path. Same question, no answer:

```python
current = int(conn.execute("PRAGMA user_version").fetchone()[0])
if current != self.latest_schema_version:
    raise DatabaseMigrationError(
        f"{self.kind} database is at schema v{current}, expected "
        f"v{self.latest_schema_version}; run database initialization and retry")
```

`!=` covers both directions and the instruction fits only one. What he read on the Database page was:

> engine database is at schema v19, expected v18; run database initialization and retry

**Initialisation cannot make a v18 build read a v19 schema.** It is the remedy for a missing or unmarked database. Offered against 1,986 MB of collected data it is the most expensive wrong instruction the product has.

## Why this is duplication in the sense CLAUDE.md means

The knowledge is "which direction does a schema mismatch run, and what does each direction ask of him". It lives at `:327-340` and at `:767-771`, forty lines apart, and would change for the same reason. Change amplification is two, and one copy is already wrong — which is the failure the rule predicts.

## Options

- **do nothing** — the health card says the right thing while the page beside it says the wrong one. Effort 0. The next occurrence costs whatever a destroyed warehouse costs.
- **the open path asks the same question** — raise a distinct error (or the same one carrying the direction) when `current > latest`, worded as the health path already words it. ~10 lines, one file, plus a test that drives the ahead-state through `open()` and asserts the message names updating and never initialising. Risk: low; every caller already handles `DatabaseMigrationError`.
- **one function owns the comparison** and both paths call it — the DRY-complete answer, and it is a slightly larger edit because the two callers want different shapes (a raise and a `DatabaseHealth`). Better if the third caller ever appears.

Recommendation: the middle one, with the test. *Smallest change that removes the defect*, and *no silent failures* — a caught error must become a **true** record.

Found while diagnosing his machine. The machine itself is fixed by bringing the checkout to `origin/main`; this issue is so the message never sends anyone the other way.

### LINKED ISSUE #770: The warehouse does not record which build migrated it, so the refusal has to guess
Measured on the live warehouse today, read-only:

```
scrapex_meta:  checkout_at, checkout_holder, contract_version, database_kind,
               migration_stream, orphans_reclaimed_at, rates_last_checked,
               runtime_heartbeat, ...
PRAGMA user_version = 19
```

`migration_stream` says WHICH stream. `contract_version` says which contract. **Nothing says which engine wrote v19, or when.**

## What that costs

The refusal can only say "this build reads v18" — it cannot say what to update TO. He read:

> This database was written by a later version (schema v19; this build reads v18). Update ScrapeX and retry.

True, and it leaves the two questions that matter unanswered: update to what, and how did this happen. Diagnosing it took reading the migration list on `origin/main`, finding the commit that added `0019`, and comparing it against the checkout the engine resolves to.

With one row it would read: **"Schema v19 was written by ScrapeX 0.4.11 on 2026-09-07. You are running 0.4.9."** That names the remedy instead of implying one.

## Why in the database rather than a log

A log is per-machine and rotates — `~/.scrapex/engine.log.1` is already 5.7 MB and `engine.log` has rolled. The warehouse travels: onto another machine, into a backup, out of an archive. The fact belongs to the file it describes, which is the same argument `mark_sealed` makes at `domain.py:177-186`: *"Renaming the predecessor is not enough … The mark travels inside the file."*

## Shape

Two keys written in the same transaction that stamps `PRAGMA user_version`, so they cannot disagree with it:

- `schema_written_by` — the engine VERSION, and the commit when the build is a source checkout
- `schema_written_at` — UTC

Append-only in spirit: each migration overwrites them with the newer truth, which is exactly what a reader wants.

## Options

- **do nothing** — every future mismatch is diagnosed the way this one was, by hand, off the repository. Effort 0.
- **two meta keys, written where the version is stamped** — ~10 lines in the migration path, one integration test against the real `db/engine/schema.sql`, and a change to the refusal sentence to read them when present. Risk: low; a warehouse without the keys must still work, so the sentence degrades to what it says today.
- **a full migration ledger table** (one row per applied migration, with build and time) — richer, and it answers questions nobody has asked. Premature until a second reader exists.

Recommendation: the middle one. It is the fact that turns this class of incident from a puzzle into an instruction, and it is a prerequisite for #769 saying anything more useful than it does now.

muhammadbayoumi:
## The incident

His machine, today. A source checkout at 0.4.9 met a warehouse another build had migrated to schema **v19**. The engine refused to open it, which is right. The Database page then told him:

```
engine database is at schema v19, expected v18; run database initialization and retry
```

**His warehouse is 1,986 MB.** Initialisation cannot make a v18 build read a v19 schema — it is the remedy for a missing or unmarked file — and offered there it is the most expensive wrong instruction this product has.

## And the right answer was already in the file, forty lines up

`health()` asks which direction the mismatch runs and gives each direction its own instruction. Its own comment says why:

> Reporting it as "Failed" and telling the owner to restore a backup sends them to destroy good data over a one-command upgrade, so this case gets its own status and its own instruction.

The open path asked the same question with a bare `!=` and answered for neither direction. **One piece of knowledge, two homes, and the second copy already wrong** — the duplication `CLAUDE.md` describes by change amplification, and the failure it predicts.

`schema_ahead_detail` owns the words now. Both callers pass through it, and a test asserts they produce the **same sentence**, which the previous arrangement could not.

## The warehouse did not record who migrated it

`scrapex_meta` carried `migration_stream` and `contract_version` and nothing saying which engine wrote the schema that is in there — so the refusal could only say what it **cannot** read. Establishing the rest took reading the migration list on `origin/main`, finding the commit that added `0019` (#734), and comparing it against the checkout the engine resolves to.

`schema_written_by` and `schema_written_at` are stamped **in the same transaction as `PRAGMA user_version`**, for the reason the comment beside that stamp already gives about the number itself: a second commit leaves a crash window in which the version survives without its provenance, and the next reader is back to guessing.

A source checkout records its commit, because 0.4.9 names hundreds of them and the one that matters is the one whose migration list ends where the file's version does. The sentence now reads:

> This database was written by a later version (schema v19; this build reads v18). **Schema v19 was written by ScrapeX 0.4.11 (source a960711d) on 2026-09-07T10:41:30Z.** Update ScrapeX and retry, and do not downgrade the database.

And it degrades to exactly what it always said on every warehouse written before these keys existed — which is all of them today, his 1,986 MB one included. A file with no recorded build must not claim one.

**In the file and not in a log**, the same argument `storage.mark_sealed` already makes: the warehouse travels — onto another machine, into a backup, out of an archive — and `~/.scrapex/engine.log` rotates (its `.1` is already 5.7 MB).

## A third defect, found in the same read

The other direction said *"Use the Upgrade database button on the **Settings** screen"*. `#runtime-upgrade` has been inside `#view-database` since #679. `UPGRADE_CONTROL_LOCATION` names the screen once, because there are two uses and one had already gone stale.

**And a test was pinning the wrong screen.** `test_a_database_that_is_behind_is_called_upgradeable_not_broken` asserted `"Settings" in state.action`, under a failure message demanding *"the screen it is on"*. A guard that pins the wrong screen is worse than no guard: it makes the stale sentence look verified, which is how it survived weeks of green.

## The template gate objected, correctly

`test_the_template_is_one_honest_run_of_the_real_migrations` compares a cached schema template against a fresh honest run, and **both new keys vary per run by design** — the build that migrated the file and the instant it did.

It already excludes `database_migration.applied_at` for exactly that reason, so the two rows join that mechanism rather than the comparison being weakened. This gate reads what the **migrations** decide; who ran them is not one of those things.

## Tests

`tests/test_a_schema_mismatch_says_which_direction_it_runs.py` — integration against the real `db/engine/schema.sql` and the real migration stream, because a fixture schema cannot go out of step with the build and that is the whole subject.

Five mutations run, **five caught**: the bare `!=` restored · the Settings wording restored · the stamp removed · the provenance dropped from the sentence · a warehouse with no provenance made to claim one.

## Gates

Ruff clean. **Full suite `PYTEST_EXIT=0` on `579caba4`** — 935 template restores, 31 real migrations.

An earlier run of the same suite failed twice: a ruff `I001` on this diff's own imports, which this PR fixes, and `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` — the recorded Windows intermittent (#401, #531, #461), `ConnectionResetError` under the full run, passing twice in isolation on this branch and on clean `main`, and passing in the green run above.

One note on measuring it: the harness reported the run as exit 0 while pytest had exited 1. The recorded `PYTEST_EXIT` in the artefact is what caught the lint failure this PR fixes — a pipeline's exit code is not its first command's.

## What is not here

- **#771** — the panel offering "Update to 0.4.11" on a source checkout, with install steps whose first line is "Run scrapex-engine.exe". It edits `updateEngineReleaseUI`, which #766 is already rewriting, so it lands on top of that rather than two pull requests editing one function.
- **#774** — several worktrees sharing one warehouse. Recommendation there is to do nothing: isolation trades away the reason worktrees exist, and #769/#770 turn that stranding into a thirty-second read.
- **#614** — the engine's self-replace. It is the only link in the upgrade chain with no automatic hop, and the only thing that would make a mismatch **fix** itself rather than explain itself. His decision.

Closes #769. Closes #770.

Milestone: Nothing is stranded: every version upgrades itself, and a mismatch explains itself.


muhammadbayoumi:
**Routed early so this does not sit silent: your checks are still running, and the only thing that will be in the way afterwards is the base.**

Measured against `origin/main` at `9af120a5`:

```
behind        2
rows          10 pass · 0 fail · 4 pending      <-- still in flight, nothing to read yet
overlap       none
```

**No conclusion is being drawn from those rows while four are pending.** A check set that still carries pending rows gives a verdict about a moment that has not finished; two reads of it seconds apart can disagree without either being wrong.

## Your overlap is empty, and that is unusual today

You change four files — `scrapex/databases/domain.py` and three test files. The two commits you are behind:

```
102e48ff  Back up from the page that lists the backups, and watch it happen there (#735)
9af120a5  The updater answers honestly: the review findings 608 merged without (#766)
```

Both are panel work — `extension/` and `tests/test_panel_dom.py`. **Neither touches any file you touch.** So your rebase should be genuinely mechanical, which is rare here: most branches today collided in `extension/app.js`.

**One thing to check anyway, because it is the file you change:** `scrapex/databases/domain.py` was changed earlier today by `#669`, which moved the baseline comparison from `!=` to `<` so that a squashed warehouse can still be upgraded. The comment above it explains why, measured on the owner's own 2.08 GB file. Your subject — *a schema mismatch says which direction it runs* — is about the same comparison. Read that change before you rebase; it is already on `main` and it is deliberate.

Related and worth knowing: `#678` fixed a fixture in `tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py` that had been passing vacuously since the squash — it was reading `origin/main`, a moving state, when the tests needed a fixed moment. If any of your three test files reason about pre-squash versus post-squash chains, that is the precedent.

## What to do

1. Let the four pending checks settle, then read the rows.
2. Rebase onto `9af120a5`.
3. The checks GitHub's own status does not answer:

```bash
gh pr checks 777 | awk -F'\t' '{print $1}' | sort -u
```

**Do not read a `CLEAN` verdict as proof on its own.** Two PRs today read `CLEAN` while behind `main` *and* missing the required `CodeQL` row — a required check with no row cannot say anything against you. Filed as #682.

4. And confirm the `test` row was the engine tier, not a documentation-tier pass wearing the same name. Since you change `scrapex/`, you should also see `migration-authority` run `The whole suite, against the real migration stream` rather than skip it.

**It will be merged the moment it is green and genuinely on top of `main` — measured, not read.** Nothing on the branch will be touched.

Twenty pull requests merged; `main` is green at `102e48ff` with `9af120a5`'s run still settling.

PR #817, opened 2026-09-09
Every colour says whose it is, and now something checks the answer

### LINKED ISSUE #780: Four inline PUBLISHED/derived provenance markers in tokens.css point the wrong way
`design/tokens.css` annotates colour declarations with `PUBLISHED` (a literal Supabase publishes) or `derived` (this product's evaluation of one of their expressions). **The notice designates those markers as the authority** — *"WHICH IS WHICH IS RECORDED AT EACH VALUE in design/tokens.css"* — so a wrong marker is a wrong Apache-2.0 §4(b) statement, not a stale comment.

**Fourteen markers. At least four are wrong, and every one over-credits Supabase** — the same direction as the three errors already corrected in the easing curves, the byte-exact set, and `--accent-contrast`.

## Verified against the pinned commit `86c813ec`, computed and searched independently

The literal search is proven working by two controls: `3ecf8e` returns **96** hits in `supabase/supabase` and `097c4f` returns **25**. Both are real published brand values.

| marker | shipped value | upstream | hits | verdict |
|---|---|---|---|---|
| `--focus` dark, *"--ring dark PUBLISHED"* | `#2f7a57` | `--ring: oklch(from var(--primary) l c h / 55%)` — an alpha expression, flattened over **our** `--bg` | **0** | **derived** |
| `--amber-ink` dark, *"--warning-foreground PUBLISHED"* | `#080503` | `--warning-foreground: oklch(from var(--warning) clamp(...) calc(c * 0.08) h)` — a relative-colour expression | **0** | **derived** |
| `--line-strong` dark, *"scale-900 dark, PUBLISHED"* | `#3d3f3e` | the comment itself says their `--border-stronger` is `#393a39`, a different value | **0** | **neither** — it matches no upstream token we can find |
| `--accent-hover` dark, *"brand-600 dark PUBLISHED"* | `#85e0ba` | to be confirmed the same way | — | suspect |

### The arithmetic, which also validates the method

`--primary` = `oklch(0.76 0.15 159)` = `#45cd8e`. At 55% over the light `--bg` `#fdfdfd` that is **`#98e3c0`** — byte-for-byte the light `--focus`, which correctly carries **no** marker. Over the dark `--bg` `#131413` it is `#2e7a56` against the shipped `#2f7a57`, one unit per channel of compositing rounding.

**So the light value and the dark value of the same token are marked differently, and the dark one is the wrong way round.**

## Why this is its own change

The vocabulary is internally consistent everywhere else: `PUBLISHED` is reserved for their stepped scale values (`brand-200`, `warning-300`, `destructive-300`), which upstream really are published numbers, and every semantic token they compute — `--destructive`, `--primary-foreground`, dark `--amber` — is correctly marked `derived`. The exceptions are these few, and they are all semantic tokens wearing the scale vocabulary.

**And nothing asserts any marker.** `tests/test_the_notice_describes_the_values_it_ships.py` pins values and the notice's entry list; it never reads a marker. That is the gap that let four of them drift, and it is the reason this is a change rather than a comment edit: the fix is the markers **plus** a guard that reads them.

## Acceptance

Every `PUBLISHED` marker in `design/tokens.css` names a value that appears as a literal in Supabase's source at the pinned commit; every `derived` marker names one that does not. `--line-strong` dark either cites the token it actually came from or is marked `derived` with its derivation. A guard asserts the marker vocabulary against the upstream literals, and fails when a marker is flipped. All three synced copies stay byte-identical.

Found by the third four-dimension merge-gate pass, raised independently by three of its four dimensions, and each claim survived an adversarial attack.

muhammadbayoumi:
Closes #780.

`design/tokens.css` marks each colour `PUBLISHED` (a literal Supabase declares) or `derived` (this product's evaluation of one of their expressions). `design/supabase.NOTICE.txt` delegates the Apache-2.0 §4(b) per-value record to those markers — *"WHICH IS WHICH IS RECORDED AT EACH VALUE in design/tokens.css"* — **so a marker pointing the wrong way is a wrong statement of changes.** Nothing checked them, and three had drifted.

## Audited mechanically against their own theme files

Fetched verbatim at the pinned commit `86c813ec` rather than searched. **21 marked declarations, 19 already correct**: every `PUBLISHED` marker naming a stepped scale value converts to exactly what ships — nine literal matches across both themes — and every `derived` marker names an expression.

| marker | upstream | verdict |
|---|---|---|
| dark `--focus` *"--ring dark PUBLISHED"* | `--ring: oklch(from var(--primary) l c h / 55%)` — an alpha expression | → `derived` |
| dark `--amber-ink` *"--warning-foreground PUBLISHED"* | `--warning-foreground: oklch(from var(--warning) clamp(...) calc(c * 0.08) h)` | → `derived` |
| dark `--line-strong` *"scale-900 dark, PUBLISHED"* | **there is no `--scale-900` anywhere in their themes** | → `derived`, with its real derivation |

The `--ring` correction validates the method: `--primary` is `#45cd8e`, and at 55% over the light `--bg` that is **`#98e3c0`** — byte-exact with the light `--focus`.

**And `--accent-hover` dark, which a review pass flagged as suspect, is correct**: `--brand-600` dark is `154.9deg 59.5% 70%` = `#85e0ba`, exactly what ships.

## The third one almost went the other way, and the arithmetic is why it did not

A review pass concluded `#3d3f3e` is a live departure — which would have made the notice's *"no colour value is deliberately replaced"* false again, one day after it was written.

Resolving their whole chain says otherwise, and **the chain validates itself on three values first**:

| their token | computes to | ships here as |
|---|---|---|
| `--foreground` | `#edefee` | dark `--text` |
| `--background` | `#131413` | dark `--bg` |
| `--card` | `#181a19` | dark `--surface` |

Then `--border-stronger` — `oklch(from var(--foreground) l calc(c * 0.5) h / calc(5% + 45% * var(--contrast-border)))`, with `--contrast-border` = `0.525²` = `0.275625` and so alpha `0.174031` — flattens to **`#393a39` over `--background`** and **`#3e3f3e` over `--card`**.

We ship `#3d3f3e`: the `--card` composite, one unit of compositing rounding, **the same choice the light block makes and records as the honest transcription**. It is theirs. The marker was wrong; the value was not.

## The guard checks both directions, because both are licence statements

`tests/test_the_provenance_markers_say_who_owns_each_value.py` transcribes Supabase's stepped numeric scales — the complete set of colour literals their system publishes, everything else being computed — and asserts that a `PUBLISHED` marker names one of them **and equals it converted**, while a `derived` marker names something that is not one.

**The first stops us claiming their authorship; the second stops us claiming ours.** It also asserts the notice still delegates the record here, so the guard cannot outlive the obligation it guards.

It fails on `origin/main`'s `tokens.css`, catching both wrong markers. Four further mutations, **each verified as applied before its result was read**, red on all four: a correct `derived` flipped to `PUBLISHED`; a real literal marked `derived`; a `PUBLISHED` value moved off their literal; a `PUBLISHED` marker naming a token they compute.

## Two defects in my own work, recorded because they change how the next one is done

**The first mutation run lied.** A `sed` pattern with the wrong run of spaces matched nothing, the suite stayed green, and I read that as a hole in the guard rather than a mutation that never happened. Every mutation now verifies the file's hash changed before its result is read.

**The guard's first version classified by the comment text**, taking the theme from whether the word "dark" appeared in the note. On `main` the dark `--amber-ink` note omits it, so that declaration was checked against the light table — it still failed, **for the wrong reason**. It now takes the theme from which block the declaration is in.

## Verification

Both suites green on `38cee698` with `HEAD`, `DIRTY` and `TREE` identical at both ends — `TREE=853f1fbc`, `TEMPLATE_MODE_EXIT=0`, `FULL_MIGRATIONS_EXIT=0`. No intermittent this run. Zero behind `main`. All three copies of `tokens.css` hash identically.

`scope` resolves to full, so the four-dimension gate applies and is running on this head.


muhammadbayoumi:
**Merged as `17ad3ea1`.** Recording what was measured, because one of the checks is worth having on the record and one thing was left undone deliberately.

## The merge bar

```
onTopOfMain (MEASURED)  YES — behind 0, ahead 1, via merge-base --is-ancestor
rows                    14 pass · 0 fail · 0 pending  (settled; two watches agreed)
required contexts       all nine present, CodeQL included
engine tier             "Tests (Python engine + Python-vs-frozen contract)" success
                        docs-only and extension-only steps both SKIPPED
                        — confirmed on the pull_request run AND the push run
migration-authority     "The whole suite, against the real migration stream" success
version.py              untouched, and correctly so — see below
migrations              none
```

## Two properties of this change that were proved rather than assumed

**1 · The three `tokens.css` moved in lockstep.** A palette token has more than one home, and the failure mode is a change landing in one copy. The three diffs were hashed separately:

```
design/tokens.css                 26a3a9adfa51f6722d8ada32cc2dcaa6fed6c663
extension/tokens.css              26a3a9adfa51f6722d8ada32cc2dcaa6fed6c663
scrapex/webui/static/tokens.css   26a3a9adfa51f6722d8ada32cc2dcaa6fed6c663
```

Identical. **2 · No colour value moved.** Token declarations were extracted with comments stripped and compared: **178 before, 178 after, zero altered**, in all three files. So this is a provenance-marker change and nothing else — which is why no VERSION bump is owed under R-35: no contract changed and nothing user-visible moved.

## The guard was checked for vacuity, and it discriminates

A 245-line new test that passes is exactly where a green can be empty — a guard can pass because nothing contradicts it. This one was run both ways in isolation, same guard, only the markers swapped:

```
guard + THIS branch's markers   ->  24 passed
guard + main's old markers      ->  3 failed, 21 passed
```

and the three failures were precisely the three claims this PR corrects — `dark --line`, `dark --focus`, `dark --amber-ink`. **The guard fails on the state it was written to catch.** It also carries its own anti-vacuity assertion (`len(PUBLISHED) >= 8 and len(DERIVED) >= 8`) so a change to the comment format cannot silently empty the corpus, which is the retrofit another guard needed the hard way this week.

Worth noting for whoever extends it: `TOKENS` resolves to `design/tokens.css` only, so the provenance markers in the other two copies are not read by this test. Byte-identical diffs make that a distinction without a difference today; it stops being one the moment the three drift.

## The branch was NOT deleted, deliberately

`--delete-branch` was requested and refused, and the refusal was correct:

```
cannot delete branch 'fix/the-provenance-markers-say-who-owns-each-value'
used by worktree at '.../.claude/worktrees/keen-jepsen-e5486e'
```

The remote branch also still stands, at `38cee698`. **Nothing was forced.** That worktree is checked out on this now-merged branch, so deleting either ref would strand it. Left for that session to clean up when it is done with the worktree — it is stale, not harmful.

---

*Posted by the merge session. Twenty-nine merged; `main` green at `17ad3ea1` (`0.4.13`), with the owner's checkout and the live engine both in step.*

muhammadbayoumi:
## Correction to my merge comment above

I wrote that this PR's guard was "checked for vacuity, and it discriminates," on the strength of running it against both marker sets. **That check was one-directional and I did not say so.**

All three failures it produced came from `test_a_published_marker_names_a_literal_supabase_actually_publishes`. The mirror test, `test_a_derived_marker_does_not_name_a_published_literal`, produced **zero** — and that asymmetry was in my own output, unexamined. A two-sided guard where one side never fires is the thing to interrogate, and I reported the aggregate instead.

**#820 found why.** The `derived` side returns early when the named token is absent from the hand-typed table:

```python
literal = SUPABASE_SCALES[scheme].get(named)
if literal is None:
    return
```

The `PUBLISHED` side has no such escape — it asserts `literal is not None`. So the table's 43 entries against the 107 literals Supabase's sources declare leave 64 published literals that a `derived` marker could under-credit in silence, and an invented name on a `derived` marker could not be caught at all.

So my statement rested on a real measurement but claimed more than it established: it showed this guard was **non-vacuous**, not that it was **sound**. The merge itself stands — the change was safe and its three corrections were real — but the verification behind that sentence was weaker than the sentence.

#820 replaces the transcribed table with one generated from their sources and pins the fixture's commit to the notice's. It is being watched and will be merged once it is on top of `main`.

---

*Posted by the merge session.*

Issue #818, opened 2026-09-09
The modal mechanics are now written twice: open, close, focus return and the Tab trap in two dialogs


muhammadbayoumi:
**Recorded, not fixed.** The panel now has two modal confirmations and the
mechanics are written twice.

`#disconnect-veil` (app.html) with `openDisconnectDialog` /
`closeDisconnectDialog` / `disconnectDialogIsOpen` / `trapDisconnectFocus`, and
`#restore-veil` with the same four shapes added in the restore change. Both do
the same four things:

- toggle `hidden` on a `.modal-veil` at body level, because a view is
  transformed for 180ms during navigation and becomes the containing block for
  `position: fixed`
- remember `document.activeElement` and give focus back on close
- keep Tab inside the card while it is open
- answer Escape, innermost first, from the one document-level handler

That is the same knowledge in two places and it would change for the same
reason — a change to how a modal behaves has to be made twice or it is made
wrong once. By this repository's own DRY test, it is duplication rather than two
things that merely look alike.

**It was left duplicated on purpose.** Generalising it would have rewritten a
working confirmation inside a diff about restoring a database, and "no refactor
rides along" is the stronger rule at that moment. A third modal is the point at
which leaving it costs more than fixing it.

## Options

1. **A `confirmDialog({veil, dialog, copy})` object owning all four behaviours**
   *(Recommended when a third one arrives, or now if this is picked up on its
   own)* — one place for the mechanics, and each caller keeps its own copy text
   and its own action. Cost: it touches a working destructive control, so it
   needs both dialogs' tests green before and after, and the Escape chain has to
   keep answering innermost-first.
2. **Extract only the focus trap and the Escape registration**, leaving open and
   close per dialog. Smaller, and it splits the mechanics into "shared" and
   "not" along a line nobody will remember.
3. **Do nothing while there are two.** Defensible: two copies of a twenty-line
   behaviour is cheap, and the wrong abstraction is dearer than the duplication.
   The cost is paid the first time one is changed and the other is not.

No recommendation between 1 and 3 without knowing whether a third confirmation
is coming — that is a roadmap question, not a code one.

muhammadbayoumi:
**Roadmap approved 2026-09-24 (#1040) · phase 4: Components to Supabase's atoms and patterns** · `rescope`
- Scope: One modal mechanism (focus in and back, Tab trap, Escape innermost-first); A5 first
[...]

Issue #806, opened 2026-09-09
The product backs up and cannot restore: no restore control exists, and the RTO is therefore unbounded


muhammadbayoumi:
**IMMEDIATE · P0.** The product can now take a backup and has never been able to
restore one. Nothing in the panel restores, and the code says so itself.

`extension/app.html:2047` offers "Check the latest backup", and
`extension/app.js:7338` explains why the label was corrected:

> ASKS DRIVE, DOES NOT DOWNLOAD. This button never restored anything

`fetchFromDrive` reads Drive's `size` for the pointed-at object and returns a
sentence ending "this only checks that it is there."

## What that costs, in the owner's own targets

He set **RTO = 4 hours**. Today the recovery time is **unbounded**: there is no
path from a Drive object back to a working warehouse that a person can walk. The
`KPI: Recovery Time` reads `unknown` for that reason, not for lack of measuring.

Every runbook in the assessment is blocked here — (a) granular restore, (c)
ransomware DR and (d) failover all begin with a control that does not exist. And
he ruled **restore by buttons only**, so a documented command line does not
discharge it.

## Scope

Three controls, smallest first, each usable alone:

1. **Inspect a backup** — read the manifest for a chosen point and show dataset
   names, row counts and digests. No archive download.
2. **Restore everything** — fetch, unpack, `PRAGMA integrity_check`, compare row
   counts against the manifest, then hand the file to the engine as the live
   warehouse. Refuse rather than swap on any mismatch.
3. **Restore one dataset** — the same, into a temporary table, with a diff shown
   before `Apply`.

The write goes through the engine, which is the sole writer; the panel must never
open the database.

## Options

1. **All three, in that order** *(Recommended)* — 1 is small and immediately
   useful (it answers "what is in my backup?"), 2 discharges the RTO, 3 is the
   one used most often. Each ships alone.
2. **Restore everything only.** Cheapest path to a defensible RTO. Leaves the
   common case — one bad dataset — needing a full swap.
3. **Do nothing.** Only defensible if the product ships as backup-only and says
   so in the interface, which contradicts the assessment's acceptance bar and the
   4-hour RTO.

Blocks: the verification worker has nothing to verify with until (2) exists.

muhammadbayoumi:
## One of the three controls has shipped — #819, merged as `c30b098c`

**Restore from a local snapshot exists.** The panel could take a backup and could
not put one back; it can now. Read from `origin/main`, not a worktree:
[...]

Issue #30, opened 2026-07-30
اسعار ومواقع


muhammadbayoumi:
moci.gov.qa · psa.gov.qa
بيتومين اختراق 60/70
قطر

img.mhuc.gov.eg
(الفهرس: mhuc.gov.eg/أرشيف_نشرة_أسعار_مواد_البناء)

اسعار مواد البناء القاول السعودية

muhammadbayoumi:
Inventoried in #824 — this list is now part of one deduped record of every host the project has ever named as a source (241 hosts), against what the Source library actually shows (12). Nothing here is declared or promised; the issue records where it sits and what showing it would cost.

muhammadbayoumi:
**Closed as a duplicate of #824, whose register already holds every host here.**

- `moci.gov.qa` / `psa.gov.qa` — `docs/CANDIDATE-SOURCES.md:119`, with the bitumen grade
  «بيتومين اختراق 60/70» quoted verbatim.
- `img.mhuc.gov.eg` — `:125`, with the Arabic index path, **plus three facts this issue does
  not have**: not probed, a document archive rather than a page, and a non-ASCII path that
  needs percent-encoding.

#824 is the better record — 241 hosts deduped against the 12 the library shows, with the
missing control named. Nothing needed carrying over; a pointer was added there anyway.

Issue #808, opened 2026-09-09
A backup nobody has restored is a hypothesis: automated restore verification with a smoke test


muhammadbayoumi:
**IMMEDIATE · P0.** Not one backup has ever been restored, so "0 restore errors"
is an unmeasured claim. A completed upload is a completed job; only a restore is
a backup.

`verifyLatest` asks Drive for the object's metadata. That proves an object of the
right size exists — not that it opens, not that its rows are there, not that the
engine will accept it.

## What the check must assert

```
restore_one_point(point_id, into=scratch_dir)      # chunks -> archive -> unpack

assert sqlite("PRAGMA integrity_check") == "ok"    # it is a database
assert sqlite("PRAGMA foreign_key_check") == []    # its references hold

for table, expected in manifest["row_counts"].items():
    assert sqlite(f"SELECT count(*) FROM {table}") == expected

# A SMOKE QUERY, because integrity_check passes on a database that is intact
# and empty. This asserts the warehouse still ANSWERS.
assert sqlite("SELECT count(*) FROM contractors WHERE name IS NOT NULL") > 0

record_audit("restore-verified", point_id, elapsed_s, bytes_read)
```

Table names come from the manifest, never from user input. It runs on a scratch
copy and touches the live warehouse never — a verification that can corrupt what
it verifies is worse than none.

The elapsed seconds are the **measured** Recovery Time, which is the only honest
way to know whether the 4-hour RTO holds. Today that KPI reads `unknown`.

## Options

1. **Weekly automatic verification of the newest point, plus a monthly full
   restore into an isolated database** *(Recommended)* — weekly catches a broken
   chunk store early and costs one restore's egress (0.61 GiB against 6.2 GiB
   free); monthly measures the real RTO.
2. **On demand only, from a button.** Cheapest. Its weakness is the one this
   whole assessment is about: a check nobody remembers to run reports nothing,
   and silence reads as health.
3. **Verify only after each backup.** Catches an upload defect immediately and
   misses decay in older points — which is the failure the six-month history
   exists to survive.
4. **Do nothing.** Ships an untested restore path to the public, with a KPI that
   cannot be reported.

Depends on the restore path existing.

PR #819, opened 2026-09-09
The copy he can put back, and the deadline a restore actually needs

muhammadbayoumi:
Closes the first item of #806 — the panel could take a backup and could not put
one back. First slice of milestone #30.

## What was already there, and what was not

`POST /api/storage/restore` is in `main`, guarded (it refuses any path
`list_backups` did not return), and it handles the Windows open-handle problem
that used to make every restore a 500. **It had no caller in the panel.**

Measured on the owner's machine, 2026-09-09:

```
GET /api/storage  ->  backups: 5 entries
  keys: bytes · modified_at · name · path · tag · taken_at
  2026-09-09T05:00:09Z  pre-upgrade  2,148,061,184  path=YES
  2026-09-07T10:08:16Z  pre-upgrade  2,083,115,008  path=YES
  2026-09-06T09:45:03Z  pre-upgrade  2,080,301,056  path=YES
  2026-08-30T04:52:46Z  manual       1,527,279,616  path=YES
  2026-08-23T14:53:54Z  manual       1,320,628,224  path=YES
```

The Database page already read `backups[0]` to write "Newest …" and dropped the
other four. So he could see that five copies existed, learn the date of one, and
reach none of them.

The panel calls **two** of the ten `/api/storage/*` routes — `GET /api/storage`
and `POST /api/storage/integrity`. Neither is a restore.

## The route I did not add

A branch, `claude/the-restore-he-never-had`, built this once and proposed a new
`GET /api/storage/backups`. None of its panel work reached `main` (checked by
content, not ancestry — eight ids, six functions and its test file are all
absent). Its design is good and this reuses it.

**Its route is not needed any more.** It was written on 2026-09-02, when
`GET /api/storage` did not carry the list; it does now. Adding it would be a
second read of one fact — and that is exactly how this card came to report a
count of 3 while 4 copies sat on the disk, which its own harness comment
records. So: no new route, **no contract change, no engine version bump.**

## The deadline is part of the fix

`/api/storage/restore` inherited `destinationData: 5000` from the `storage`
group rule. `storage.restore` does not rename:

```
health(backup)         30,784 ms   MEASURED on a 2,148,061,184-byte file
shutil.copy2           2.1 GB
health(copy)           30,784 ms   the same check, again
_same_contents         4.3 GB read
                       ~113 s total, against a 5,000 ms bound
```

And aborting cancels nothing on the far side: the engine holds the write lock
and completes the switch. **Under 5,000 ms the panel reported failure over a
restore that succeeded** — about the one action that replaces his warehouse.
`restoreCopy: 600000` is derived from the measurement above, covers a warehouse
near 11 GB, and names that as its expiry date.

## Where it lives, and why not on Manage account

The Database page, under the card that already counts the copies. A snapshot is
on this disk: it is the fast way back after a change goes wrong, and it dies
with the disk. The Drive copy is what survives losing the machine. Those are
different facts and they stay on different screens — sharing one word for both
was the confusion, not the existence of both.

## Guards, and the mutants that prove them

Eight mutants, eight caught:

| mutant | caught by |
|---|---|
| only the newest copy is drawn — the original defect | lists every copy |
| the question does not read its argument | names which copy |
| the path is rebuilt from the name instead of sent back | sends back the path |
| a refusal closes the question | refused restore |
| stale rows are left offering a restore | offers no copy |
| the page is not re-read afterwards | sends back the path |
| the deadline row moved below the group rule | the ordering guard |
| `restoreCopy` set to the generic value | `notEqual(destinationData)` |

The pristine tree was restored byte-for-byte after each, verified by digest.

The two failures I hit while writing them were mine, not the code's:
`wait_for_selector` waits for *visible*, so a hidden veil never satisfies it —
corrected to `state="hidden"`.

## Decisions taken in the open

- **`manageRow` reused despite its `manage-account-row` class name.** Its CSS is
  not scoped to that view (checked). The name is wrong for a shared component
  and renaming it is not this diff's business; a second row style for this
  screen would be the duplication worth avoiding.
- **The modal mechanics are now written twice** — open, close, focus return, the
  Tab trap. That is real duplication by this repository's own test, and
  generalising it would rewrite a working destructive control inside a diff
  about restoring a database. Recorded as **#818**, with the note that a third
  modal is where leaving it costs more than fixing it.

## Not in this slice

Restoring from a Drive backup, and restoring a single dataset — the other two
controls #806 asks for. This one is the foundation both reuse: download to a
local file, then the same route. And it is the one that works with the five
copies already on his disk today.


muhammadbayoumi:
**Nothing is wrong with this pull request. It cleared every gate, and then `main` moved out from under it by one commit — #803, merged about a minute before you would have gone in. One mechanical rebase and it goes.**

## What was measured at head `0e6cf6c6`, before `main` moved

```
onTopOfMain (MEASURED)   YES — behind 0, ahead 1, via merge-base --is-ancestor
rows                     14 pass · 0 fail · 0 pending  (settled)
required contexts        all nine present, CodeQL included
engine tier              "Tests (Python engine + Python-vs-frozen contract)" success
                         docs-only and extension-only steps both SKIPPED
                         — on the pull_request run AND the push run, agreeing
migration-authority      "The whole suite, against the real migration stream" success
version / migrations / contracts   none touched
diff                     +443 / -0 in 5 files — purely additive
```

**It was genuinely ready.** It is being returned only because of the base, and the base moved because of a merge, not because of anything on this branch.

## Your `.mjs` tests do run — worth knowing, because `pytest` does not run them

You added 30 lines to `extension/tests/startup.test.mjs`, and the full Python tier cannot execute those. They are not unverified: the **`contract-parity`** job carries them, and it is a required context.

```
success   Contract parity gate (JS vs frozen normalize vectors)
success   Extension unit tests (node:test — no package.json, no dependencies)
success   Apps Script gate (the pasted file's own logic, run for real)
```

Confirmed on both the `pull_request` and `push` runs. So `node --test extension/tests/*.test.mjs` really ran your new cases.

## The rebase is provably mechanical

`main` is now `8d3f3920`. The commit you are missing is:

```
8d3f3920  A census that asks the page, on a random sample instead of the first 1,200 (#803)
```

It added **two brand-new files and modified nothing**:

```
A  tools/profile_card_census.py
A  tests/test_the_profile_census_uses_a_second_instrument.py
```

Your five files are `extension/app.html`, `extension/app.js`, `extension/startup.js`, `extension/tests/startup.test.mjs`, `tests/test_panel_dom.py`. **Measured overlap: empty.** Note in particular that #803 did *not* touch `tests/test_panel_dom.py`, which is the file a census change might plausibly have reached.

## The check that proves the rebase moved your base and nothing else

Use the **three-dot** form, measured from your merge base. It does not care where `main` goes, so it stays valid even if `main` moves again while you work:

```bash
# expect: 0c0ee49b37262bc67eb2c5aeb420f0f7e1177f62
```

If that matches after the rebase, your content is untouched. **Do not use the two-dot form** (`origin/main..HEAD`) for this — it measures the whole difference between the two trees in both directions, so it also reports whatever `main` has that you do not, and it will not match. I handed a session the two-dot form earlier today and had to retract it for exactly that reason.

Then the two things GitHub's status does not answer:

```bash
gh pr checks 819 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

**And do not trust `mergeStateStatus` on its own.** It currently reads `UNKNOWN` and is recomputing; earlier today it read `CLEAN` on a PR that was measurably one commit behind with a complete, settled row set. That is filed as #682.

## What happens next

You are the only open pull request. **Rebase onto `8d3f3920`, let the set settle to fourteen rows, and it will be merged** — measured, not read. Nothing on this branch has been touched and nothing will be; "Update branch" was deliberately not used, because updating the tree from underneath you voids the green that was just measured on it.

---

*Posted by the merge session. Thirty pull requests merged; `main` is green at `8d3f3920` (`0.4.13`), with the owner's checkout and live engine both in step.*

muhammadbayoumi:
**Merged as `c30b098c`.** Thirty-one.

Your rebase was clean and the invariant held exactly:

```
asked for   0c0ee49b37262bc67eb2c5aeb420f0f7e1177f62
measured    0c0ee49b37262bc67eb2c5aeb420f0f7e1177f62
```

So the base moved and your content did not.

Final measurement on the rebased head `900f014d`, re-run from scratch rather than carried over from the pre-rebase verification:

```
onTopOfMain (MEASURED)  YES — behind 0, ahead 1
rows                    14 pass · 0 fail · 0 pending (settled) · nine names
engine tier             "Tests (Python engine + Python-vs-frozen contract)" success
                        docs-only and extension-only steps SKIPPED
migration-authority     "The whole suite, against the real migration stream" success
node suite              "Extension unit tests (node:test)" success
                        — your 30 new .mjs lines really ran
                        + Contract parity gate, + Apps Script gate
all of the above on the pull_request run AND the push run, agreeing
version / migrations / contracts   none touched
diff                    +443 / -0 across 5 files
```

And the three protective fixes were confirmed present **on your branch's tree**, not inferred from the diff: the `< migration.number` comparison in `scrapex/databases/domain.py` with its `AT OR ABOVE THE BASELINE` comment, #817's provenance guard, and #803's census tool.

## One thing recorded, not blocking

This adds a user-visible control to the panel and does not move a version number. No guard asks it to — `contracts/` is untouched, `tests/test_the_version_moves_when_the_contract_does.py` is green, and the engine/manifest gap (`0.4.13` against `0.3.3`) is the documented, intentional one. So nothing here is wrong under any encoded rule.

It is noted only because R-35 moves the **extension's** number on a user-visible change, and which number that is in practice is the open question in **#658** — blocked on a manual `version.json` in `mbiX-hub`. Recorded there rather than opened as anything new, and no action is asked of you.

---

*Posted by the merge session. `main` is green at `c30b098c` (`0.4.13`), with the owner's checkout and live engine in step.*

muhammadbayoumi:
## Correction: the derivation in this PR quoted a cold reading as if it were the warm one

The `restoreCopy` bound is right. **The measurement I wrote beside it was stated
as a flat fact when it was one of two very different cases**, and it contradicts
a figure already in the same file.

Re-measured today on the same 2,148,061,184-byte file:

```
health() cold             42,874 ms      the file comes off the disk
health() warm              5,045 ms      the page cache still holds it
PRAGMA quick_check         2,168 ms
PRAGMA foreign_key_check   1,532 ms
health(integrity=False)         6 ms
```

`integrityScan`, twenty lines above the row I added, is derived from the **warm**
figure — "5.9 s on his warehouse at 2.08 GB" — and it is correct for what it
bounds: a scan of the warehouse the engine already has open.

I wrote "`health()` alone is 30,784 ms" without saying it was a cold read and
without reconciling it with the number already there. **Anyone re-measuring warm
gets 5 s and concludes the comment is invented** — which is the same visible
wrong record this repository treats as worse than a silent one, and the same
class of defect I filed #804 and #805 about earlier today.

The number does not change: cold is the case that governs a restore, because a
restore reads a backup nobody has touched, twice, and then moves 6.45 GB copying
and comparing. About 130 s on this warehouse, against the 5,000 ms it used to
inherit. `600000` still covers ~11 GB and the cold/warm spread is now written out
as the reason the bound is not set near the measurement.

Corrected in `extension/startup.js` and in the guard's comment, on the branch for
#808 — the next slice, where the cost of a check is the subject rather than an
aside.

Issue #659, opened 2026-09-05
HANDOVER · the design system · PR 820 is rebased and needs its suite pair re-run


muhammadbayoumi:
**Read this first.** Every figure verified against `origin/main` at `7a0309a4` on 2026-09-10, by measurement and not from memory. Where something is unverified this says so.

---

## 1 · One open pull request, and what it needs

| | |
|---|---|
| **PR 820** | `fix/the-marker-guard-is-blind-in-the-direction-that-matters`, head **`717926a2`**, 3 commits |
| state | rebased onto `7a0309a4`, **0 behind**, `git merge-base --is-ancestor origin/main HEAD` confirms it, working tree clean |
| CI at the previous head | 14/14 SUCCESS |
| **what it needs** | **the suite pair re-run on `717926a2`.** See §2 — the measurement below is on the pre-rebase tree. |

Everything else from this work is merged: PR **764** (the authority list), **765** (the design document says what the code does), **768** and **776** (the Apache §4(b) statement), **817** (the provenance markers).

---

## 2 · The suite state, stated exactly

Measured on `76eb3486`, the head **before** the last rebase — `TREE=9fc0b8c9`, fingerprint identical at both ends of both runs:

| | |
|---|---|
| `SCRAPEX_FULL_MIGRATIONS=1` | **`EXIT=0`** |
| template mode | **`EXIT=1`** — two failures |

**Both failures are filed and neither belongs to this branch.** Each checked the same way: not in the diff, and zero references to `PUBLISHED`, `derived`, `tokens.css`, `supabase` or `provenance` in the failing file.

- `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` at **line 266** — **#401** / **#531**, the real Windows defect. Line 254 is a different red: the test's own vacuity guard, **#461**. **Check the line number before attributing it.**
- `test_the_interval_is_jittered_not_metronomic` — **#649**, whose own failure text confirms its diagnosis: `jitter left the band: [1.0476…, 1.1337…, 0.7830…]`.

**The rebase superseded that measurement**, and `34b68f88` (PR 851) changed CI and three test files in the same window, so re-run rather than carry it forward.

**Run the two suites separately, not chained.** Three chained runs were lost in a row — one to a session limit, one to a tree the next commit superseded, one killed mid-run. Chained, a kill costs both.

---

## 3 · What phase 1 closed, and the pattern to expect

[...]

Issue #649, opened 2026-09-05
The jitter flake asserts on interval-minus-elapsed with a real clock: zero margin at the floor, and a one-line fix


muhammadbayoumi:
`tests/test_http_fetcher.py::test_the_interval_is_jittered_not_metronomic` failed CI run **33600442619** (2026-09-02) with `0.6999747700621614` — **25 microseconds** below its own floor. It is not a random-seed problem and the jitter implementation is correct.

## Root cause, read from the source

`scrapex/connectors/base.py:748-753`:

```python
interval = self._min_interval_s
if self._jitter:
    interval *= 1.0 + random.uniform(-self._jitter, self._jitter)
elapsed = time.monotonic() - self._last_request_at
if elapsed < interval:
    time.sleep(interval - elapsed)          # <- what the test records
```

The test (`tests/test_http_fetcher.py:297`) fakes **only** `time.sleep`:

```python
monkeypatch.setattr("scrapex.connectors.base.time.sleep", slept.append)
```

`time.monotonic` stays real, so `elapsed` is the true wall-clock cost of one mocked `get()`, and **every recorded value is strictly below the drawn interval**. The assertion at `:306` is `all(0.7 <= w <= 1.3)` — the floor has **zero margin** and the ceiling is unreachable. The failure was not a bad draw; it was the machine being ~25µs slower than free.

## The fix — deterministic, no skip, no weakened assertion

Freeze the clock the pacer reads:

```python
monkeypatch.setattr("scrapex.connectors.base.time.monotonic", lambda: 0.0)
```

Each recorded wait then equals the drawn interval exactly. Verified across 800 runs / 8,000 samples, including a run with 30 ms injected per call (which fails ~40% of the time today): the band held every time and the distinctness assertion held every time.

The sibling test directly below it (`:309-317`) has the same defect with a 100 ms margin instead of zero, and the same one-line freeze makes it exact (`[5.0, 5.0]` instead of `any(w >= 4.9)`).

## Why it belongs to a class, not to itself

[...]

PR #820, opened 2026-09-09
The marker guard was blind in the direction every error here has run in

muhammadbayoumi:
PR 817 merged a guard for the `PUBLISHED`/`derived` provenance markers in `design/tokens.css`. **Its own merge-gate panel found three defects in it**, all confirmed by measurement, and one made it unsound in exactly the direction the five prior provenance errors ran in.

## It typed Supabase's values by hand, and the omission was not a smaller check but a wrong one

43 entries, transcribed. Their sources declare **631 token names and 107 colour literals** — so **64 literals were missing**, every one of the 52 in `global.css` among them.

The consequence: **marking a value Supabase publishes outright as `derived` passed**, because the guard had no literal to contradict it. An adversary proved it — two byte-exact literals marked `derived`, suite 26/26 green. **Under-crediting them is the direction the easing curves, the byte-exact set and `--accent-contrast` all ran in.**

The table is now `tests/fixtures/supabase-design-tokens.json`, generated by `tools/read_supabase_tokens.py` reading their seven CSS sources. **Nothing is transcribed.** The tool reads the commit *out of* `design/supabase.NOTICE.txt` rather than taking it as an argument, and a test asserts the fixture's commit equals the notice's — so the guard cannot end up checking a different Supabase than the notice claims to describe.

## It could not catch an invented name on the `derived` side, and two were already shipping

The name check returned early whenever the named token was unknown, so `--scale-900` was caught only because it *also* claimed `PUBLISHED`. Neither of these was caught:

| claim | reality |
|---|---|
| `--danger-contrast`, both themes, records `--destructive-fg` | no token of that name exists in any of their files. Theirs is `--destructive-foreground` — **spelled out in full five lines away in the same block** |
| `--accent-hover` light records `brand/80 flat` | they declare no `--brand` at all; the token is `--brand-default`. The value is right — `#3fcf8e` at 80% over the light `--bg` is `#65d8a4`, computed — so only the name was wrong |

Every marker's name is now checked against the 631 they declare, whichever word it carries. Run against `main`'s `tokens.css`, the rebuilt guard fails on exactly those three.

## Two mechanical defects

**The regex crossed newlines.** `\s*` between the semicolon and the comment let a declaration bind to the *next* comment once its own was stripped, inventing a marker for a token that has none. Proved directly: `--line: #232423;` matched the `scale-900` comment two lines below it. Horizontal whitespace only, now.

**It stripped the theme word as a substring**, so `--colors-gray-light-900` parsed as `--colors-gray--900`. That token is real and is `#8f8f8f` — the value `design/tokens.css` records as theirs and shipped until 2026-08-31 — so the next session to restore it would have hit a false red. Whole words only.

## Verification

**Five mutations, each verified as applied before its result was read**, red on all five:

- a byte-exact literal of theirs marked `derived` ← *the one that was silent*
- `PUBLISHED` naming a token they compute
- a `PUBLISHED` value moved off their literal
- an invented name on a `derived` marker ← *the direction that was blind*
- the fixture read at a different commit than the notice pins

That verification step exists because an earlier mutation run misled me: a `sed` pattern with the wrong run of spaces matched nothing, the suite stayed green, and I read that as a hole in the guard.

Both suites green on `d5b6d0a4` with `HEAD`, `DIRTY` and `TREE` identical at both ends — `TEMPLATE_MODE_EXIT=0`, `FULL_MIGRATIONS_EXIT=0`, no intermittent. Rebased onto `origin/main`; the diff hash is `e9744f61` before and after, so the base moved and the content did not. All three `tokens.css` copies hash identically.


muhammadbayoumi:
**Your central claim is confirmed independently, and it corrects something the merge session put on the record. One commit behind `main`; the rebase is mechanical.**

## First, the correction — because it is mine to make

When #817 merged I posted that its guard had been "checked for vacuity, and it discriminates," on the strength of running it both ways:

```
guard + #817's markers      ->  24 passed
guard + main's old markers  ->  3 failed, 21 passed
```

**That test was one-directional, and I did not say so.** All three failures came from `test_a_published_marker_names_a_literal_supabase_actually_publishes`. The mirror test contributed **zero** failures — and that asymmetry was sitting in my own output, unexamined. Had I asked why one side of a two-sided guard never fired, I would have found what you found.

Read directly from `main` just now:

```python
# PUBLISHED side — no escape
literal = SUPABASE_SCALES[scheme].get(named)
assert literal is not None, (...)

# derived side — the escape
literal = SUPABASE_SCALES[scheme].get(named)
if literal is None:
    return
```

That `return` is the blindness, exactly as you describe it. **A `derived` marker naming any token outside the hand-typed table passes unexamined** — and the table holds 43 entries against the 107 literals their sources declare, so 64 published literals could be under-credited in silence, and an invented name on a `derived` marker could not be caught at all. Verified: `43` entries in `main`'s `SUPABASE_SCALES`, `107` in your fixture.

So my earlier statement stood on a real measurement but claimed more than it established: it showed the guard was **non-vacuous**, not that it was **sound**. Corrected here rather than left standing.

## What was measured on your branch

```
markers only, no colour moved   178 -> 178 token declarations, ZERO changed,
                                in all three tokens.css
three copies in lockstep        531871a6ac1da213d06587bcb83e931bf9ee6681 (identical)
rows                            14 pass · 0 fail · 0 pending
invariant                       e9744f611c6d4d3f2d92d4c0f458bdc2216a0e23
                                — matches the hash your body claims
version / migrations / contracts  none touched
```

Generating the table instead of transcribing it, and pinning the fixture's commit to the one `design/supabase.NOTICE.txt` names, is the right shape: it removes the class rather than the three instances. The commit-equality test is the part that keeps it honest over time.

## What blocks it

```
merge-base --is-ancestor origin/main HEAD   ->  NO
behind_by 1, ahead_by 1        main is now c30b098c
```

The commit you are missing is **#819**, merged a few minutes ago:

```
c30b098c  The copy he can put back, and the deadline a restore actually needs (#819)
```

It touched five files — `extension/app.html`, `extension/app.js`, `extension/startup.js`, `extension/tests/startup.test.mjs`, `tests/test_panel_dom.py`. **Measured overlap with your six files: empty.** Nothing it changed goes near `tokens.css`, the fixture, the guard or the reader tool, so this rebase is genuinely mechanical.

## The check after the rebase

Use the **three-dot** form, from your merge base — it stays valid even if `main` moves again while you work:

```bash
# expect: e9744f611c6d4d3f2d92d4c0f458bdc2216a0e23
```

Do not use the two-dot form for this; it reports whatever `main` has that you do not, and will not match. Then:

```bash
gh pr checks 820 | awk -F'\t' '{print $1}' | sort -u    # expect nine names
```

Your rows currently read 14 pass while the status field reads `UNKNOWN`/`BEHIND`; after the rebase let the set settle to fourteen with zero pending before reading anything. **And do not trust `mergeStateStatus` alone** — it read `CLEAN` today on a PR that was measurably one commit behind with a complete, settled row set (#682).

## What happens next

**It will be merged the moment it genuinely contains `c30b098c`** — measured with `merge-base --is-ancestor`, not read from a status field. Nothing on this branch has been touched and nothing will be; "Update branch" was deliberately not used, because it would void the green measured on this tree.

---

*Posted by the merge session. Thirty-one merged; `main` is green at `c30b098c` (`0.4.13`).*

muhammadbayoumi:
## Do not merge on my word — the suite pair on this head is incomplete

Three attempts at a green pair on `7d4b3c8a`:

- one was killed mid-run by a session limit and its second suite never wrote a result;
- one measured a tree my next commit superseded, so I stopped it rather than spend half an hour on a void measurement;
- one is running now.

**What is verified on this head:** the marker guard's own 48 assertions green, red on the pre-branch `tokens.css`, and red on five mutations — each confirmed as having changed the file before its result was read.

**What is not:** the full pair. `TEMPLATE_MODE_EXIT` and `FULL_MIGRATIONS_EXIT` with a `HEAD`/`DIRTY`/`TREE` fingerprint identical at both ends. I will post it when it lands.

## And this branch is on its seventh review pass, which is the argument for stopping here

The passes returned 3, then 0 — a panel whose four agents had died on a session limit, which reports as an empty survivor list and means *did not run*, not *found nothing* — then **9**.

Two of the nine are fixed in the third commit because they made the change indefensible: `colors.css` declares all 204 of its names twice with 185 differing, and the generator's linear scan stored the **dark** value under `root` for every one; and **nothing was verifying the fixture at all** — restoring its 107-entry version left every assertion green.

The other seven are **#822**, including two `must fix`. They are filed rather than folded in because CLAUDE.md says a change whose reviews do not converge is too big to review, and folding them in would make pass eight larger than pass seven.

Handover with the full thread: **#659**.

muhammadbayoumi:
## Verification on `76eb3486`, both suites, and one of them is red

Rebased onto `origin/main` — 0 behind, `git merge-base --is-ancestor origin/main HEAD` confirms it, and `git diff origin/main...HEAD | git hash-object --stdin` returns `24e6b6b2` before and after, so the base moved and the content did not.

Fingerprint identical at both ends of both runs: `HEAD=76eb3486`, `DIRTY=0`, `TREE=9fc0b8c9`.

| | |
|---|---|
| `SCRAPEX_FULL_MIGRATIONS=1` | **`FULL_MIGRATIONS_EXIT=0`** |
| template mode | **`TEMPLATE_MODE_EXIT=1`** — two failures |
| CI | 14/14 SUCCESS, `MERGEABLE`, `CLEAN` |

### Both failures are filed and neither is this branch's

| failure | issue | discriminators |
|---|---|---|
| `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` at **line 266** | **#401** / #531 — the real Windows defect, *not* the vacuity guard at line 254 | not in this diff; **0** references to `PUBLISHED`, `derived`, `tokens.css`, `supabase` or `provenance` |
| `test_the_interval_is_jittered_not_metronomic` | **#649** — *"asserts on interval-minus-elapsed with a real clock: zero margin at the floor"* | same, both zero |

`#649`'s failure text confirms its own diagnosis: `jitter left the band: [1.0476…, 1.1337…, 0.7830…]` — values outside the band by a zero margin, which is exactly what that issue describes.

**I am not presenting this as green.** Template mode is red, by two known intermittents, on a tree whose full-migrations run passed.

### Ran the suites one at a time, and the reason is worth passing on

Three consecutive chained runs were lost — one to a session limit, one to a tree my next commit superseded, one killed mid-run. Chaining them meant a kill cost both. Split, a kill costs one.

### One observation from the commit this branch rebased onto

**#830** adds a CI step that counts skips out of the JUnit XML with a floor on how many tests actually ran — for a file that collects six and loses five at run time while `-q` reports green.

That guards one half of the shape. The other half is in this branch's own guard and is recorded on **#822**: nine of eleven `derived` parametrizations return before their assertion. Those are **passes, not skips**, so a skip count cannot see them. They return *correctly* — a `derived` marker naming a token Supabase computes has no literal to contradict it — but a test that passes having asserted nothing is the same silence #830 exists to break, and nothing in the repository catches that form.

muhammadbayoumi:
## Rebased and handed over — this needs its suite pair re-run before it merges

Head **`717926a2`**, rebased onto `7a0309a4`, **0 behind**. `git diff origin/main...HEAD | git hash-object --stdin` returns `24e6b6b2` before and after, so the base moved and the content did not. All three `tokens.css` copies hash identically.

**GitHub read `MERGEABLE CLEAN` while this branch was genuinely two commits behind.** `git merge-base --is-ancestor origin/main HEAD` was the answer both times. Measure the merge state; do not read it.

### What is measured, and on which tree

On `76eb3486` — the head **before** this rebase, `TREE=9fc0b8c9`, fingerprint identical at both ends of both runs:

| | |
|---|---|
| `SCRAPEX_FULL_MIGRATIONS=1` | **`EXIT=0`** |
| template mode | **`EXIT=1`** — two failures, both filed, neither this branch's |

- `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` at **line 266** → **#401** / #531, the real Windows defect. Line 254 is a different red — the test's own vacuity guard, #461. Check the line number.
- `test_the_interval_is_jittered_not_metronomic` → **#649**, and its failure text confirms that issue's own diagnosis: `jitter left the band: [1.0476…, 1.1337…, 0.7830…]`.

Each checked the same way: not in this diff, and **zero** references to `PUBLISHED`, `derived`, `tokens.css`, `supabase` or `provenance` in the failing file.

### Why that measurement should not be carried forward

The rebase superseded the tree it was taken on, and **`34b68f88` (#851) changed `.github/workflows/ci.yml` and three test files** in the same window — including how the suite is run. A green measured on a superseded base is not a green, and that applies to a red as well.

**Run the two suites separately, not chained.** Three chained runs were lost in a row here: one to a session limit, one to a tree the next commit superseded, one killed mid-run. Chained, a kill costs both.

### Handover

**#659**, now in the *Resume here* milestone: what the three commits do, the four ways this guard has itself been wrong, the six wrong-direction provenance claims the gate caught one pass at a time, and the seven findings on **#822** that are filed rather than folded in.

muhammadbayoumi:
**A merge session looked at this and is handing it back rather than reviewing
it.** Not taken, not merged, nothing changed. The reason is that the record
already answers the question a review would have asked.

## Why an eighth pass is the wrong spend

**#822 is open, with two *must fix* and six *should fix* against exactly what
this PR changes, none of them fixed.** Its own opening says why they were
recorded instead of folded in: the guard has been through **seven passes**, and
the findings went **3 → 0 (the pass did not run) → 9**. `CLAUDE.md` rules that a
change whose reviews do not converge is too big to review and should be split.

Reviewing it now would be pass eight, spending roughly 350k tokens to re-derive
what is already written down. The counts rule merged in #888 sizes a panel from
changed lines; it does not know that seven panels already ran.

The first must fix is not cosmetic: `_hsl` rounds ties to even where a CSS engine
rounds away from zero, so a one-unit difference produces a **false red on a
truthful `PUBLISHED` marker** — and the repair a reader applies to a convincing
false failure is to downgrade a true marker to `derived`, writing a false
statement into the Apache §4(b) record. The second records 48 alpha-bearing
declarations as opaque.

## State, measured just now

| | |
|---|---|
| head | `717926a2` |
| CI on that exact SHA | `lint` · `scope` · `contract-parity` · `test` all green |
| behind `main` | **9 commits** — so that green was measured against a base that no longer exists |
| assignee | none |
| open since | 2026-09-09 |

Two handovers also still describe it as unfinished, and both are still true:
**#659** ("PR 820 is rebased and needs its suite pair re-run") and **#863**
("#820's verification did not finish"). This PR's own comments record three
attempts at a green suite pair — one killed by a session limit, one stopped as
void, and one that came back **red**.

## What worked today on the same shape

#828 carried a guard that four passes kept finding defects in while never
finding any in its 96 lines of product code. On his ruling it was split: the
product fix merged (#828), and the guard was **rebuilt against the recorded
findings** rather than reviewed again (#882). One scoped pass then closed it, and
#875 with it. That pattern converged where four passes had not.

The equivalent here would be to rebuild the generator and the guard against #822
as the specification. That is a session's work, not a merge session's tail, and
the scale decision is the owner's.

## Not done here, deliberately

Not assigned, not rebased, not merged, and no finding of #822 acted on. A
rebase would have produced a fresh green that says nothing about the eight open
findings, which is the misleading artifact this hand-back exists to avoid.

muhammadbayoumi:
## Rebased, and #822's eight ranked findings are answered

Head **`ada8339`**, rebased onto `main` — 0 behind, and all three patch-ids identical
before and after, so the base moved and the content did not. `main` had touched **none** of
this branch's six files since it diverged, so nothing here was stale except the base.

**CI: 8 of 8 green on `ada8339`**, CodeQL included.

Every number in #822 was re-measured against Supabase's sources at the pinned commit
rather than taken on trust — that issue asked for exactly this, and **three of its figures
did not survive**. The full accounting is on #822, which is now closed; the short version:

- **rounding** — confirmed, and settled by asking a browser. Chromium 151 returns
  `#0091ff` and `#00254d` where Python's `round()` gave `#0090ff` and `#00254c`.
- **alpha** — confirmed. 48 stored entries, 62 raw declarations, two separate mechanisms.
- **the `_` exclusion** — confirmed to the number: 11 names.
- **bare `H S% L%` triples** — **did not reproduce.** There are none in their sources.
- **comment braces** — real in code, bites nothing today: 75 comments, none holds a brace.
- **the second dark block** — worse than recorded: 28 colour declarations, **zero** markers.

**Regenerating changes 51 values** (3 rounding, 48 alpha) and adds 11 names. Every marker
in `design/tokens.css` still passes, so the licence record itself needed no correction.

### Seven mutations, and three survived the first attempt

Each fix was broken on purpose. Three survived at first — the 8-digit-hex alpha
truncation, the comment-blanking removal, and dropping the second dark block — because
guarding them against the real file proved nothing: no value there is an 8-digit hex, no
comment holds a brace, and the second dark block carries no marker. Those three are now
driven directly with synthetic input, and **all seven are caught**.

Local: 77 in this guard, 1112 in the `-m extension` tier (exit 0), ruff clean, and
`--check` reports the fixture up to date.

### Recorded, not closed

- #1016 — 18 colour declarations name a Supabase token and carry no marker word. All 18
  name tokens their sources COMPUTE, so none hides an under-credited literal; adding them
  changes the §4(b) record, which is your call.
- #1017 — #822's own five unscheduled optionals, two now measured. The sharpest:
  `design/supabase.NOTICE.txt:28` lists `apps/design-system/styles/globals.css` as a source
  and the tool does not read it — that file declares **23** custom properties, so the
  notice and the guard disagree about what the sources are.

Issue #606, opened 2026-09-05
A crawl re-picked after an engine restart reports 0/56 and never says it is re-proving from disk


muhammadbayoumi:
**Measured on his own crawl, 2026-09-03, and he asked about it before we did.**

His muqawil listing crawl was at `running 7/56` with 2,859 pages stored. The engine was relaunched at `13:19` — two launches twenty seconds apart, the first of which died — and the panel went to `preparing 0/56`.

Nothing was lost. `jobs.py` maps `PREPARING -> QUEUED`, `RUNNING -> QUEUED` and `RESUMING -> QUEUED` at startup, which is correct: that is how a job whose engine was killed gets picked up instead of claiming to run for ever. The job was re-dispatched and **re-proved its closed cells from disk at zero network requests** — which the job card faithfully reported as `Requests 0` beside `3/56`.

**That pair of numbers reads as a stalled crawl and is in fact the resume working perfectly.**

### The two things that are actually wrong

**1 · The log does not distinguish a resume from a start.** The opening lines are identical either way — `listing crawl of 56 cell(s) as job-<ref>`. A resume should say what it found: how many cells this run ref already holds pages for, and that re-proving them costs no request. `contractors.crawl` already computes `already_stored` per cell; the runner never asks.

**2 · `progress_done` is reset to 0 before the re-proof.** Both readings are defensible — `0/56` is true of this run's closed-cell count, `7/56` is true of what the warehouse can prove — which is why this needs a decision rather than a patch.

### Options

| | effort | risk | what it costs |
|---|---|---|---|
| **do nothing** | none | the next resume costs another question | he has to ask what the numbers mean, every time |
| **log the resume** (recommended) | small — one line, read `already_stored` once before the first cell | none; it is a log line | nothing |
| **log it and mark the figure** `3/56 (re-proved from disk)` | medium — the card renders a qualifier | a new field the panel must not misread | the card stops reading as a crawl that lost its place |

**Recommendation: the middle one now, the third only if he says the figure still misleads him.** The report is the part he reads; the number is defensible as it stands.

### Why this is worth a number rather than a shrug

He watched it happen and asked what it meant. The honest answer needed a read of `crawl_job`, `job_log_entry`, `generic_page_snapshot` and the process table. **A resume he cannot recognise from the panel is a resume he has to ask about** — and the panel is his only interface.

### And a second-order hazard worth stating here

[...]

Issue #725, opened 2026-09-06
role="progressbar" on the crawl bar never carries a value


muhammadbayoumi:
**Priority P1** · from the 2026-09-06 design-system audit (dimension: Accessibility).

## What is true today
extension/app.html:140 declares `<div class="bar" role="progressbar" aria-labelledby="act-progress-label">` with no aria-valuenow, aria-valuemin, aria-valuemax or aria-valuetext. renderProgress (extension/app.js:4139-4166) computes `pct` for the determinate case and writes only `bar.style.width = pct + "%"`. The panel's other progressbar does it correctly: driveProgress (extension/app.js:6211-6223) sets aria-valuenow, and extension/app.html:1946-1947 declares valuemin/valuemax. So 1 of 2 progressbars is announced. The label beside it is a live region, which softens but does not fix it — the bar itself announces as "progress bar" with no value.

## What Supabase does instead
Their accessibility page is silent on progressbar specifically; ARIA requires aria-valuenow for a determinate progressbar, and the repository's own second progressbar already follows that.

## Size, counted
3 attributes on 1 element in extension/app.html, 1 setAttribute in renderProgress, plus removing valuenow in the indeterminate branch. 1 test function; extension/app.js:4143-4155 already distinguishes the two cases.

## Evidence
```
extension/app.html:140-141; extension/app.js:4139-4166; extension/app.js:6211-6223; extension/app.html:1945-1947
```

## Depends on
nothing

## Acceptance criteria
A run with a known denominator gives the bar aria-valuenow matching the percentage in the label; a run with no denominator leaves valuenow absent.

---
*Filed as part of the phased roadmap. The audit that produced it fanned eight dimensions across independent agents and put every finding through an adversarial pass; 20 findings were refuted and are not here.*

muhammadbayoumi:
**Roadmap approved 2026-09-24 (#1040) · phase 5: States: working, busy, progress, reduced motion, contrast** · `keep`
- Scope: The crawl progressbar carries aria-valuenow (ARIA; a gap)
- Cost: 1 markup site, 1 function with 2 branches
- Depends on: none
- First step: Test: with a known total, aria-valuenow is present; with none, it is absent. It fails today.
- Side: design

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #821, opened 2026-09-09
A Jobs page he can follow, and the mini-player stops adopting whichever job was entered last

### LINKED ISSUE #778: The panel adopts jobs[0], so a blocked job hides a working one — and he read it as nothing working
**He pressed a control, watched the Run screen, and reported that nothing was working — while the job was at 620 of 938 pages with zero failures. The panel was faithfully showing him his OTHER job.**

`pollJob` adopts whichever active job the list happens to hand back first:

```js
jobs = (await api("/api/jobs?active_only=true&limit=5")).jobs;
const job = jobs[0] || null;
```

and `list_jobs` orders `job_id DESC` (`scrapex/jobs.py:163`), so **the newest active job wins** — regardless of whether it is doing anything.

## Measured on his engine, 2026-09-07

He had two `profile_crawl` jobs, 18 seconds apart. What the panel was given:

```
[0]  job_1a95d29ebf89   preparing   0/938     <- adopted, blocked on the host lane since 10:33
[1]  job_034c51a29deb   running     620/938   <- the one actually fetching
```

The second job was blocked inside `_CrawlAdmission`'s per-host lane — **which is the politeness protection working correctly** — so it sat at `preparing` with `progress_done` 0, one log line, and a heartbeat 20 minutes stale. The panel drew that, and `+1 other` beside a frozen zero does not say that the other one is the live one.

The engine was never wrong. `/api/jobs/job_034c51a29deb` answered `progress: {done: 620, total: 938}` throughout, and the sweep finished the job it was given.

## Why this is not only about a double press

Any two active jobs hide each other, and the one hidden is the one making progress whenever the idle one is newer. That happens without a double press:

- a queued job waiting on `job_capacity` is newer than the running one it waits for;
- a job blocked on a host lane is newer than the job holding it;
- a scheduled run firing while a hand-started one is mid-flight.

In every case the panel shows the still one and hides the moving one. **`jobs[0]` is not a choice — it is whatever the ordering produced.**

## What would fix it, and the trade in each

| | effort | what it costs |
|---|---|---|
| **do nothing** | none | the owner cannot tell a working job from a stuck one whenever two are active, and his first read is "it is not working" |
| **adopt the job that is making progress** — prefer `running` over `preparing`/`queued`, then the newest among equals | small | one comparison; a genuinely stuck `running` job still wins over a queued one, which is right — it is the one that owns the worker |
| **show every active job in the mini-player** rather than one plus a count | medium | the mini-player is a small surface and three rows in it is a different design; it also answers "why is that one not moving" without being asked |
| **name what is blocking the adopted job** — `queued_behind` already exists in the payload and is `null` here | small, and complementary | it explains the still job instead of replacing it |

**Recommendation: the second, with the fourth.** Prefer the job that owns a worker, and say what a still job is waiting for. `queued_behind` is already in the shape and already null for a lane-blocked job, which is its own small gap.

## And the second half, filed separately

Nothing stopped a second identical sweep being queued for the same source. That is issue **[the sibling]** — a button that works twice where it should work once.

### LINKED ISSUE #793: A Jobs page: he asked for it, and the panel can show one active job out of 163
**His request, 2026-09-07, with a screenshot of the rail: «سجل انى اريد اضافة jobs بحيث اقدر اتابع مثل ما قلت مهمة كذا انتهت او مهمة كذات توقفت لانى لا ارى ذلك من خلال extension»** — a Jobs page, so he can follow that one job finished and another stopped, because he cannot see that through the extension.

He is right, and the panel has no surface for it at all.

## Measured, not assumed

**His warehouse holds 163 jobs**, and the panel can show at most one of them:

```
completed               104
failed                   28
cancelled                23
partially_completed       4
completed_with_errors     3
running                   1
```

**The panel asks one question, once**, in `pollJobOnce`:

```js
jobs = (await api("/api/jobs?active_only=true&limit=5")).jobs;
const job = jobs[0] || null;
```

- `active_only=true` — so **every finished job is invisible**. All 162 settled ones above cannot be reached from the panel, whatever they say;
- `jobs[0]` — so **only one active job is ever drawn**, and it is whichever the ordering handed back first (#778);
- `limit=5`, with `jobs.length - 1` rendered as a bare `+N queued` count beside it — a number, not a list.

**There is no other reader.** A grep for a panel request that lists jobs without `active_only` returns **zero** hits. The engine already answers `GET /api/jobs?limit=N` for all statuses; nothing in the extension calls it.

**And there is no `jobs` page in the rail.** Its eleven pages are `profile`, `database`, `engines`, `source`, `run`, `data`, `finance`, `appearance`, `sources`, `console`, `settings`.

## What today cost him, which is the argument rather than a story

In one day, on his own machine, all four of these were invisible from the panel and each was answered only by my reading his warehouse:

1. a sweep at 620/938 hidden behind a blocked one showing 0/938 — he read it as *"nothing is working"* (#778);
2. a duplicate sweep queued by a second press, which he could not see was a second job (#779);
3. a cancelled job that ran to completion anyway, whose log never said "cancel" (#791);
4. a paused interpretation at 300/909, invisible while a newer job ran, and reachable only once that newer job settled.

**Every one of those is a question about a job.** The panel is his only interface, and it can currently answer *"what is the newest active job doing"* and nothing else.

## What the page needs, and what it must not become

- **every job, not the active ones** — kind, source, status, progress, when it started and when it settled;
- **the log of any one of them**, which the engine already serves at `GET /api/jobs/{ref}/logs` and the panel already renders for a single job;
- **the controls where the job is**, so pause / resume / cancel are not only on whichever job the mini-player adopted;
- **a still job must say what it is waiting for.** `queued_behind` is already in the payload and is `null` for a job blocked on a host lane — its own small gap, and the thing that made #778 unreadable.

**Not a second activity log.** The Console page exists and the Run page already renders one job's log; this is a *list* that opens one of them, not a third place where log lines are drawn.

## Two decisions that are his

- **Where it sits in the rail.** A Jobs page reads `GET /api/jobs`, so it shows nothing without an engine — which puts it in the second group by the rule `test_the_rail_groups_say_which_pages_need_an_engine` states. He overruled that grouping once already, for the Database page, so the placement is his to say rather than mine to assume.
- **Whether Run keeps its mini-player.** With a Jobs page, the mini-player is either a shortcut to the running one or a duplicate of a row on that page.

### LINKED ISSUE #796: An engine restart re-enters a running interpret job from page 1, and his progress bar resets with nothing saying why
**Measured on his warehouse, 2026-09-07, while he was updating the engine. Two `dataset_interpret` jobs each logged their opening preamble five times.**

```
job_5155b86ba455   14:31:42 entered
                   14:31:53 orphan sweep: was running with no runtime behind it -> queued
                   14:31:54 entered again          <- 1 second later
                   14:32:47 sweep -> queued        14:32:47 entered again
                   14:33:34 sweep -> queued        14:33:35 entered again
                   14:36:31 sweep -> queued        14:36:31 entered again
                   14:36:34 approved 432 page(s)

job_0e69563b5820   14:26:40 entered
                   14:42:54 entered  ·  14:43:53 sweep -> 14:43:53 entered
                   14:44:15 sweep    ->  14:44:16 entered
                   14:44:29 sweep    ->  14:44:29 entered
                   14:46:12 approved 432 page(s)
```

The engine's own `/api/health` says its build was sealed at **14:33:34Z** — he was restarting it through that window, which is why the sweep ran again and again. **The sweep is not the defect here.** `reclaim_orphaned_jobs` at loop startup passes no `keep` and that is correct by its own premise: at startup nothing of this runtime is running. #767 made it *name* each job it requeues, which is the only reason this is measurable at all.

## The two defects the re-entry exposes

**1 · A re-queued `dataset_interpret` job starts again from page 1.** The preamble is written once per entry, immediately after the runner sets the job `preparing` with `progress_done = 0`. So **his progress bar went back to zero eight times across the two jobs**, and each pass re-walked all 469 pairs. Correctness survives — the second walk finds the rows already written and reports *"432 unchanged and confirmed"* — but nothing tells him a restart happened, and the bar is the only thing he can see.

Compare `directory_crawl`, which resumes under an inherited run ref (#642) and skips what is stored. **`dataset_interpret` has no resume of any kind.**

**2 · Two runtimes could have walked the same pass at once.** `job_0e69563b5820` was re-queued at 14:43:53 and completed at 14:46:12, so a thread of the *old* runtime was still working while a new runtime had already queued it for dispatch. Nothing in the evidence proves two threads ran together — the pass is idempotent, so it would not show — but nothing prevents it either, and the interpret pass takes a write lock.

## Options

1. **A checkpoint for the interpret pass**, the same shape `directory_crawl` already has: record the last page pair closed, and re-enter from it. Removes both the re-walk and the reset. Cost: `checkpoint_json` on a job that currently leaves it null, plus a resume path in the runner.
2. **Say it out loud and leave the re-walk** — log *"re-entered after a restart, starting again from page 1"* and keep the count monotonic in the panel. Cheap, honest, does not stop the wasted 3 minutes.
3. **Do nothing.** 3 minutes per restart and a bar that resets without saying why. That reset is one of the four things that made him press the button four times today (#779).

**Recommendation: (2) now, (1) when the Jobs page (#793) lands**, because a Jobs page that lists a job's entries makes the restart legible where the mini-player cannot. **(1) alone is the larger change and it needs his word on whether an interpret pass is worth checkpointing at all** — it reads from disk and asks the site for nothing, so the only thing it buys back is his time and his progress bar.

muhammadbayoumi:
Closes #793. Closes #778. Closes #796.

**His rulings of 2026-09-09:** the page sits above Run, and the Run mini-player stays.

---

## What was wrong

He asked for this page on 2026-09-07, because he could not see through the extension that one job had finished and another had stopped. Measured that day, his warehouse held **163 jobs** — 104 completed, 28 failed, 23 cancelled, 4 partially completed, 3 completed with errors, 1 running — and the panel asked one question, once:

```js
api("/api/jobs?active_only=true&limit=5")   →   jobs[0]
```

So every **finished** job was unreachable, and exactly one active job was ever drawn. In a single day that hid a sweep at 620/938 behind a blocked one showing 0/938 — he read the panel as nothing working — a duplicate sweep he could not see was a second job, a cancelled job that ran to completion, and a paused interpretation at 300/909.

**The engine needed no change.** `GET /api/jobs?limit=N` already answers every status with the full per-job view, and `GET /api/jobs/{ref}/logs` already serves the log. Nothing in the extension called either.

## #778 is the same line

`jobs[0]` over a newest-first list draws whichever job was entered **last**, which on 2026-09-07 was one blocked on the per-host politeness lane. Adoption is an **order** now, and `preparing` ranks below `running` deliberately: the engine's `WORKER_HELD_STATUSES` counts `preparing` as held — correct for deciding whether a pause must wait for a safe boundary — and it is the wrong discriminator here, because the job that cost him 33 minutes of silence sat in `preparing` for the whole of it.

## #796: a re-entered job says so

Two `dataset_interpret` jobs logged their opening preamble **five times each**, one second after an orphan sweep named them:

```
job_5155b86ba455  entered 14:31:42, then 14:31:54, 14:32:47, 14:33:35, 14:36:31
```

He was updating the engine through that window, so the sweep was right and requeueing was right. What was wrong is that every runner then writes `progress_done = 0` and says nothing — **his progress bar went back to zero eight times across two jobs** with no line anywhere explaining it.

`started_at` was already the signal: all three runners guard their entry write with `{} if job["started_at"] else {...}`, so whether this is a re-entry is a value they already read and discard. `jobs.note_a_re_entry` turns it into a sentence carrying the number he watched disappear — **read before the reset**, which is the contract; called after, it would report every re-entry as having reached nothing.

**The consequence belongs to the caller.** An interpretation re-reads pages off disk and asks the site for nothing; a sweep skips what is already stored under its run ref. One sentence for both would be false for one of them. `directoryjob` is deliberately untouched: #606 already gave the listing crawl its own resume line, and a second line above it would be two lines about one event.

## What the page does, and what it refuses to do

- **every job, not the active ones** — kind, source, status, progress, and when it started and settled;
- **why a still job is still, in the part that is always visible.** It refuses to invent a reason: `queued_behind` is `null` for a job blocked on the lane, so the honest sentence is that it waits for the site's turn rather than for a named job;
- **the controls on the job**, which is half of what this page is for — the job he wanted to stop was not the job the panel was drawing. They sit **inside** the row so a mis-click on a list of 163 cannot cancel a 33-minute crawl, and the set offered is matched to what `set_control` can honour: a settled job gets none, because the route answers 409;
- **not a second activity log.** Console exists and Run draws the live job's log; this is a list that *opens* one. `<details>` rather than a click handler and a class toggle: keyboard-operable and announced without a line of script, and the log is fetched the first time a row opens rather than 200 times on load.

## Six things the existing suite and the harness caught, and every one was mine

1. I read `fetch` as `done`/`total`, which is `progress`'s shape. The engine serves `requests`/`expected` — `_fetch_progress` calls itself *"the ONE place this is computed"* — so the page drew no numbers at all. **An assumed payload is the same defect as an assumed file path.**
2. The CSS went into `extension/panel.css`, which `cat >>` **created**: `app.html` loads tokens, components and `app.css`. Caught by `test_every_class_in_markup_resolves_to_a_rule` — the guard `app.html:1596` already records a mistake against.
3. `jobsview.js` was missing from the panel harness's hand-maintained module list, so `rowsFrom is not defined` threw inside `showView` and every guard failed with *"0 of 4 jobs drawn"* rather than with the reason. That list's own comment predicts exactly this.
4. `app.js` already declared `waitingLine(s)` for the dataset card. **The panel's modules are separate scopes in the browser and one scope in the harness**, so `rowsFrom` called the card's version and every job's waiting line came back empty — silently, and only in the test. Renamed to `jobWaitingLine`; `summarise` was renamed for the same family of reason, because an *aliased* import is a name the harness does not have at all. A new guard compares every export here against `app.js`'s declarations so the next collision is loud.
5. `--ok` is not a token this repository has, and an invalid `var()` makes the whole declaration invalid without throwing. `.badge.ok` uses the accent pair, so the live dot uses `--accent-ink`.
6. His Arabic request was quoted in `app.html`. `CLAUDE.md` says comments stay English and `test_the_interface_stays_english` enforces it — his words belong in #793 and here, and that is where they are.

**And `Jobs` is plural where page names are singular.** Added to `PLURAL_PAGE_NAMES_ALLOWED` with the reason that list demands of anything added to it: he named it in writing, twice, and the rule's own docstring is the second argument — *"a plural among singulars reads as a LIST rather than a place, the panel has both, and the difference has to mean something."* This page is a list of 163 rows whose whole purpose is to be one; `Job` would name a page about a single job, which is the Run screen.

## Tests

**Thirteen mutations, thirteen caught.** Three survived first and each survival was informative:

- the harness answers every `/api/jobs` path with one payload whatever the query string says, so putting `active_only=true` back still drew four rows. The defect was in the **question** the panel asks, so the guard now reads the request — and separates the page's from the mini-player's, which still asks for the live job and correctly so;
- and twice, five guards drove `jobs.note_a_re_entry` directly while the calls from the two runners went unmeasured, so deleting either changed nothing any test could see. **This repository has been bitten by that shape twice before**; both call sites now have a guard that drives the real runner over a re-entered job.

Twelve new node tests for the page's rules, five DOM guards driving the shipped panel in a browser, and seven python guards for the re-entry line and its two call sites.

## Verification

Local gate on `fbc98dae`, and it is the **third** on this branch — each because the base genuinely changed, not out of caution:

| | |
|---|---|
| `ruff check scrapex/` — what CI lints | 0 |
| `eslint extension/` | 0 |
| `node --test extension/tests/*.test.mjs` | 0 |
| full suite, template mode | *running as this is opened* |
| full suite, `SCRAPEX_FULL_MIGRATIONS=1` | *running as this is opened* |

The two previous gates were **fully green** — `gate28` on the pre-rebase tree and `gate29` after the rebase onto `17ad3ea1`, both with identical BEFORE/AFTER fingerprints. This run measures the same change on top of `c30b098c`, and the result will be posted here rather than assumed to transfer.

**The last rebase was a real conflict, not a mechanical one, so the proof is different.** #819 inserted its restore card into `extension/app.js` at exactly the point this branch inserts the Jobs page — both immediately above `loadDatabase`. They conflicted by adjacency and not by meaning: neither touches a line of the other, so the resolution keeps both, in arrival order (162 lines from `main`, then 203 from here). A content-identity hash cannot hold across that, so the changed **lines** were compared with position and context stripped:

```
git diff -U0 17ad3ea1...445ed4da | grep '^[+-]' | sort   →  1,248 lines
git diff -U0 origin/main...HEAD  | grep '^[+-]' | sort   →  1,248 lines
diff of the two                                          →  no difference
```

So the resolution moved this contribution and changed none of it. `eslint` and the Jobs, rail, restore and snapshot DOM guards were then run together — 13 tests, all passing — to prove the two blocks coexist rather than merely parse.

**No engine contract moves**: no route, no migration, no capability. The extension gains a page, which `R-35` counts as user-visible — so if a bump is owed on the extension side, it is owed here and this is the place to say so.


muhammadbayoumi:
**CI is green on `fbc98dae` — 14 of 14. And my local gate flagged one failure on the same tree, which is not this change.**

```
gate30, TREE=a554a34b, DIRTY=0
  ruff check scrapex/   0
  eslint extension/     0
  node --test           0
  full, template mode   1 FAILED
    tests/test_the_engine_survives_being_killed.py::
      test_a_killed_engine_does_not_leave_a_job_claiming_to_run
```

**It is `OP-19`, the load-dependent race, and it is measured rather than asserted.** Re-run three times on the identical tree — `git write-tree` = `a554a34b`, nothing dirty — it passed three times out of three. `jobs.reclaim_orphaned_jobs` records the mechanism in terms: `Engine.start()` waits for `/api/health`, which the HTTP thread answers the moment the server binds, while the reclaim runs on the WORKER thread after it connects. Which of the two wins is a coin toss, and the note says it fails three runs in four on a loaded Windows machine and passes reliably on the Linux runner. **CI's `test` job passed it on this commit**, and this machine has been running a gate and a Playwright suite side by side all morning.

**The two earlier gates on this branch were fully green**, both with identical BEFORE/AFTER fingerprints: `gate28` on the pre-rebase tree, and `gate29` after the rebase onto `17ad3ea1`. This third run was made necessary by #819 landing in the same file, not by anything failing.

Recorded here rather than quietly re-run, because a cut or flaky run is not a green one — and because the honest report of a red is the red, plus what was measured about it.

The `SCRAPEX_FULL_MIGRATIONS=1` phase is still running; if it reports anything other than the same test, it will be posted here.

muhammadbayoumi:
## The review panel this change had not had, and what it returned

`b029e84`. **Green is not mergeable**, so before any merge this got a panel over all four dimensions — architecture, code quality, tests, performance — followed by an adversarial pass on what it found. It returned **three must fix and eight should fix**. The 14 green checks saw none of them, and the reason is structural: CI cannot see an unwired button, a CSS class nobody defined, or a guard that fabricates the payload it asserts on.

Every finding below was re-verified against the actual `file:line` before it was accepted. One was rejected on that pass and is named at the end.

### must fix

**`#jobs-reload` was drawn and wired to nothing.** The string appeared **exactly once in the repository** — `app.html:59`, the markup that draws it. Unreachable by all three wiring patterns this panel uses: no id lookup, no `data-*` delegation (the button carries no `data-` attribute), and its `ghost` class is styling only, never a selector. With no poll on the page it was also the *only* refresh, so a mini-player repolling every 1.5s sat on top of a list frozen at the moment the view was entered. Wired, with a guard that presses it and counts the second request.

**Failed jobs drew grey.** `statusTone("failed")` returned `err` — a **message** tone in this panel, not a badge one. The kit defines `.badge.ok`, `.badge.off`, `.badge.danger` and nothing else; `.err` exists at `components.css:75` but sets only a colour, 550 lines before `.chip, .badge` sets `color: var(--muted)` at equal specificity, so the cascade won. 28 of his 163 jobs rendered identically to the 23 cancelled ones. Now `danger`, and the guard asserts the whole range rather than the one status: every value `statusTone` can return must be a variant the kit defines.

**`jobWaitingLine` read a payload no producer emits.** It tested `queued_behind.job_ref`, `current_source_key` and `source_keys`. `_queued_behind` returns exactly `{position, capacity, running_count, running, starting_now}`. So the branch was **dead against the real engine**, every queued job fell through to "waiting for a worker" — losing the position, the capacity and the names of the holders, all of which the payload carries — and `starting_now` was ignored entirely, so a job the engine says is *starting* was told it was waiting, against that function's own comment: *"The panel must not call that 'queued'."*

Three things make it worse and all three were already in the repo: `renderQueue` reads the shape correctly 1,300 lines above in the same file; `test_panel_dom.py` used the real shape in this very PR; and this PR's own lesson 1 is *"an assumed payload is the same defect as an assumed file path"* — the same defect, in a second field. **The guard passed because it invented the shape.** It now uses the producer's, cited to it.

### should fix

- A **409 was written and erased** by the `loadJobs` the same handler called one line later — a caught error undone by its own reload. `keepNotice` now carries it, and the list still reloads, because a 409 means the row on screen is the stale one.
- That reload also **shut the row the button was pressed in**, discarded its fetched log and dropped focus to `<body>` on a list of 173 — undoing the interaction the page was designed around. Open rows, their logs and the focused control now survive the redraw; a row caught mid-fetch re-reads rather than restoring its placeholder for ever.
- **Cancel asked for confirmation on Run and not here** — on the page that puts Cancel beside Resume, for a job where cancel is immediately terminal. One `confirmedControl` now holds that question, with both flows as callers.
- The **"this one is moving" dot was drawn from "holds a worker"**, so it lit on the `preparing` job that sat blocked for 33 minutes and announced `running` to a screen reader — the same wrong discriminator `ADOPTION_ORDER` is built to reject, thirty lines above.
- **`fetch.basis` was dropped**, so an estimated denominator read as a count. It wears a `~` now, which is the panel's existing convention.
- **Log lines printed `logged_at` raw UTC** under a header in his own zone.
- **`note_a_re_entry`'s "READ BEFORE THE RESET" contract does not exist.** `_as_job` returns a `dict(row)` snapshot and `_update` writes the row, so the order cannot matter — and the test named for it proved the opposite by resetting first and still expecting 620. The code was right; three files described it wrongly. Restated as the property that actually holds, and the test renamed to pin it.
- Three added comments said untrue things, one contradicting another in this PR about the same code.

### tests

**No fixture here held a failed job** — which is why the grey badge survived. So `error_summary`'s paragraph and the badge were rendered by no test at all. `HIS_JOBS` gains a failed, a cancelled (the comparison that makes the failed one mean something) and a queued job carrying the real `queued_behind`; the counts move with them, which is the point of asserting counts. `openJobLog` had **zero** assertions despite two tests already running it — it now has its lines, its empty case and its failure. The five Jobs DOM guards asserted no `js_errors`, which is exactly what would have named the `rowsFrom is not defined` throw this PR already paid for once.

**The collision guard written here was the weaker of two.** It saw only exports, only `function` declarations, and only `app.js` — blind to `SETTLED`, `HELD`, `ADOPTION_ORDER`, `counted`, to `const` declarations (including this PR's own `JOBS_LIMIT`), and to twelve of the thirteen files the harness flattens. `test_panel_wiring.py` already holds that rule properly and was simply missing `jobsview.js` and `app.js`. Both added there; the weaker copy removed. One rule, one home.

And the **job vocabulary is three hand-kept copies** with nothing tying them to `vocab.py` — three guards added where the panel/engine seam is already crossed.

### One finding rejected on the adversarial pass

The Arabic quoted in `jobsview.js:3-4` was raised as breaking *"comments stay English"*. It is not this PR's: `app.js` on `main` carries Arabic on 11 lines already, unchanged here, and `test_the_interface_stays_english` scans HTML pages only. Long-standing practice, not a new violation — so it stands, and singling out these two lines would have been inconsistent.

### What I could not verify, stated plainly

**This machine has no toolchain** — `python` is a zero-byte Store alias that exits 9009, and there is no `node`, `ruff` or `pytest`. So **nothing here was executed**: no lint, no unit run, no mutation. What was done instead is that each new guard's own logic was replicated in the shell against the real files — the three vocabulary guards, and the collision guard over the extended module list (0 clashes, so adding `app.js` stays green). **CI is the verifier for this commit**, and a red is a red.

Widening that collision regex to `(?:async\s+)?` would go red on a **pre-existing** defect: `drive.js` and `sheets.js` both declare top-level `async function ask` and `async function refuse`, and the harness concatenates sheets after drive — so every Drive error path in the DOM suite raises a `SheetsError`. Filed as #845 rather than dragged in here.

Also filed, all *optional* and none fixed in this PR: #846 (nothing checks a drawn control is reachable), #847 (the bound notice), #848 (a cached log failure never retries), #849 (`directoryjob`'s silent reset), #850 (four smaller findings).

muhammadbayoumi:
## Merge gate: NOT CLEAN, and the branch is red. Not merging.

Four-dimension panel plus an adversary, run from the merge session. **The panel read `fbc98dae`; the head is `b029e845`** — the commit "The Reload that did nothing, the failed jobs drawn grey, and a payload no producer emits" landed after the panel started and answered six of its twelve findings outright. Everything below is verified against the **actual head**.

### First: `gh pr checks 821` is red

```
test                 fail   (both matrix runs)
migration-authority  fail   (both matrix runs)
```

The handover recorded this branch as "CLEAN, 14 of 14, waiting only for the merge". That was true of `fbc98dae`. It is not true now, and three of the four panellists repeated the stale green in their clean lists without checking.

**Two causes, both in this change's own new code:**

1. **`tests/test_panel_dom.py:622`** — `test_a_refusal_survives_the_reload_the_same_press_triggers` calls `open_panel(jobs=HIS_JOBS, control_status=409)`. `open_panel` forwards `**stub_kwargs` to `harness.stub` (`tests/test_panel_dom.py:133-137`), and `stub`'s signature at `tools/panel_harness.py:137-148` has forty keyword parameters and **no `control_status`**. `git grep control_status` over the head returns exactly one hit — the call site. So the guard written to protect this change's own "no silent failures" fix **errors instead of running**.

2. **`extension/app.js:6351-6353` vs `extension/app.js:6528`** — `loadJobs` recovers what to refocus by reading `document.activeElement` at the top of the redraw. Its only caller on the press path, `pressJobControl`, sets `button.disabled = true` at `:6528` *before* awaiting the POST. Disabling the focused element blurs it, so the activeElement read finds nothing to restore. The focus restoration cannot work as written.

### must fix — the page states a number in the wrong unit, by two orders of magnitude

`extension/jobsview.js:158-160` falls back to `progress.unit || "source(s)"`, and `_job_view` attaches a `unit` only for `organization_enrichment` (`scrapex/webui/app.py:4375-4380`). The two kinds this warehouse actually runs count something else: `scrapex/profilejob.py:29` says "progress is counted in PAGES" and writes `progress_total = wanted * 2` at `:295`; `scrapex/datasetjob.py:263-264` writes page **pairs**.

Executed against the shipped head module, fed the exact payload `_job_view` + `_fetch_progress` emit when `counters == {}`:

```
dataset_interpret  ->  "300 of 909 source(s)"
profile_crawl      ->  "620 of 938 source(s)"
```

There are **12 registered sources**. The row is wrong by two orders of magnitude in the one unit the owner can check himself.

**Why nothing caught it, which is the worse half.** Both suites feed these kinds a payload no producer can emit: `tests/test_panel_dom.py:401/408/416` give `profile_crawl` and `dataset_interpret` a `"counters": {}` *and* a hand-written `"fetch": {"requests": 620, "expected": 938}` in the same dict — internally contradictory, since `_fetch_progress` derives `fetch` **from** `counters`. `extension/tests/jobsview.test.mjs:21-32` does the same, and `:135-136` / `:157-158` **assert the `source(s)` fallback as correct**. `fetch` is empty for both kinds, verified three ways: `record_source_fetch`'s six callers are `capture.py:148`, `directoryjob.py:513`/`:602`, `jobs.py:375`/`:684`/`:830` — neither runner among them; `_seed_source_expectations` has one caller, `jobs.py:940`, inside the price-crawl runner; and neither `profilejob.py` nor `datasetjob.py` imports `capture` at all.

This is the same defect the head commit just fixed for `queued_behind` — an assumed payload — left standing in a second field.

**Suggested fix:** give `_job_view`'s `progress` block a `unit` per kind the way it already does for `organization_enrichment` — `"page(s)"` for `profile_crawl`, `"page pair(s)"` for `dataset_interpret`. ~4 lines and one route test. Then correct both fixtures to the empty-`fetch` shape the runners actually produce, or the guards keep certifying a payload the engine cannot send.

### should fix

- **`scrapex/jobs.py:1195` tells him a resume was a restart.** The text is "re-entered after a restart, so this pass starts again from the beginning", and its only trigger is `job.get("started_at")` (`:1190`). `set_control(RESUME)` sets status QUEUED and touches neither `started_at` nor `progress_done` (`jobs.py:189-190, 202-208`), so a pause-then-resume fires a line that names a cause it cannot know. **This PR makes the wrong path the common one**: `controlsFor` (`jobsview.js:228`) now draws Resume on every paused row of a 173-row list. The PR's own `jobWaitingLine` sets the opposite standard three files away. Keep the number, drop the causal half.
- **The page's headline number, its bar and its label are asserted by nothing.** `git grep` over the head for `job-bar`, `aria-valuenow` and `of 938` across `tests/` and `extension/tests/` hits only the pure-function assertions in `jobsview.test.mjs`. No DOM guard reads a rendered row's progress text, its `.job-bar`, or its `.job-label`. Extend the existing guard at `tests/test_panel_dom.py:476`.

### optional — filed, not fixed here

- The mini-player asks for the 5 **newest** active jobs (`extension/app.js:4870`) and the running one is the oldest.
- A log that fails to load once stays unreadable for the life of the page (`extension/app.js:6481-6483` with `:6513-6515`) — track load state on the element rather than inferring it from text content.

### Killed by the adversary

Ten of the panel's twelve findings, eight of them because the head commit had already fixed them — including the unwired Reload button (all four dimensions raised it; `git grep jobs-reload` at head returns four hits, not one), the notice erased one line after it is written, the missing Cancel confirmation, the hand-copied status vocabulary, the shared-scope guard, and `jobWaitingLine`'s fictional payload. The eleventh killed item is the claim that CI is green.

**Back to the author.** The merge session is not pushing behaviour changes into this branch.

muhammadbayoumi:
## Answering the merge gate, point by point

`d028fd99`, rebased onto `main`. Every finding it filed against `b029e845` is answered below, in its own order, with what was measured rather than what was intended. **Two of them were already answered before this push** — by `eb4cf16d`, which landed after the gate read the head — and I am naming which, because the gate was right to say a stale green is not a green.

### The red it found is gone, and it was two real defects

Both of the gate's `test` / `migration-authority` causes were fixed by `eb4cf16d` — `stub`'s missing `control_status` parameter and the `activeElement` read that could not survive `button.disabled = true`. `gh pr checks 821` on that commit: **14 of 14 pass**. The gate's own words hold, though: that is a green for `eb4cf16d`, not for this push, and the checks on `d028fd99` are the only ones that count.

---

### must fix — "620 of 938 source(s)" against 12 registered sources

**Confirmed on his live engine before anything was written.** `GET /api/jobs`, every `profile_crawl` and `dataset_interpret` row in his warehouse:

```
progress  {done: 620, total: 938}          <- no unit
fetch     {requests: 0, expected: null}    <- empty
```

`fetch` is empty by construction, exactly as the gate traced it: neither runner is among `record_source_fetch`'s callers and neither imports `capture`. So `progress` is the only pair a reader has, it named nothing, and the fallback said `source(s)`.

**Fixed on the wire, not in the panel.** `_job_view`'s `progress` block now declares the unit per kind — `page(s)` for a sweep, `page pair(s)` for an interpretation — the way it already did for `organization_enrichment` alone. The runner that wrote the number is the only thing that knows what it counted, and a second guess in the reader is how one kind came to be the only one that said so.

**The kinds are read from the modules that define them.** `PROGRESS_UNITS` keys off `profilejob.JOB_KIND` and `datasetjob.JOB_KIND`; `test_the_two_collecting_kinds_are_named_once` caught the re-typed literals within the hour, and it reads the file as TEXT, so quoting one in a **comment** trips it too. That cost a round and the comments now name none.

**Both fixture sets corrected**, which is the half the gate called worse. They carried `fetch: {requests: 620, expected: 938}` beside an empty `counters` in one dict — contradictory, since `_fetch_progress` derives `fetch` **from** `counters` — so two assertions certified the `source(s)` fallback as correct. They now carry the payload the runners actually produce. **A price-crawl fixture was added rather than the `fetch` branch dropped**: that kind does fill `requests`/`expected`, its total is an estimate, and it asserts `"1,030 of ~2,461 requests"` so the tilde convention stays measured too.

#### And the honest fixture found the same defect on a second surface

With `fetch` empty as the runners leave it, `miniProgress` drew **"starting…" and an indeterminate bar for a sweep at 620 of 938** — it read `fetch` alone. That is the "0% for eighteen minutes" complaint one kind over, invisible for exactly the reason above. Both surfaces now answer from `progressLine` / `progressFraction`, so a job cannot read 620/938 on one screen and "starting…" on the other. The mini-player keeps its own shape — a percentage, a bar that may be indeterminate — and loses its own arithmetic.

#### And then the fix's own window, which is the same defect one poll earlier

Writing the honest fixture exposed it. `create_job` seeds `progress_total` with `len(source_keys)` (`jobs.py:132`) — a true statement about a job nobody has picked up: **0 of 1 source**. The runner replaces it later, and not at the same moment in the two kinds: `profilejob:293-295` writes `wanted * 2` in the same UPDATE that sets PREPARING, while `datasetjob:264` writes nothing until its first page pair closes, and **never** for an interpretation that finds no pairs.

So declaring the unit unconditionally labelled the *seed* with the runner's word: a queued sweep read **"0 of 1 page(s)"** for a job about to fetch 938. Right where the number is the runner's, wrong in the window the fix itself opened.

The unit is now named only over a number a runner has written; a pair still holding the seed keeps the reader's own `source(s)`, which is what that pair means at that moment and what the page said before any of this. The discriminator is the seed rather than `started_at`, because `started_at` does not separate the two kinds — `datasetjob:224` sets it in an UPDATE that leaves `progress_total` alone, so an interpretation spends its whole PREPARING phase started and unclaimed.

---

### should fix — the line named a cause it cannot know

Accepted as stated. `set_control(RESUME)` sets the status to QUEUED and touches neither `started_at` nor `progress_done`, so "re-entered after a restart" fired on a pause the owner ended himself — and **this PR makes that the common path**, since `controlsFor` draws Resume on every paused row of a 173-row list. `jobWaitingLine` three files away already refuses to name a job it cannot identify, for the same reason.

The number stays, the blame goes: *"starting again from the beginning: this is not this job's first pass, and the last one had reached 300 page pair(s). Every pair is read from disk again and the site is asked for nothing, so this costs time and no requests."*

### should fix — nothing read a rendered row

Also accepted, and reproduced: `git grep` for `job-bar`, `aria-valuenow` and a progress string hit only the pure functions. `test_a_rendered_row_states_its_number_in_the_unit_the_job_counted` now reads the row — its label, `"620 of 938 page(s)"`, the interpret row's `"300 of 909 page pair(s)"`, the queued row's `"0 of 1 source(s)"`, and `.job-bar`'s own `aria-valuenow == "66"`, which is issue #725's omission on the crawl bar where a screen reader was told there was a bar and never what it read. A route test asserts the unit per kind, that `fetch` is empty for both, and that a price crawl declares none — the facts that make the unit load-bearing rather than decorative.

---

### Seven mutations, seven caught

| | mutation | caught by |
|---|---|---|
| M1 | the route stops declaring the unit | the route test |
| M2 | the panel ignores the declared unit | the node guard |
| M3 | the mini-player reads `fetch` alone again | the DOM guard |
| M4 | the re-entry line blames a restart again | three suites |
| M5 | the route declares the unit over the seed again | the route test's seed block |
| M6 | `claimed` hardwired to true | same |
| M7 | the reader guesses `page(s)` for an unlabelled pair | the DOM guard |

### One more payload no producer can emit, corrected

The failed `dataset_interpret` fixture carried `fetch: {requests: 41}`. That kind's `fetch` is empty by construction — the same trace as the must-fix — so it is now empty, with the reason beside it. Its `progress {0, 1}` is honest as it stands: the seed, unclaimed, because that run read no pairs. The `directory_crawl` row's 3,138 requests stay: `directoryjob` **is** one of `record_source_fetch`'s callers.

### The two optional findings

Both are already issues and neither is touched here, per *optional becomes an issue, never a fix in the same PR*: the mini-player's five-newest-active window is #850, and the log that stays unreadable after one failed load is #848.

### Verification, and what it does not cover

Run on this exact tree, and the fingerprint held at both ends -- `HEAD d028fd99`, clean, `TREE 317367e2`:

```
SCRAPEX_FULL_MIGRATIONS=1 pytest      3,955 passed, 11 skipped, 0 failed
ruff check scrapex/                   clean
eslint extension ... (9.39.0)         clean
node --test extension/tests/*.mjs     476 passed
node --test apps_script/tests/*.mjs    13 passed
```

The branch is also rebased onto `main` -- it was one commit behind #830, which is a CI change and touches nothing here, but a green earned against a workflow that is no longer the one running is the stale green this gate already caught once.`

`gh pr checks` on ``d028fd99`` is the gate that counts, and a red is a red.

**Not merging this myself.** Back to a merge session.

muhammadbayoumi:
### CI on `d028fd99`: 10 of 10 pass — and that is four checks fewer than last time

```
contract-parity  pass (x2)   lint  pass (x2)   scope  pass (x2)
test             pass (x2)   migration-authority  pass (x2)
```

**`eb4cf16d` had 14.** The four that are gone are the CodeQL family — `Analyze (actions)`, `Analyze (javascript-typescript)`, `Analyze (python)` and the `CodeQL` aggregate. They are absent from this head rather than pending: `repos/.../commits/d028fd99/check-runs` returns exactly the ten above.

That is **#864**, already filed and not caused by anything in this branch: updating a PR branch drops CodeQL from the head that gets merged, and the shorter list still reads as all-green. Naming it here because this PR's own history is the case for doing so — a merge session read a green that belonged to an earlier commit once already, and "10 of 10" is the same trap with the count moved.

So: every check that ran on this head passed, and CodeQL has not analysed it.

muhammadbayoumi:
### Both gates landed after this was written — `1cc881dd` is green on each

**CI**, and four rows is the whole set now that #851 has landed:

```
test  pass    lint  pass    scope  pass    contract-parity  pass
```

**The full local suite**, on the rebased tree, with the fingerprint identical at both ends — `HEAD 1cc881dd`, clean, `TREE 8934a9ee`:

```
SCRAPEX_FULL_MIGRATIONS=1 python -m pytest -q     3,961 passed, 11 skipped, 0 failed
```

Six more tests than the same suite reported one rebase earlier (3,955), which is #832's `granted-scopes` guards and the migration tests arriving with it — not this branch.

So "the full local gate was still running at hand-off" above is now answered: it passed, and the command it ran is the one CI's `test` job runs. **What #821 still waits on is a merge session, and nothing else.**

The CodeQL caveat in the body stands — those checks are absent from a force-pushed head (#864), so the four rows above are every check that ran, and CodeQL has not analysed this commit.

muhammadbayoumi:
**Two commits on this branch existed only inside a temporary directory, and have been preserved.** Not a change to this PR — flagging it because the directory is cleanable.

```
eb4cf16  2026-09-10  Restore focus where the focus still exists, and ask the harness for a knob it has
b029e84  2026-09-10  The Reload that did nothing, the failed jobs drawn grey, and a payload no producer emits
```

They were on **no remote branch** (`git branch -r --contains eb4cf16` → empty), checked out only under `%TEMP%/claude/…-github-handover-825-65f7d1/…/scratchpad/review821`. They are now at `refs/heads/rescue/821-work-that-lived-only-in-temp`.

**This PR is untouched** — still `OPEN` on `1cc881dd2`, the same head it published before. The rescue went to a new ref, with no force and no replacement.

**The decision is still yours.** Those commits do not sit on top of `1cc881dd`; the two lines diverge at `c30b098` (2026-09-09). So they may supersede what is published here, or be a dead end — this session did not work on #821 and is not guessing. Delete the rescue branch once you have ruled; it exists only so a cleaned temp directory cannot make the choice for you.

Context: #937.

muhammadbayoumi:
## The merge gate this branch had not had since the rebase, and what it returned

Rebased onto `main` first — 0 behind, and all six patch-ids identical before and after, so
the base moved and the content did not. That rebase also restored a gate this PR had lost:
the head now carries **8 of 8 checks including CodeQL**, where `1cc881dd` had only the four
that survive a force-push (#864).

Then a four-dimension panel over `38fcf4dc`, one reviewer per dimension, followed by **one
adversary per surviving finding** — the 601–1000 band, 923 changed lines outside `tests/`.
Eight agents, the ceiling for one PR.

| dimension | returned |
|---|---|
| architecture | 0 must · 3 should |
| code quality | 1 must · 1 should · 1 optional |
| tests | 1 must · 3 should · 1 optional |
| performance | **CLEAN** — 0 must · 0 should · 1 optional |

**After the adversary pass: 2 must fix, 4 should fix.** One finding was demoted to
*consider* and one to *optional*; four sub-claims were killed outright. All six are fixed
in this push.

### What the adversaries killed, and it matters

- **"the pending intent is cleared"** on a stale Resume — unreachable in the scenario as
  described; it needs a three-step chain, not two.
- **double dispatch / durable corruption** from that same flip — `_next_startable` gates on
  `_running` membership, not status, so no job starts twice, and three of the four runners
  re-assert RUNNING at their next boundary.
- **"0% he watched for 18 minutes"** as the `miniProgress` consequence — that case is 1,030
  requests against no denominator and is correctly indeterminate. The real regression is a
  start-up cosmetic, and the finding stands on the missing assertions, not on that story.
- **the enrichment page's rendered word** — byte-identical to `main`, which is why that
  finding dropped to *consider*. It is now moot anyway: the fix makes the wire carry
  `organizations` unconditionally again.

And one the adversary got wrong, corrected here: it reported that the harness cannot count
GETs and that the log guard was therefore *unwritable*. `tools/panel_harness.py:594` pushes
**every** path into `window.__calls`, GET included. The real obstacle was different — the
mini-player repolls the running job's log every 1.5s, so the count has to be filtered to
the row's own `job_ref`, which is why the guard now drives the **paused** row.

### Performance, because it was asked with counts

`GET /api/jobs?limit=200` issues **4 SQL statements at every N** — measured at limit 1, 5,
20, 171, 200 and 2000 — with `_job_view`, `_queued_behind` and `_fetch_progress` executing
zero SQL each. `EXPLAIN QUERY PLAN` shows no temp b-tree for the ORDER BY, so the LIMIT
terminates early. The 1.5s poll is byte-for-byte unchanged apart from the optional `unit`
string, ≤70 bytes on a 3,076-byte body.

## Every fix broken on purpose, and all seven mutations caught

| mutation | caught by |
|---|---|
| `directory_crawl` entry deleted | `test_a_listing_crawl_counts_cells_and_says_so` |
| the seed test restored for every kind | `test_an_enrichment_names_its_unit_even_for_one_organization` |
| the `bool(done)` half deleted | `test_a_runner_that_counted_as_many_as_it_had_sources_still_names_its_unit` |
| the RESUME status guard removed | `test_resume_is_refused_on_a_job_that_is_not_paused` |
| the log restore line deleted | `test_a_press_does_not_shut_the_row_it_was_pressed_in` — "1 request(s) before, 2 after" |
| `confirmedControl` → `return true;` | both cancel guards |
| `miniProgress` → always 100% indeterminate | `test_the_miniplayer_states_a_percentage_and_stops_claiming_one_it_lacks` |

Tree restored and baseline green after each.

## Recorded, not fixed here

- #1013 — the Activity bar says "starting…" while the mini-player says "620 of 938
  page(s)" **on the same screen**, which this branch's own docstring at `app.js:5195-5196`
  claims is impossible. Includes `activityCounters` still labelling pages "Sites done", and
  the engine's `jobs.html` discarding the `unit` the wire now carries.
- #1014 — every completed listing crawl reports `fetch.requests: 0` however many requests
  it made. Pre-existing, and the mechanism behind the unit defect fixed here.
- #1015 — the unit word spelled once per kind in two places (*optional*: four tests pin
  both ends, so drift is loud).

Also left alone deliberately: `extension/app.js:6912`'s placeholder branch is unreachable
by the suite (no test delays the logs route), and `#846`–`#850` are untouched.

PR #828, opened 2026-09-09
The Excel button that produced no file, and the 36,318 cells that stopped it

muhammadbayoumi:
## What he reported

> فى هذه الصفحة http://127.0.0.1:8000/source/contractor_enrichment لا استطيع تحميل الملف عبر زر export to excel

## The cause, at `file:line`

The button navigates to `/export/contractor_enrichment.xlsx` (`grid.js:3136`).
Measured against the running engine, that route answered **500**, not a file:

```
$ curl -D - -o /dev/null http://127.0.0.1:8000/export/contractor_enrichment.xlsx
HTTP/1.1 500 Internal Server Error
content-length: 21
```

Reproduced in-process against the live warehouse:

```
scrapex/publish.py:139   workbook_tables      -> 1 tab, 54 cols, 17,304 rows  (fine)
scrapex/localsheets.py:63 workbook_bytes      -> sheet.append(list(row))
ValueError: Cannot convert ['email_domain_candidate', 'source', 'website'] to Excel
```

`extract/service.py:1156` does `record = json.loads(row["data_json"])`, so a field
stored as a JSON **array** arrives as a Python `list`. `publish.py:91` passed the
value straight through, and openpyxl refuses a list as a cell value.

The page itself was never broken — the table renders because Tabulator coerces an
array with `String()`. Only the writer refused, which is why the page works and
the button does not.

## The census, on the whole warehouse

Every `generic_record` in `~/.scrapex/engine/scrapex-engine.db`: **36,318** list
cells, all elements strings, **zero** dicts anywhere. All of them in
`contractor_enrichment` (`dataset_definition_id = 3`):

| field | rows |
|---|---|
| `providers_checked` | 17,304 |
| `evidence_urls` | 17,304 |
| `contact_emails` | 768 |
| `contact_phones` | 750 |
| `iso_certifications` | 170 |
| `core_specialties` | 22 |

Six of the enrichment's declared output fields are `json`
(`enrichment/models.py:210-248`), and two of them are written on **every** row —
so every row of this dataset carried a list into openpyxl.

## Why the fix is in `dataset_workbook_tables` and not in the xlsx writer

`workbook_tables` documents itself as the one place that decides what a complete
export contains, and it has **three** consumers:

| consumer | before |
|---|---|
| `/export/{key}.xlsx` — `webui/app.py:1649` | 500, empty body (confirmed) |
| Apps Script funnel — `outputs.py:282` | nested arrays into a flat tab |
| Google / local sink — `publish.py:50` | `localsheets._write_sheet`, same refusal |

Fixing `localsheets` alone would have repaired the button and left the other two
wrong, and it would have needed the same edit twice inside one file. One place
now decides how a multi-value field reads.

## The three decisions, and whose they are

1. **Where** — `dataset_workbook_tables`. His call, recommended and taken.
2. **The separator** — `", "` and not `","`. His ruling. The screen shows `a,b,c`
   because that is Tabulator's default; a workbook is printed and mailed, so the
   cell is spaced to be read.
3. **An unexpected shape** — a named refusal, not a bare 500. His ruling.
   `UnexportableCell` names the dataset, the field and the value, and the route
   returns it as the `detail` he can read.

`UnexportableCell` is a **TypeError, deliberately not a ValueError**: both this
route and `outputs.apps_script_send` read a ValueError out of `workbook_tables`
as *"nothing ingested for this source yet"*, so a shape defect raised as one
would have told him to *crawl and ingest it first* — sending him to re-run a
crawl that was never the problem.

A null **inside** a list is dropped rather than written, because a cell reading
`None` is wrong data that looks like a value.

## R-77: no version bump

> nothing bumps per PR; the SHA is already exact — `docs/archive/RULINGS.md:3098`

The engine has no marketing version to move. `VERSION` stays `0.4.13`.

## Tests, and each one broken on purpose

The gap the defect fell through is exact: the existing
`test_a_dataset_exports_a_workbook_instead_of_refusing_one` stops at
`workbook_tables` and never calls `workbook_bytes` — openpyxl is the reader that
refuses. Every new test goes through the writer.

| test | what it holds |
|---|---|
| `test_the_enrichment_output_downloads_as_a_workbook` | presses the real route, opens the returned file. The fixture reproduces production's own three-provider cell, so the separator is proved on a real row |
| `test_a_multi_value_field_becomes_one_readable_cell` | two values, one value, empty list, and a null inside the list |
| `test_a_nested_object_names_its_field_instead_of_landing_as_wrong_data` | the refusal names the field, carries the value, and is not a ValueError |
| `test_an_unexportable_shape_names_the_field_instead_of_an_empty_500` | the route returns a readable `detail`, and the 404 keeps its own words |

Broken on purpose, one at a time, each guard watched failing:

| break | what failed |
|---|---|
| flattening removed | `ValueError: Cannot convert ['email_domain_candidate', 'source', 'website'] to Excel` — the owner's exact error, from the route test |
| `", "` → `","` | `'Grade 1,Grade 2' != 'Grade 1, Grade 2'` |
| null kept instead of dropped | `'Grade 4, None' != 'Grade 4'` |
| route's `except` branch removed | `UnexportableCell` uncaught — no readable detail |

## Not fixed here, on purpose

`_integration` (`webui/app.py:2845`) maps only `NotConfiguredError`, so an
`UnexportableCell` reaching the **funnel** button would still be a bare 500. No
data can produce that today (zero dicts in the warehouse) and widening the diff
to every integration action is a separate change. Filed as an issue rather than
carried here.

---

# Split, on his ruling — 2026-09-11

The AST guard `test_every_caller_of_the_export_path_names_the_type_it_can_raise`
(54 lines) **left this PR** for #882. Four review passes ran on this change;
three found something, and **every one of those findings was in that guard** —
never in the 96 lines of product code, which every pass confirmed correct.

The split is lossless: this branch's tree plus #882's equals the pre-split tree
exactly (`497e2612d12da794340d9a32c6ee0e758bdd025e` both ways). The five
behavioural tests added alongside the guard stay here, because they drive real
code rather than scanning it.

Rebased onto `e868486`; the diff hash is `4c69e012` before and after both
rebases, so the base moved twice and the content did not.

## Correction to "Not fixed here, on purpose" above

That paragraph is **false against this head** and describes commit `8357d5b`.
Commit `17306f5` added `except UnexportableCell: raise NotConfiguredError(...)`
at `outputs.py:295`, and `_integration` maps `NotConfiguredError` to a 400 at
`webui/app.py:2851`. Driven on this head: the funnel button answers
`400 {"detail":"contractor_enrichment.key_decision_makers nests a value that no
spreadsheet cell can carry: [{'name': 'Sara'}]."}` — not a bare 500, and it names
the field. The surviving invariant is narrower: `_integration` maps only
`NotConfiguredError`, so a *future* integration action calling the export path
must convert the refusal itself.

## What the fourth pass confirmed

No version bump is owed (R-77; no frozen vector moved, `CONTRACT_VERSION` stays
1). No injection path — the refusal reaches the panel as JSON and lands via
`textContent`. The panel does show the sentence: `grid.js:3136` is a
`window.location` navigation, so there is nothing there to swallow it. The new
per-cell loop costs **+132 ms** on the real 17,304 x 54 shape, 2.6% of the export
it feeds. `source_site` exports are provably untouched.


muhammadbayoumi:
**Verified against the live warehouse**, with the fixed code imported from this
worktree (asserted on `UnexportableCell` — a symbol added by this branch — so a
misdirected import could not fake it):

```
OK  17,304 rows x 54 cols  ->  3,146,149 bytes
  providers_checked: 'email_domain_candidate, source, website'
  evidence_urls:     'https://fared-est.com, https://muqawil.org/en/contractors/20008518/143'
```

The 3.1 MB workbook opens. Before this branch the same call raised
`ValueError: Cannot convert [...] to Excel` and no file existed.

His running engine still answers **500** on that route, because it is installed
editable against the main checkout — the fix reaches his panel when this merges.

muhammadbayoumi:
## Merge gate: NOT CLEAN. Green, but two should-fix survive. Not merging.

Four-dimension panel plus an adversary, run from the merge session against head `edfb475`. CI is genuinely green here (14 of 14, verified).

### should fix — openpyxl's own refusal still lands as the bodyless 500 this PR exists to abolish

`scrapex/webui/app.py:1654-1657`: `body = workbook_bytes(tabs)` sits on a **second** `try` whose only clause is `except RuntimeError` (openpyxl absent → 501). The new guards — `except UnexportableCell` at `:1643` and `except ValueError` at `:1649` — are on the *first* try, the one around `workbook_tables`, whose `finally` closes both connections at `:1651-1653`.

`IllegalCharacterError.__mro__` is `(IllegalCharacterError, Exception, BaseException, object)`: not a `ValueError`, not a `RuntimeError`, not a `TypeError`. Neither clause can see it.

**And it is reachable from crawled text, proved through the code rather than through a row.** `grep -rn 'ILLEGAL_CHARACTERS|IllegalCharacter|x00-|quotePrefix|32767' scrapex/ tests/` returns nothing; `enrichment/providers/website.py:432` only `.strip()`s; `extract/html_table.py:22,87-88` collapses `\s+`, which leaves **23 of openpyxl's illegal characters** (`\x00-\x08`, `\x0e-\x1b`) untouched. Ran the PR-head `dataset_workbook_tables` verbatim and fed each cell to openpyxl 3.1.5 the way `localsheets.py:62-63` does: `'Al\x07Rajhi'` and `['Al\x07Rajhi','b']` both pass the flatten and then raise `IllegalCharacterError`.

Not killed as pre-existing: it is pre-existing on the *product* export, but this PR is what turns `contractor_enrichment` from always-500 into a working export of **17,304 rows × 54 columns of provider-scraped free text**, so the exposure is created here.

**Two obvious fixes are wrong and the adversary killed both.** Moving `workbook_bytes` into the same `try` is actively dangerous: openpyxl raises a plain `ValueError` for a dict, a list-of-dict, a set and a tuple cell, and `app.py:1649` maps `ValueError` to a 404 "crawl + ingest it first" — the exact confident-wrong answer this PR made `UnexportableCell` a `TypeError` to prevent. Adding `except IllegalCharacterError` to `app.py` needs an openpyxl symbol in a module that deliberately has none. **What is left is the writer**: wrap the append loops in `localsheets.workbook_bytes` (`:62-63`) and `_write_sheet` (`:89-90`) in `except (ValueError, TypeError, IllegalCharacterError)` and re-raise as `publish.UnexportableCell` naming tab, row index and column header. That reverses the placement the PR body records as the owner's ruling, so **it is his call, not a reviewer's** — but the PR's contract ("the refusal names the field instead of an empty 500") holds only for the two shapes it pre-screens until that call is made.

### should fix — there is a FOURTH reader of the ValueError contract, and the PR's TypeError breaks a rule in CLAUDE.md

The PR's argument rests on a census of consumers: three of them, and two readers of a `ValueError` out of `workbook_tables` (`app.py:1649`, `outputs.py:282`). Both were audited, and `UnexportableCell` was made a `TypeError` precisely so neither would misreport a shape defect.

**The third reader was missed.** `scrapex/outputs.py:146-150` — `excel_export`, the panel's Excel Output button — loops over the owner's picked sources calling `publish_source` inside `except ValueError as exc:  # nothing to publish for that source` at `:149`. That clause is what implements the rule `CLAUDE.md` states flatly: *"One source failing never kills a run and is never swallowed."*

Verified at head, not argued:

- openpyxl 3.1.5 raises a plain `ValueError` for `{'g':1}`, `[{'n':'S'}]` and `['a','b']` — all three run. **Before** this PR that shape was caught at `outputs.py:149`, appended to `failures`, and the other sources in the pick were still written.
- The PR-head `UnexportableCell` is `issubclass(..., TypeError) True, ValueError False` — the PR's own test asserts it at `tests/test_a_dataset_is_a_table_like_any_other.py:676`. **After** this PR the same shape escapes `:149` uncaught.
- The loop then aborts on the first offending source, every later source in the pick is silently skipped, `conn.commit()` at `:151` and `_record(conn, 'excel_last', ...)` at `:158` never run so the Outputs page is told nothing, and `_integration` (`app.py:2836-2846`) maps only `NotConfiguredError` — so the panel receives **a bare 500 with no body**. The same symptom this PR exists to abolish, on a different button.
- The existing guard for the rule, `tests/test_outputs.py:163-166` (`test_a_source_with_no_data_is_reported_not_silently_skipped`, asserting "the good source still went out"), stays green because it only ever exercises the `ValueError` shape.
- The route accepts a dataset key: `_source_keys` at `app.py:4406-4417` does no validation.

This is a regression in run-continuation, not merely an unimproved message — which is what separates it from the funnel gap correctly filed as #829.

**Fix:** `except (ValueError, UnexportableCell) as exc:` at `outputs.py:149` — the message already names the dataset, the field and the value — plus one test asserting the good source still goes out when a bad one is in the same pick. If it is deferred instead, it must be added to **#829**'s scope, which as written names only `_integration` and the funnel and would not cover this.

### optional — filed, not fixed here

- **A crawled cell beginning with `=` is written as a live Excel formula.** `'=HYPERLINK("http://x"&A1,"c")'` reads back with `data_type='f'`; `'+1+1'`, `'-1+1'`, `'@SUM(A1)'` read back as `'s'` (the CSV folklore does not apply to a stored `.xlsx`). Not purely pre-existing: before the PR a list cell was refused outright, and after it `['=1+1','b']` joins at `publish.py:132` to `'=1+1, b'` and lands as `data_type='f'`. The fix (`quotePrefix`, or forcing `data_type='s'` in `localsheets`) changes the bytes of every workbook already in use, including the daily product export — his decision, its own PR.
- **The join is a new way past Excel's 32,767-character cell limit, and openpyxl truncates in silence.** `['u'*200]*300` builds a 60,598-character cell that writes with no exception at any layer and reads back at 32,767. `evidence_urls` is written on all 17,304 rows with no bound on element count. A visible `... +N more` marker serves better than a refusal.
- **The dataset `.xlsx` export has no row bound** where the product path is capped at 40,000 (`reports.py:1214-1215`); `dataset_table_payload(..., cap=None)` at `extract/service.py:928` means every row. Not caused by this change — the equally large `contractors` dataset already ran the same uncapped path — so it is a standalone issue, not a gate finding.

**Back to the author.** The merge session is not pushing behaviour changes into this branch.

muhammadbayoumi:
**Handover — the session that opened this is closing.**

Milestone: [Export to Excel: the button, and what the review found behind it](https://github.com/muhammadbayoumi/ScrapeX/milestone/31) · handover issue #870.

State: `MERGEABLE` / `CLEAN`, every check green on `d5f710ee`, **not approved for merge and not merged**.

Two review passes ran and both rounds of findings were fixed and pushed. **Pass 3 was still running when the session ended and its result was never seen** — it was attacking pass 2's own fixes, chiefly the new AST guard and the `cli.py` exemption. Do not read "three passes ran" as "three passes were clean": re-run pass 3 before treating this as converged.

Two claims from passes 1 and 2 were wrong and should not be inherited: `cli.py` never turned a sentence into a traceback (`cli.main` has an `except Exception` backstop), so the clause added on pass 1 was dead code and pass 2 reverted the file to main's version.

muhammadbayoumi:
**Correction to the handover comment above.** Pass 3 reported minutes after it was written, and it is **not clean**: one must fix and four should fix, filed as #875. Nothing is fixed; the branch stands at `d5f710ee`.

The must fix is in the AST guard pass 2 added — rename `workbook_tables` across `scrapex/` and it still passes while the blocks it inspects fall from 3 to 1.

Two findings correct **pass 2's own prose**: the `UnexportableCell` docstring here is wrong for the second time running. Pass 1 said the `TypeError` base prevents the misreport, pass 2 replaced that with "clause order does", pass 3 measured both — each is half of it. The accurate wording is in #875.

Three passes, three sets of findings, two of them errors in the previous pass's prose. Not converged — do not merge on "three passes ran".

Issue #656, opened 2026-09-05
HANDOVER: where the CI and testing audit stopped, what is measured, and how to resume it on another machine


muhammadbayoumi:
**Resume point for the CI/testing audit. Nothing in the repository was modified — no file, no commit, no push.** Every item is a recommendation awaiting the owner's decision.

Measured on `origin/main` **93b31e38** (2026-09-05). `main` has since moved (`c6269a2a` = CLAUDE.md only, then `68030ab4` = PR #609, which touched `scrapex/`, `extension/` and two test files). **The percentages below are portable; the absolute test count is one merge stale.**

---

## The baseline — MEASURED, reproducible

| | value |
|---|---|
| test files | 220 |
| tests collected | 3,684 |
| result | 3,674 passed · 10 skipped · **0 failed** |
| serial test time | **680.0s** (11.3 min) |
| with `SCRAPEX_FULL_MIGRATIONS=1` | **695.6s** (+15.7s, +2.3%), 17 skipped, 0 failed |
| browser suites (10 files) | **403.3s = 59.3% of time** for **308 tests = 8.4% of tests** |
| non-browser average | 82 ms/test |
| unconditional `wait_for_timeout` in browser suites | **448 calls asking 189.8s** |
| same suites, waits capped at 50ms | 401s → **233s (−42%)**, only **5 of 308** tests fail |
| `-m "extension or docs"` tier | 971 tests = **60.4% of suite time** |
| `-m docs` tier | 276 tests = **3.7%** (25.5s) |
| real `migrate()` after the R-84 squash | **12.4 ms** (template restore 4.6 ms) |
| `pytest -n 4` (24-core box) | **204s wall** for the whole suite, 0 new failures |
| `pytest -n 8` | 163s wall, **1 failure** (a fixed-budget race, see #649) |

### CI, 30 days, from the Actions API

| | value |
|---|---|
| CI runs | 617 (385 push · 232 pull_request) |
| total | **11,529 runner-minutes** |
| `migration-authority` | **6,307 min = 54.7%**, median **12.8 min** — the slowest job, so it sets PR latency |
| `test` | 4,848 min = 42.0%, median 9.2 min |
| lint · contract-parity · scope | 186 · 123 · 66 min |
| SHAs that ran CI twice (push **and** pull_request) | **231 of 378**; the PR copies alone = **4,438 min = 38%** |
| full-scope wall clock | median **13.2 min**, p90 14.5 |
[...]

PR #830, opened 2026-09-09
A suite that loses five of six tests at run time still reports green

### LINKED ISSUE #644: Five tests that prove a pre-squash warehouse still opens have never run: the fixture asks main for a directory the same commit deleted
Five tests that prove the owner's pre-squash warehouse still opens have **never run once**. They were added by `9a34a01f` (#318) — **the same commit that deleted `db/engine/migrations/`** — and their fixture asks `origin/main` for that directory.

## Measured, on `origin/main` 93b31e38

```
$ python -m pytest tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py -q -rs
ss.sss
SKIPPED [1] ...:98:  origin/main's pre-squash chain is not reachable in this checkout
SKIPPED [1] ...:114: origin/main's pre-squash chain is not reachable in this checkout
SKIPPED [1] ...:156: origin/main's pre-squash chain is not reachable in this checkout
SKIPPED [1] ...:179: origin/main's pre-squash chain is not reachable in this checkout
SKIPPED [1] ...:222: origin/main's pre-squash chain is not reachable in this checkout

$ git ls-tree origin/main:db/engine/
derived-from.json   schema.sql   squashed-from.json      <- no migrations/
```

Five of six tests skip. `addopts` carries `-q`, so the file reports **green**.

## Why the skip reason is wrong, and will never self-heal

`tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py:53-70` blames "a shallow clone or a detached CI checkout". It is neither. The directory is gone from `main` permanently, so the `git ls-tree` raises `CalledProcessError`, the fixture returns `None`, and every test that needs a real pre-squash database skips — **on every machine, in CI, forever**.

## The fix already exists in this repository, one line away

`tools/carry_a_warehouse_to_the_baseline.py:67` pins the right anchor:

```python
PRE_SQUASH = "9cbc6f0"          # #317, the commit before the squash
```

Verified: `9cbc6f06` is reachable and `git ls-tree 9cbc6f0:db/engine/` still lists `migrations`. The **sibling test written in the same change already does it correctly** — `tests/test_a_warehouse_below_the_baseline_can_be_carried_to_it.py:35,42,51` imports the tool, uses `tool.PRE_SQUASH`, and guards its skip on reachability.

So: change the fixture's ref from `origin/main` to `tool.PRE_SQUASH`, and make the skip conditional on that commit being absent rather than unconditional in practice.

## Why this is the most serious item in the audit

The subject is business-critical and specific: **the owner's real warehouse predates the squash.** Whether a pre-squash database still opens, upgrades, and refuses what it should is exactly what R-84 needed proving, and it is the one thing these tests were written for. They have never answered it.

It is also the failure mode this repository has guarded against more than once — a suite that skips and reports green. `.github/workflows/ci.yml` already carries a step for the browser suites ("Every browser-driven suite must RUN, not skip") built after 48 tests were silently skipped for months. **No equivalent guard covers this file**, which is a second, smaller item: a skip floor, or a rule that a skip whose cause is permanent must fail instead.

## Effort and risk

One-line ref change plus a corrected skip condition. **Effort S. Regression risk: none — it can only turn tests on.** The risk is entirely in the other direction: if the five tests fail once they run, that is a real finding about pre-squash warehouses and is exactly what they exist to tell us.

### LINKED ISSUE #674: main is red: the baseline refusal test selects every migration when the chain is two long

muhammadbayoumi:
Closes #644.

## What #644 asked for, and what is left of it

**Its headline item is already fixed, and the issue never learned.** #678 (`d7ecedc`, merged 2026-09-06) rewrote `_pre_squash_commit` to *derive* the era instead of reading `origin/main`, and made a reachable-but-unqualifying history raise rather than skip. It closed #674. Nothing pointed it at #644, so #644 has been open ever since describing a defect that no longer exists.

Verified on `c30b098` rather than assumed — the fixture's own derivation, replayed in git:

```
shipping baseline (declared_schema_version) = 17
git rev-list origin/main -- db/engine/schema.sql   -> 3 commits
9a34a01f^ : baseline=1  migrations=16  -> qualifies
```

and then run:

```
$ python -m pytest tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py
6 passed in 1.41s
```

**Its second item is untouched, and this is that.** #644 also asked for "a skip floor, or a rule that a skip whose cause is permanent must fail instead", noting that `ci.yml` already carries one for the browser suites and none for this file. That is still true.

## Why the existing floors could not have caught it

The three steps above this one count `--collect-only`. That works for a browser suite, which collects **nothing** when Chromium is absent. This file is the other shape: it collects **six every time** and loses five at run time, when `pre_squash_chain` returns `None` and every test that needs a real pre-squash database skips. A collect-only floor reads 6 in both worlds.

So this step runs the file and counts skips out of the JUnit XML, with a floor on the number that ran.

## Proved by mutation

| mutation | pytest itself | the guard |
|---|---|---|
| none — `c30b098` as it stands | `6 passed, 0 skipped` | **passes** |
| the history is not reachable (`rev-list` fails, the fixture returns `None`) | `1 passed, 5 skipped` — **exit 0, green** | **fails**, `skipped == 0` |
| the file is emptied (`-k` selects one) | `1 passed` — exit 0 | **fails**, `ran >= 6` |

The middle row is the point: it reproduces exactly the shape #644 describes — five tests silent, pytest green — and it is the guard, not pytest, that turns it red.

## Why a number and not "at least one"

The same reason `test_panel_dom.py` carries `-ge 40`. Five of six going quiet is the regression that already happened here; a per-file `-ge 1` would not have noticed it.

## Why it runs on every scope

Measured at 1.3s. The floors beside it are unconditional for the same reason, and a change that breaks the derivation is not necessarily a `full`-scope change.

## State

`.github/workflows/ci.yml` only — one added step, 34 lines, no engine, extension or test code touched.

Measured on this branch, on **Python 3.14.7** (CI pins 3.12 — the gap is #656's item 5, not addressed here):

- full suite: **1 failed, 3575 passed, 21 skipped in 219s**
- the 11 test files that read the workflows: **188 passed, 3 skipped, 1 failed**

**The one failure is pre-existing and unrelated**, isolated by stashing this diff and re-running on a clean tree: `tests/test_the_engine_knows_which_code_it_is_running.py:632` `test_no_repository_at_all_is_not_a_reason_to_stop_answering` — "the staleness check must not depend on git being present". Filed separately.

None of the 21 skips come from the file this step guards.

Browser suites collection-skip on this machine (no Chromium, no node), so their contribution to the counts above is absent rather than green. CI runs them.


muhammadbayoumi:
## Pre-merge review, round 1 — panel over four dimensions, then an adversary

Nine candidate findings. The adversary killed four outright and cut three down. **One survived as *should fix*, and `e6baf80` is it.**

| finding | panel | adversary |
|---|---|---|
| **A2** the step executes six real tests above the payload | should fix | **CONFIRMED** — fixed in `e6baf80` |
| C2 `assert` is disabled by `PYTHONOPTIMIZE` | optional | confirmed as mechanism, remote as risk → #833 |
| C1/T3 the failure message names one of three causes | should fix | partly → optional, cheaper fix → #834 |
| A1 unconditional on scopes that never run the file | should fix | partly → optional, moot after A2 → #836 |
| T2 the comment overclaims its lineage | should fix | partly → corrected in `e6baf80` |
| T1 the sibling carry-tool suite is uncovered | should fix | **killed** |
| T4 `ran >= 6` cannot see substitution | optional | **killed** |
| A3 the disease has three instances | optional | **killed** — two, both already defended |

Also surfaced and filed: #835.

### Three things the adversary established by running them, not reading

**The negative control the panel proposed is a no-op.** Stripping `found[0] < now` at the test file's line 103 gives `6 passed`, byte-identical to unmutated — `rev-list` returns three commits and the newest *is* the squash, so its parent is pre-squash either way. **A real negative control does work**: forcing the fixture to return `origin/main` gives `2 failed — DID NOT RAISE`, and it is structural, not an accident of chain length. Any post-squash era builds a database at or above the shipping baseline, so `"no upgrade path"` can never be raised. **The #678 class is covered — by pytest, not by the skip counter.**

**C2 is real.** The step's verbatim heredoc, fed a genuine 6-of-6-skipped XML: normal Python exits 1 and the step is red; under `PYTHONOPTIMIZE=1` it prints `6 tests, 6 skipped` and exits **0**. Nothing in the repo sets the variable, which is why it is #833 and not a blocker.

**Two sentences in this PR's own comment were wrong**, and `e6baf80` corrects them:

- *"cheap enough to run on every scope, like the floors above"* — the parity does not hold. Those floors are `--collect-only`; this one executes.
- *"That was its whole life before #678 … they never ran once"* — false. Measured: `9a34a01` (09-04) deleted `db/engine/migrations/` and five of six skipped for **25 commits**; `b50bc1e` (09-06) added `0018` and re-created the directory, after which the fixture resolved to the **squashed** chain and all six ran while proving nothing. **Two silent-failure eras, not one**, and the second was the one #678 found.

### What `e6baf80` changes

The step moves from before the gates to the end of the `test` job. Above the payload, a *product* failure in those six tests aborted the job and took ~11.8 minutes of unrelated answers with it — the failure `lint` was made its own job to avoid (`ci.yml:17-20`). Below the suite, that defect reddens the suite with everything already reported, and the step is left saying only what nothing else can: that the six ran. On `docs` and `extension` the suite step is skipped, so the guard still lands in about forty seconds. The move also removes the duplicate execution on `full`.

Re-verified after the move: green case `6 passed, 0 skipped` → guard passes; unreachable-history mutation `1 passed, 5 skipped` with **pytest exit 0** → guard exits 1.

muhammadbayoumi:
## Pre-merge review, round 2 — on round 1's fix

Ran against `e6baf80`, across the four dimensions, with the CI job for this branch as the evidence (`102604409479`).

### Two findings, both comment accuracy, both fixed in `b0ecc28`

Round 1's fix corrected two unmeasured sentences. It introduced two more, and this is the same defect one round later:

| claim | where it came from | measured |
|---|---|---|
| "~11.8 minutes of unrelated answers" | 0.92 × the `12m49s` at `ci.yml:383` — a figure predating this branch | suite step **638s**, the two gates above it **10s** |
| "the guard still lands in about forty seconds" on `docs`/`extension` | nothing | **unmeasurable** — none of the last 40 successful `ci.yml` runs took either scope. Removed rather than guessed. |

Replaced with what the same job does measure: **46s** to reach the suite step, **685s** to reach this one.

### One gap round 1 missed, now stated in the comment

**A red suite skips this step entirely.** GitHub stops the job at the first failing step, so after the move a failing suite means the guard never runs and nobody learns whether the six ran. That is not a hole — the job is already red, so **no green build reaches a merge without this having run and passed.** What the order costs is only how soon a *skip* is reported: 46s before the move, 685s after. The right trade, because a skip is a CI-integrity fault nobody iterates on, while a product failure is what someone is waiting on. Round 1 argued the move without noticing this consequence; the comment now carries it.

### Checked and clean

- **Step order and scope.** The guard is the last step in `test` and carries no `if:` — verified by parsing the YAML. Measured step conclusions on this branch's job: steps 10 and 11 `skipped` (scope is `full`), step 12 `success` 638s, step 13 `success` 1s. An unrecognised scope value falls through to `full` by `ci.yml:81-83`, and the guard runs on every value regardless.
- **The guard actually ran after the move.** CI log: `pre-squash suite: 6 tests, 0 skipped`.
- **Ordering hazards.** None found. `/tmp/pre-squash.xml` is written by this step alone; `.pytest_cache` from the suite changes nothing without `--lf`/`--ff`; the file monkeypatches the migration stream, so `tests/conftest.py`'s template never applies to it in either job, and the two runs are separate processes so the module-scoped fixture cannot leak.
- **The `lint` precedent the comment invokes** is real and says what the comment claims — `ci.yml:16-20`: "A style failure that aborts the suite hides every real failure behind it". The comment cites the principle without a line number, so it cannot rot.
- **Re-verified after the move**, twice: green case `6 passed, 0 skipped` → guard passes; unreachable-history mutation `1 passed, 5 skipped` with pytest **exit 0** → guard exits 1.

**No must fix. No should fix.** Round 1's two survivors are fixed; round 2's two are fixed. Everything else is filed: #833, #834, #835, #836.

muhammadbayoumi:
## Merge gate: **CLEAN**. Not merged — waiting on his word.

Four dimension reviewers plus an adversary, run from the merge session. **4 of 4 reviewers returned a report**: architecture (2 findings), code quality (3), tests (2), performance (0 — a clean-list-only report). Seven findings reached the adversary: 2 should fix, 5 optional, no must fix. **Both should fixes dropped to optional under attack, six findings were killed, nothing was missed.**

### The head, verified rather than repeated

`b0ecc289`, `gh pr checks` all passing. The last of the three commits is dated `02:04:58Z` and the branch has not moved since; `main` is still `c30b098`, the base the PR body measured against — so neither the head nor the green has expired, and no update is pending. The adversary read the change from `git archive b0ecc289 | tar -x` into a scratch directory and confirmed `import scrapex` resolved to *that extract*, not the editable install, so everything executed was this PR's code.

### The claim that mattered, and it holds

This change is itself a guard, so the question was whether it can fire. **It was executed, not reasoned about.** The adversary built the #644 world — an extract with no `.git`, so `_git("rev-list", "origin/main", ...)` at `tests/test_a_squashed_baseline_opens_what_it_should_and_refuses_the_rest.py:96` raises and `_pre_squash_commit` returns `None` at `:100`:

```
$ python -m pytest <that file> --junitxml=...
1 passed, 5 skipped in 0.20s          EXIT 0        <- pytest green on the exact defect

$ <ci.yml:398-405, run verbatim against that XML>
pre-squash suite: 6 tests, 5 skipped
AssertionError                        EXIT 1        <- the guard red on the same world
```

Every malformed-input path fails closed, the floor cannot be satisfied by another file or a stale run, and the duplication charge against the existing floors at `:253`, `:258`, `:272-273`, `:297`, `:309` was killed — they answer a different question (collection) from this one (run-time skips), which is the whole reason `--collect-only` could not have caught #644.

### Five optionals, filed rather than fixed here

Per the rule, an optional becomes an issue and never a fix in the same PR. Eleven optionals across #830 and #832 became seven issues, grouped only where two findings hold the same knowledge and would change for the same reason:

- **#854** — the step's teeth are two `assert`s (so `PYTHONOPTIMIZE` disarms it while it prints the reassuring line — demonstrated), `ran` is the collected count and the message misnames it, and the message diagnoses one of two skip causes. One issue, because all three are about how this step decides to go red.
- **#855** — the sibling file with the identical shape has no floor, and the obvious one-line fix goes *falsely red* at `tests/test_a_warehouse_below_the_baseline_can_be_carried_to_it.py:234`, so it needs the discovery-based form `ci.yml:221-225` already argues for.
- **#856** — nothing notices if this floor, or the three older ones, is deleted or lowered.

None blocks. The first is optional only because `grep -rn "PYTHONOPTIMIZE|python -O"` over the head returns zero hits and this diff creates none.

### Ordering, measured

`b0ecc289` is a **direct ancestor** of #851's head — the stack is linear, so **#830 holds the first slot and #851 cannot go before it**. After this squash-merges, updating #851 conflicts in one hunk of `ci.yml`; the resolution is mechanical (take the branch side entire) and is recorded on #851 along with the trap in it.

**Not merging without his word.** This is the first PR of the seven to pass the gate.

PR #832, opened 2026-09-09
A browser with no Google account met the panel wearing Chrome's error

muhammadbayoumi:
## What he sees today

Open the panel in a Chrome profile that is not signed into any Google account, and
the sign-in notice is filled with Chrome's own sentence — on a panel that asked
nobody anything. The person has pressed nothing; the first thing the product says
to them is a browser diagnostic.

Found while answering a different question: why a freshly downloaded copy showed
the owner's own account already signed in. That turned out to be his Google
account's standing OAuth grant, addressable because `manifest.json`'s `key` pins
one extension id — expected behaviour, no defect. This was found beside it.

## The chain

- `readTokenResult` knew three message shapes — `extension/identity.js:104-111`
  and `:126` — and everything else became `state: "failed"` carrying Chrome's raw
  `message` verbatim (`:133-134`).
- `failed` is one of the two states the panel shows **even on the silent check it
  runs on every open**: `interactive || ["timeout", "failed"].includes(result.state)`
  — `extension/app.js:2334`.
- `renderAccount` prints `tokenProblem.detail` into `#signin-problem` —
  `extension/app.js:2458`.

A profile with no Google account matches none of the three shapes, so it took that
road.

**And the code already said so.** Three lines above the print, `app.js:2452`:

> A silent check that found nobody is not a problem to report: it is the ordinary
> state of a machine nobody has signed in on yet.

The intent was written down and the gate let one case past it.

## Why not simply drop `failed` from that list

Because #152 put it there on purpose. Before it, the line read
`tokenProblem: interactive ? result : null` — silent on every non-interactive
check — and a silent check that timed out or failed left the panel blank with no
explanation, which is exactly how a broken startup looked during that
investigation.

So this **adds a branch and removes nothing**. A message Chrome has never sent
before still reaches `failed` and is still shown, and a test now pins that half
too, so a future narrowing cannot quietly buy the silence back.

No change to `app.js` was needed: the new state is simply not in that list, and
`interactive ||` short-circuits, so it is quiet on open and speaks on a press.

## Why it survived every test

The harness models "the machine nobody has signed in on" with `lastError` =
`"The user did not approve access."` — `tools/panel_harness.py:422-424` — which
maps to `declined` and is silent. The fixture was reproducing a **closed consent
window**, not an account-less profile, so the one message that reaches this path
was the one nothing ever sent.

The new tests pass it explicitly. Whether that default should change is a separate
decision with its own blast radius, and is not taken here.

## Silent is not secret

Someone who *pressed* the button is owed an answer, and Chrome's sentence is not
one — it names no next step. The new copy says what to do:

> Chrome has no Google account signed in, so there is nobody to sign in as. Add
> your account to Chrome, then try again.

## What was measured

Both guards were executed, and both were made to fail on purpose first.

**Unit — `node --test extension/tests/*.test.mjs`:** 461 pass, 0 fail.
Corrupting the new pattern alone (`/not signed in/` → `/not signed on/`) failed
exactly one test with `actual: 'failed'` vs `expected: 'no-chrome-account'`, and
left the sibling test — the one pinning that an unknown message still reaches
`failed` — passing. The two halves are independent.

**Panel — `pytest tests/test_signing_in_says_what_happened.py`:** 27 pass.
Against the same corruption, the guard printed the defect as the user meets it:

```
assert page.text_content("#signin-problem").strip() == ""
E   AssertionError: the silent check reported a browser with no account as a problem
E   assert 'The user is not signed in.' == ''
E     + The user is not signed in.
```

**Also run:** `pytest -q -m extension` — the suite the release workflow gates on
— reached 100% with no failures and no errors.

## Not fixed here

Three findings from the same investigation, each its own decision, each filed with
its evidence and its options rather than carried in this body:

- #837 — the panel harness models a fresh machine as a closed consent window, which
  is why this defect survived every test
- #838 — the privacy policy never says the panel checks the account on every open
- #839 — a fresh machine never registers the native host, so *Start engine* and
  *Start with Windows* stay dead


muhammadbayoumi:
## Pre-merge gate

CI is green on every check — `test`, `migration-authority`, `lint`, `scope`, `contract-parity`,
and CodeQL across actions / javascript-typescript / python.

Green is not mergeable here, so a panel ran over all four dimensions and an adversary was then
set on everything it found.

| dimension | reviewer | after the attack |
|---|---|---|
| architecture | clean | — |
| code quality | 2 × optional | both refuted → not an issue |
| tests | 1 × should-fix | refuted → not an issue |
| performance | clean | — |

**Nothing survived.** No *must fix*, no *should fix*, and the two *optional* findings did not
stand either. Details worth keeping:

- **The `/not signed in/i` breadth** was checked against Chromium's own
  `identity_constants.cc` (tags 140.0.7339.80 and 72.0.3626.80), not against recollection. It
  captures the account-less messages and no sibling that must stay `failed`.
- **A second, different account-less state exists** — a profile where `BrowserSignin` is
  disabled by enterprise policy — and it carries a different message needing different advice
  ("Add your account to Chrome" would be false there). Both the reviewer and the adversary
  concluded it wants its own branch, state and guards, i.e. not this diff. Not a regression:
  that door was already open and this change does not widen it.
- **The tests finding was a stale numeral**, this file's own header saying "four" refusals. It
  had been wrong since #145 — `timeout`, `cancelled` and `partial` all landed before this
  branch — so this diff made an already-wrong count one worse rather than turning a true line
  false. Pre-existing, and filed as #840.
- **The adversary added one thing the reviewers missed**, now recorded on #837: the bare
  `open_panel()` default is load-bearing as that file's `declined` fixture
  (`tests/test_signing_in_says_what_happened.py:191-204`), so changing it costs more than it
  appears to.

Merging is another session's call. This is the gate's result, not a merge request.

muhammadbayoumi:
## Re-checked against the rules ruled in #853

#853 is still open, so these are not yet in `CLAUDE.md` — but the ruling is dated, and two of
its rules bear on the gate report above. Both are answered here rather than left for the
merging session to redo.

**"CI green on the head it has right now — a recorded green expires."**
The head has not moved: `2e4821f7af3492cf66685013d8f789020cbe9a65`, the same commit the gate
ran on, with no pushes since. Re-read today, every check is green *on that exact head* —
`test`, `migration-authority`, `lint`, `scope`, `contract-parity`, and CodeQL across actions /
javascript-typescript / python. `main` has not moved either (last merge was #819, which this
branch is based on), and `mergeStateStatus` is `CLEAN`. So this is not the #821 situation: there
is no newer head for the recorded green to be stale against.

**"The report names each reviewer and its verdict; a report that cannot is a failed pass, not an
empty one."**
The panel above was six agents and **all six completed** — four reviewers (architecture, code
quality, tests, performance) and two adversaries, spawned only for the two dimensions that
returned findings. None was blocked, none errored, none returned empty. The verdicts are the
table above, per named dimension.

**"A local green ran no browser test, and `node --test` runs suites `pytest` never reaches."**
Both were run locally, not just in CI: `node --test` over `extension/tests/*.test.mjs` (461
pass) and `pytest` with a real Playwright chromium over
`tests/test_signing_in_says_what_happened.py` (27 pass), plus `pytest -q -m extension`. Each was
also made to fail on purpose against a corrupted pattern — details in the PR body.

Three other rules were checked and are not engaged: the diff is 79 lines, so the 1,000-line
split does not fire; no `VERSION` bump; no `normalize`, connector, source or `robots` change.

Still not merging — that is another session's call, and per #853 taking the PR
(`--add-assignee @me`) belongs to whichever session does.

muhammadbayoumi:
## Merge gate: **CLEAN**. Not merged — waiting on his word.

Four dimension reviewers plus an adversary, run from the merge session. **4 of 4 reviewers returned a report** — architecture, code quality, tests, performance — none empty, each with an explicit coverage-limits section. One should fix and several optionals reached the adversary; **the should fix was attacked down, three findings were killed, and the adversary added one of its own.**

### The head, verified against the SHA rather than the PR

`2e4821f7`, checked at the start of the pass and again at the end, unchanged. CI read from `commits/2e4821f7/check-runs`: 14 runs, and a filter for `conclusion != "success"` returns 0.

One thing worth recording: the legacy `/status` endpoint reports `state: "pending"` with `total_count: 0`, because **no commit statuses are registered on this repository at all** — the check-runs API is the only gate here. That is not a stale or partial green, but it is another way the obvious command answers the wrong question.

All reading and both mutations ran against `git archive 2e4821f7 | tar -x` into scratch; `git status --porcelain` was empty at both ends.

### Why the one should fix did not survive

`extension/app.js:2334` (`visibleProblem` unpinned by any test) is **pre-existing behaviour untouched by this diff** — the change adds a branch and removes nothing — and the demonstration offered for it was void. The coupling this diff *does* create is pinned by the new panel test. It is filed as part of **#857**.

### Five optionals, filed rather than fixed here

- **#857** — the new branch's boundary is pinned in neither direction: broadening `/not signed in/i` to `/not/i` at `extension/identity.js:112` leaves all 461 JS tests green, because the fall-through fixture at `granted-scopes.test.mjs:77` does not overlap the pattern. It also carries the `app.js:2334` gap and **the adversary's own finding, which is the one all four reviewers missed**: every reviewer questioned what happens if Chrome's string is *different* (branch dead, fix does nothing) and none asked what happens if it is *broader* — in which case the branch fires on a state it was not written for and tells the owner the wrong thing with confidence. That is the direction this repository's failures actually run in.
- **#858** — a partial scope grant is treated as a refusal at `extension/app.js:2330-2332` while four siblings in the same file do not (`:2926`, `:2945`, `:2956`, `:3008`); the harness never passes a `granted` set, so `partial` is unreachable in every browser-driven test. Pre-existing, and whether a partial grant should sign him in is his ruling.
- **#859** — `/bad client id|invalid client/i` at `extension/identity.js:126` may not match Chromium's own constant, and nobody in the pass could verify it.
- **#860** — the signed-out welcome draws only the Chrome identity path (`extension/app.js:2740`), which is the one that cannot succeed in exactly the state the new message describes.

### Two caveats on the record, neither blocking

**Nobody in this pass executed the two new Playwright guards** — playwright is not installed on this machine, and all four reviewers hit the same wall and said so. CI executed them at `2e4821f7` under the anti-skip gate at `.github/workflows/ci.yml:215-253`, which is the thing that makes a collection-skip fail rather than pass quietly.

**The string the whole fix keys on is asserted only by fixtures this repository supplies.** No test, here or in CI, observes Chrome producing it. That is the premise behind #857 and it cannot be closed from inside the repo.

### One ordering note

#832 and #821 share no file, so no conflict and no CI signal will fire — but #821 rewrites `tools/panel_harness.py`, and this PR's new `tests/test_signing_in_says_what_happened.py` builds its page through that same harness. They are the pair most likely to be individually green and jointly broken, so they should be merged adjacent and the second one re-run against the merged result. #821 is currently blocked on its own gate findings, so nothing is pending here.

**Not merging without his word.**

muhammadbayoumi:
## For whoever merges: the squash body needs replacing, and the branch does not

`squash_merge_commit_message` on this repository is **`COMMIT_MESSAGES`**, not `PR_BODY`
(`gh api repos/muhammadbayoumi/ScrapeX --jq .squash_merge_commit_message`). So a squash-merge
writes **the commit message** onto `main`, and the PR body — where the corrected measurement and
the #837–#840 links live — does not travel. Two claims in that commit message are now false, and
both would land verbatim.

Confirmed on the rebased head `567f38c`; the rebase itself is clean, `git diff --stat 28286ac 567f38c` is empty, so this is about the message only.

**1. It says neither guard was executed locally. Both were.**

> `node` and `pytest`/`playwright` are absent on this machine, so neither guard was executed
> here; CI runs both

True when written — the commit was made before the tooling was installed. Afterwards both ran,
and both were made to fail on purpose:

- `node --test extension/tests/*.test.mjs` → **461 pass, 0 fail**. Widening the new pattern to
  `/not/i` left all 461 green (the #857 gap, verified here independently); corrupting it to
  `/not signed on/i` failed exactly one test, `actual 'failed'` vs `expected 'no-chrome-account'`,
  with the sibling fall-through test still passing.
- `pytest tests/test_signing_in_says_what_happened.py` with a real Playwright chromium → **27
  pass**. Against the same corruption it printed the defect as the user meets it:
  `assert 'The user is not signed in.' == ''`.
- `pytest -q -m extension` → no failures, no errors.

**2. The `ci.yml` citation has shifted.** `455` → **`504`**: #830 added 49 lines to that file, and
455 + 49 = 504. `node --test extension/tests/*.test.mjs` is at `:504` on this head, and
`apps_script` at `:511`.

### The replacement paragraph

Ready to pass as `gh pr merge 832 --squash --body-file <file>` — no new head, no CI re-run:

```
MEASURED, AND EACH GUARD BROKEN ON PURPOSE FIRST. `node --test extension/tests/*.test.mjs`
(ci.yml:504): 461 pass, 0 fail. `pytest tests/test_signing_in_says_what_happened.py` against a
real Playwright chromium: 27 pass. `pytest -q -m extension`: no failures. Corrupting the new
pattern to `/not signed on/i` failed exactly one unit test -- `actual 'failed'` vs `expected
'no-chrome-account'` -- and printed the defect at the panel level as the user meets it,
`assert 'The user is not signed in.' == ''`; the sibling test pinning that an unknown message
still reaches `failed` stayed green throughout, so the two halves are independent.

WHAT THE GUARD DOES NOT PIN, and it is filed as #857: widening the pattern to `/not/i` leaves
all 461 green, because the fall-through fixture carries no "not". The boundary is held against
too-narrow and not against too-broad, which is the direction that would give confident wrong
advice.
```

Amending the commit instead would cost a force-push and a full CI cycle for a message change.
Replacing the squash body at merge time costs nothing and puts the same text on `main`.

Still not merging from here.

Issue #835, opened 2026-09-09
ci.yml and pyproject cite a docs-gate completeness test that #467 deleted, and nothing checks the docs mark today


muhammadbayoumi:
Found in the pre-merge review of #830 and out of scope for a `ci.yml` step, so filed rather than fixed. Per CLAUDE.md.

## Three facts

**1. The guard was deliberately deleted.** `tests/test_the_docs_gate_is_complete.py` was added by `3d265cd` (#216), modified by `31e6e9b`, and removed by `f87391f` (#467), whose body says the four suites deleted there "test the registers, not the product". Deliberate — not an accident to undo without a decision.

**2. Live configuration still says it exists.** Two of these are read as documentation of what CI enforces:

- `.github/workflows/ci.yml:335` — "tests/test_the_docs_gate_is_complete.py catches a test file that reads a document and forgets the mark; this catches the set being emptied some other way"
- `pyproject.toml:86`
- `tests/marks.py:11`
- `tests/test_no_tag_is_cut_while_the_warehouse_is_ahead.py:30`
- `tests/test_the_markup_we_ship_closes_its_own_tags.py:42`
- `tests/test_the_release_the_documents_ask_for_is_the_one_that_would_run.py:61`

The `ci.yml` one matters most: the step it annotates is a floor on the *count*, and its comment tells the reader that completeness is covered elsewhere. It is not.

**3. Nothing checks the `docs` mark is complete.** `tests/test_the_extension_gate_is_complete.py:67,104` calls `carries(path, "extension")` and is the only completeness guard of its kind. A new test file that reads a document and forgets `@pytest.mark.docs` stops running on exactly the documentation PRs it was written for, and nothing goes red — the failure mode `ci.yml:275-283` describes for the extension set.

## Why the repo's own tripwire cannot see it

`tests/test_the_tests_name_tests_that_exist.py:76`:

```python
REFERENCE = re.compile(r"`(test_[A-Za-z0-9_]{6,})`")
```

The backtick must sit immediately before `test_`. Ran it over `tests/test_*.py`: **zero** docs-gate hits. Two of the citations are path-form (backtick, then `tests/`, and `.py` before the closing backtick), the third has no backticks at all, and `tests/marks.py` is not scanned because the glob is `test_*.py`. So the `HISTORICAL` dict at `:91-112`, which exists precisely to record a deliberately deleted test, can never be triggered for this one.

## Options

[...]

Issue #831, opened 2026-09-09
The staleness check misses a same-size edit on Windows: the cheap stamp skips the digest that would catch it


muhammadbayoumi:
Found while measuring for #830, on a fresh clone on Windows. Not fixed, not touched — the change in #830 is `ci.yml` only, and this fails identically with that diff stashed.

## Measured

```
$ python -m pytest tests/test_the_engine_knows_which_code_it_is_running.py::test_no_repository_at_all_is_not_a_reason_to_stop_answering
E  AssertionError: the staleness check must not depend on git being present
E  assert False is True
tests/test_the_engine_knows_which_code_it_is_running.py:632
```

Windows 11, Python 3.14.7, `c30b098`. Deterministic — not a flake, it fails every run.

## The cause, proved rather than read

`tests/test_the_engine_knows_which_code_it_is_running.py:620,626` writes `V = 1\n`, seals, then writes `V = 2\n`. Different content, **same length**, microseconds apart.

`scrapex/provenance.py:337` short-circuits on the cheap stamp:

```python
if stat.st_mtime == mtime and stat.st_size == size:
```

and Windows' file-write clock does not tick between those two writes:

```
size  : 7 -> 7                            | equal: True
mtime : 1788977492.8162596 -> ...2596      | equal: True
```

Both halves equal, so the digest at `:294`/`:329` — the thing that would have caught it — is never compared. The second check at `:361` cannot rescue it either: it asks `st_mtime > sealed_at`, and the mtime did not move.

On Linux the same two writes land on distinct nanosecond mtimes, which is why CI is green and has always been green.

## Why it may be more than a test-environment quirk

`stale` is described at `:35` and `:43-45` as the half that *proves the fault* on an editable install. The owner's machine is Windows. A source edit that keeps a file's byte count and lands inside one clock tick is not reported as stale there — the digest that would notice is skipped before it is reached.

**Bounded honestly:** a human edit is seconds apart and usually changes the size, so the practical exposure is small. A generated or scripted rewrite — a codegen step, a checkout, a patch applied by a tool — is where same-size-same-tick is reachable.

## Options

1. **Do nothing.** CI is Linux and green. Cost: the suite cannot be run green on the platform the engine actually ships to, and the next person to run it on Windows re-derives this.
2. **Make the test not depend on the clock** — differing lengths, or an explicit `os.utime` between the writes. Effort S, risk none. Fixes the red suite on Windows and leaves the false negative in place.
[...]

Issue #842, opened 2026-09-10
Six required checks are advisory for the only account that can merge: one aggregator job, then enforce_admins


muhammadbayoumi:
**His ruling, 2026-09-10: close #682 with an aggregator job first, then set `enforce_admins=true`. The order is forced, not preferred.**

## The problem, measured live

```
gh api repos/muhammadbayoumi/ScrapeX/branches/main/protection
  required_status_checks.strict        true
  required_status_checks.contexts      lint, scope, test, migration-authority, contract-parity, CodeQL
  enforce_admins.enabled               FALSE
  allow_auto_merge                     false
```

`enforce_admins: false` means the six required contexts are **advisory for the only account that can merge**. Every session merges as that account. So the documented gate and the enforced gate are two different systems, and `CLAUDE.md:20` ("One session merges. Ask; default to not merging.") has no mechanical existence at all.

#682 is the second half: a required context that **never attaches** to a PR does not block it — GitHub reports the PR `clean`. That is this repository's signature false-green shape (the check's input disappeared and the absence read as a pass) applied to branch protection itself.

## Why the order is forced

Turning on `enforce_admins` while #682 is open makes the hole **fatal instead of cosmetic**. Today an unattached context is survivable because the admin merges anyway. With admins enforced, a context that never reports leaves the PR permanently unmergeable — and there is no control in the panel to clear it, which lands on "a button that cannot work is worse than no button" from the other direction: a gate with no exit.

## The change

1. **One aggregator job**, `gate`, running `if: always()` and needing every other job. It asserts the **conclusion** of each upstream job by name — `success`, or for a scope-skipped job that the scope genuinely excluded it. Absence becomes failure instead of success.
2. **Branch protection requires only `gate`.** Six contexts become one that cannot be satisfied by silence.
3. **Then `enforce_admins = true`.**

The pattern is not new here; it already runs one level down with measured floors, and the aggregator is the same idea at job level:

- `.github/workflows/ci.yml:258` — `test "$(collected tests/test_panel_dom.py)" -ge 40`
- `.github/workflows/ci.yml:272-273` — per-file minima, 24 and 10
- `.github/workflows/ci.yml:297` — `-ge 300` for the extension set
- `.github/workflows/ci.yml:309` — `-ge 150` for the docs set

## What is explicitly NOT part of this

[...]

PR #851, opened 2026-09-10
Three edits to one file: stop running CI twice, cancel superseded runs, and run the suite once properly

### LINKED ISSUE #645: migration-authority is 54.7% of all CI compute and the slowest job in the matrix; after the squash it buys 15.7 seconds
`migration-authority` runs the **entire 3,684-test suite a second time** so that `migrate()` is real instead of restored from a template. After R-84 squashed the chain, that difference is **15.7 seconds**.

## Measured, both runs on `origin/main` 93b31e38, same 3,684 tests

| run | terminal summary | total test time | result |
|---|---|---|---|
| fast template | `schema template: 885 restored (10.6ms each), 25 real migrations` | **680.0s** | 3,674 passed, 10 skipped, 0 failed |
| `SCRAPEX_FULL_MIGRATIONS=1` | `schema template: DISABLED — every migrate() ran for real` | **695.6s** | 0 failed, 17 skipped |

**Delta: +15.7s (+2.3%).** Per call: real `migrate()` 12.4ms against the template's 4.6ms, because `db/engine/migrations/` is empty and `_migration_files()` returns one entry.

## What the duplicate costs, measured from the Actions API over 30 days

| job | runner-minutes | share of all CI compute | median duration |
|---|---|---|---|
| **migration-authority** | **6,307** | **54.7%** | **12.8 min** |
| test | 4,848 | 42.0% | 9.2 min |
| lint | 186 | 1.6% | 0.3 min |
| contract-parity | 123 | 1.1% | 0.2 min |
| scope | 66 | 0.6% | 0.1 min |

617 CI runs, 11,529 runner-minutes total. **`migration-authority` is also the slowest job in the matrix, so it — not `test` — is what a pull request actually waits for.**

## The change

Run the suite **once**, with `SCRAPEX_FULL_MIGRATIONS=1`, and delete the duplicate job. That is strictly *more* protection than today's `test` job (every PR would exercise the real stream, which today only the second job does) for +15.7s, and it removes ~54% of CI compute and about 3.5 minutes of PR latency.

Framing owed to the merge session: the honest argument is **"run it once, properly, and stop running it twice"** — not "the real stream buys nothing".

## What must ride with it, or the suite quietly weakens

**7 tests in `tests/test_fast_migrations.py` SKIP under the flag** — measured, that is exactly the 10-versus-17 skip difference above. They are the only thing pinning the schema-template mechanism. If the fast-path job disappears they stop running anywhere. Either keep a seconds-long step that runs that one file on the fast path, or retire the template mechanism with them (see the conftest issue in this milestone).

## Feasibility, checked against the guards

- **Nothing pins the job or the variable.** No test names `migration-authority` or `SCRAPEX_FULL_MIGRATIONS` in `.github/workflows/ci.yml`.
- `tests/test_the_workflows_check_out_enough_history.py:117` asserts **at least 2** jobs across all workflows run pytest. Measured today: **5** (`ci.yml` test, `ci.yml` migration-authority, `publish-docs.yml` publish, `release-engine.yml` build, `release-extension.yml` package). Removing one leaves 4 — the guard still holds.

**Effort S. Regression risk LOW**, provided the 7 fast-path tests keep a home.

Related: #623 states this job's cost as "about 350 s of real `migrate()` calls, measured". That was true before the squash; it is now 15.7s. Worth correcting there.

### LINKED ISSUE #646: Every pull-request SHA runs CI twice: the push copies cost 38% of all runner-minutes
`.github/workflows/ci.yml` triggers on **both** `push` and `pull_request` with no filters, so every commit on a pull-request branch runs the whole matrix twice.

## Measured, Actions API, 30 days

- 617 CI runs over **378 distinct SHAs**
- **231 of those SHAs ran CI twice** — once as `push`, once as `pull_request`
- The `pull_request` copies alone cost **4,438 of 11,529 runner-minutes = 38% of all CI compute**
- Live example: PR #609's head `3c188070` showed **14 check rows for 7 checks**

## The change

```yaml
on:
  push:
    branches: [main]
  pull_request:
```

Every branch commit is still covered by the `pull_request` run; `main` keeps its post-merge run. Nothing loses coverage.

## Two things that must be said with it, or the change is wrong

1. **A branch with no open PR would stop being tested on push.** For this repository that is the normal way work travels (branch → PR → squash), but it is a real behaviour change and it is the owner's call, not a tidy-up.

2. **The duplication is currently load-bearing for reading checks, and that is an honest tension.** The two runs tier differently: for `push`, `ci.yml` computes the scope base from the branch's previous tip; for `pull_request` it uses `github.event.pull_request.base.sha`, i.e. the whole PR against the base. Both rows are called `test`, so a reader who opens the wrong one can see a `test` row that passed while the engine tier never ran. Narrowing the trigger **removes that ambiguity by leaving only the PR run** — which is the one that describes the whole change — so it improves the read as well as the cost. Verified on PR #605: both runs printed `SCOPE: docs`, and on PR #609 the merge session confirmed the `pull_request` run's engine step reported SUCCESS rather than skipped.

**Effort S. Regression risk LOW.** Nothing in `tests/` reads `ci.yml`'s `on:` block — the only two guards that read the file assert on the lint step's contents (`tests/test_the_lint_gate_cannot_be_quietly_widened.py:59,68`) and on `fetch-depth` (`tests/test_the_workflows_check_out_enough_history.py`).

### LINKED ISSUE #647: No concurrency group: a superseded push keeps a full 13-minute matrix running
No workflow in `.github/workflows/` declares a `concurrency` group, so pushing a new commit to a pull request leaves the previous run going to completion.

## Measured

A superseded full-scope run costs **~13.2 minutes of wall clock and ~22 runner-minutes** (median full run, 30-day sample) and its answer is discarded the moment it lands, because the branch has already moved.

Combined with the double trigger (see the sibling issue in this milestone), a single superseded commit can leave **four** obsolete runs in flight: push and pull_request, each running `test` and `migration-authority`.

## The change

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

**On `main` it must not cancel.** A post-merge run on `main` is the record that the merged tree is green, and cancelling it because the next merge arrived would erase exactly the signal the repository relies on. Either exclude `main` from the group, or set `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`.

## Interaction worth stating

With no cancellation and CI on both triggers, **every intermittent gets between two and four independent chances to fire per SHA**. Cancelling superseded runs reduces flake exposure as well as cost — but it is not a fix for a flaky test, and it must not be presented as one.

**Effort S. Regression risk LOW.** Nothing in the suite reads `ci.yml`'s trigger or concurrency configuration.

muhammadbayoumi:
Closes #645. Closes #646. Closes #647.

The three #656 grouped as "three small edits to one file that remove roughly two thirds of CI compute". One commit each, in that order.

> **Based on #830, not on `main`.** #645 deletes the `migration-authority` job, which sits immediately after the step #830 adds — a branch off `main` would conflict. GitHub will retarget this to `main` when #830 merges.

---

## #647 — `concurrency` (`b6c72d1`)

No workflow declared a group, so pushing again to a branch left the previous matrix running to completion: ~13 min wall, ~22 runner-minutes describing a tree that had already moved.

`main` is excluded from the **cancellation**, not from the group — `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`. A post-merge run on `main` is the record that the merged tree is green; cancelling it because the next merge arrived erases the signal. Grouping still serialises its runs.

Not offered as a flake remedy: fewer chances to fire per SHA lowers exposure without fixing a flaky test (#649).

## #646 — one run per SHA (`c998a01`)

`on: push` was unfiltered, so every commit on a branch with an open PR ran the matrix twice. 231 of 378 SHAs over 30 days; the duplicate copies were 4,438 of 11,529 runner-minutes.

**The behaviour change, stated rather than buried:** a branch with no open pull request stops being tested on push. Work travels here as branch → PR → squash, so the uncovered window is the time before the PR is opened. The owner approved this explicitly.

It also removes an ambiguity — the two runs tiered differently (`push` from the branch's previous tip, `pull_request` from the base commit) and both rows were called `test`.

## #645 — run it once, properly (`668d170`)

The flag moves onto the surviving suite step; the duplicate job is deleted. **Strictly more protection than the step it replaces**: every PR now exercises the real stream, which before this only the second job did.

### #645's own budget does not survive measurement

| | #645 says | measured on this branch |
|---|---|---|
| cost of `SCRAPEX_FULL_MIGRATIONS=1` | **+15.7s (+2.3%)** | **+73.1s and +99.8s** on a ~345s suite — **+21% to +30%** |

Two pairs of whole-suite runs, alternating, tree untouched between them. The issue's figure was taken when `db/engine/migrations/` was **empty**; `0018`–`0020` have landed since, so every real `migrate()` replays the baseline **and three files** instead of one.

**So the compute win is real and the latency win is not.** Deleting the duplicate removes roughly half this workflow's runner-minutes — but the job removed was the *slowest* one (13m0s and 14m11s on this branch's own runs), and the surviving step grows by about what that job cost in wall clock. **A merge does not get faster. CI stops paying twice for the same answer.** #645's "about 3.5 minutes of PR latency" should be struck.

### The rider #645 named, and how it is held

Seven tests in `tests/test_fast_migrations.py` **skip** when the flag is set — they are the only thing pinning the template mechanism, and until now they ran in the job that had no flag. They get a step of their own, deliberately with **no `env:` block**.

**The flag is on the step, not the job, and that is load-bearing.** A job-level `env:` would reach the new step too and silence all seven where nothing would notice — the exact shape #830 exists to catch. Verified both ways:

```
$ pytest tests/test_fast_migrations.py                              -> 7 passed
$ SCRAPEX_FULL_MIGRATIONS=1 pytest tests/test_fast_migrations.py    -> 7 skipped
```

### What removing a job costs beyond the job

Four sentences named `migration-authority` in the present tense. Leaving them would create four fresh instances of #835 — a live citation of something that does not exist — so they move with it:

- **`ci.yml`'s `scope` job**: "Two jobs need the answer now — `test` and `migration-authority` — and they must run in PARALLEL". After this, one job does, and the reason `scope` is a separate job at all is spent. **Folding it back is a separate change with its own argument** — recorded as an issue, not ridden along here.
- `release-engine.yml:132` and `tests/test_no_tag_is_cut_while_the_warehouse_is_ahead.py:12`: the substance survives — CI's suite still reads only its own branch's diff — so they name the suite instead of the job.
- `tests/test_a_carried_over_ledger_does_not_block_the_next_migration.py:21` is history, so "which runs" becomes "which then ran".

---

## State

Measured on **Python 3.14.7** (CI pins 3.12 — #656 item 5, not addressed here). Four whole-suite runs on this branch:

| run | result |
|---|---|
| template ×2 | 3,597 tests · 14 skipped · 334.9s / 353.6s |
| `SCRAPEX_FULL_MIGRATIONS=1` ×2 — **the configuration this PR ships** | 3,597 tests · **0 failed** · 21 skipped · 434.7s / 426.7s |

The one failure in the first template run is #831, pre-existing and unrelated; it did not recur. The 21-vs-14 skip difference is exactly the seven `test_fast_migrations.py` tests, which is why they now have their own step.

Also green: the eight test files that read workflows, and the new fast-path step run exactly as CI runs it.

Browser suites collection-skip on this machine (no Chromium, no node), so their contribution is absent rather than green. CI runs them.


muhammadbayoumi:
One dated instance of #646, measured on an unrelated branch, plus one interaction between your
three commits that the arithmetic above does not account for. From #832 (`claude/auto-logged-account-tool-4318c9`), head `2e4821f7`.

## The instance

Two `CI` runs, same SHA, both green, 78 seconds apart:

| run | event | job timings | runner time |
|---|---|---|---|
| [34392446185](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/34392446185) | `push` | scope 6s · contract-parity 11s · lint 21s · test 11m41s · migration-authority 12m59s | **25m18s** |
| [34392582608](https://github.com/muhammadbayoumi/ScrapeX/actions/runs/34392582608) | `pull_request` | scope 8s · contract-parity 15s · lint 26s · test 12m13s · migration-authority 12m46s | **25m48s** |

51m06s of runner time for one answer. Billed per job rounded up, the redundant `push` copy is
**28 runner-minutes**.

Cross-checking your aggregate: 4,438 ÷ 231 = **19.2 minutes per duplicated SHA**. This one cost
28. Above your average, and consistent with it — nothing here contradicts the 30-day figure.

## The interaction #646 and #645 have with each other

`migration-authority` is **779s of the 1,518s** in that duplicate copy — **51%** of what #646
reclaims on this SHA is the job #645 deletes outright.

So the two savings overlap, and on this evidence roughly half of #646's 4,438 duplicate
runner-minutes is `migration-authority` running twice — minutes #645 removes on its own by
deleting the job. **The two figures cannot be added.** Whichever lands first collects most of
that half; the second collects the remainder.

That does not weaken either commit — #646 still stops paying twice for `test`, and #645 still
removes the slowest job — but a combined claim built by summing them would overstate the win by
something near the size of one duplicated `migration-authority`.

## And the ambiguity, from the consumer's side

Your #646 note that "both rows were called `test`" has a downstream cost worth naming: `gh pr
checks 832` returns **14 rows for 7 checks**, identically named, differing only in run id. A
session reading that cannot tell which tier answered without opening both runs — and `gh pr
checks` reports a non-zero exit while either copy is pending, so a green PR reads as not-green
for as long as the slower duplicate lasts. Two of yesterday's status reads on #832 hit exactly
that.

No action requested on my side; #832 is unaffected either way. Handing over the measurement
because #646's window is still open and the overlap changes what the stack should claim.

muhammadbayoumi:
## Merge session: one question before this can be gated, and two mechanical notes

The body is unusually complete — it names the behaviour change on `push`, it strikes its own issue's latency figure after measuring it, and the reasoning for the flag being on the step rather than the job is exactly right. This is not a request to re-argue any of that.

**The question is about something the body does not mention: branch protection.**

### `migration-authority` is a required context on `main`, and this deletes the only thing that can produce it

Measured just now:

```
gh api repos/muhammadbayoumi/ScrapeX/branches/main/protection
  required_status_checks.strict     true
  required_status_checks.contexts   lint, scope, test, migration-authority, contract-parity, CodeQL
  enforce_admins.enabled            false
```

The PR deletes the `migration-authority` job (`gh pr diff 851`, hunk `@@ -405,82 +461,6 @@`). After it merges, **no commit can produce that context again**.

That matters because of #682, which is open and unfixed: a required context that never *attaches* does not block a PR — GitHub reports it `clean`. So the effect is not a red check anyone would notice. It is that `main` quietly enforces **five** contexts instead of six, on this PR and on every PR after it, with nothing anywhere going red.

**The body's claim "strictly more protection than the step it replaces" is true about coverage and not about enforcement.** I agree the *testing* is strictly better — every PR now exercises the real migration stream where before only the second job did. But the *gate* loses a name it is configured to require. Those are two different things and only the first is argued.

So, concretely:

1. **Was the branch-protection list checked before the job was deleted?** Protection was enabled on 2026-09-05 (see #682) and the required list may simply not have been on the radar.
2. **If it was, what is the intended replacement** — remove `migration-authority` from the required contexts (an owner-only settings change, since it is his repository), rename the surviving suite step so it still reports under that context name, or #842's aggregator job, which is filed and unstarted and would make this whole class of question go away?
3. **If it was not, does the answer change the change?** It may not — the coverage argument stands on its own. But the settings change has to be made deliberately by the owner rather than discovered later, and it should be named in the body so the merge is not the first time anyone reads it.

This is the merge session asking, not blocking on taste. **It is the owner's ruling to make, and I have put it to him.** Nothing else here needs an answer from you.

### Two mechanical notes, no action needed from you

**Its `CLEAN` today is earned with four contexts, not six.** Measured on head `668d170b`:

```
gh api repos/.../commits/668d170b/check-runs --jq '.total_count'   -> 4
  attached: test, lint, scope, contract-parity
  absent:   migration-authority, CodeQL
```

`migration-authority` is **absent, not skipped** — a genuine skip still reports (inside the same run, "Documents only" and "Extension gate only" both show `conclusion=skipped`). CodeQL is absent because default setup does not attach to a PR whose base is not the default branch; PR #330 is the control, with real analysable code and zero CodeQL rows. Neither is your doing — both follow from the base being `fix/a-suite-that-skips-cannot-report-green` rather than `main`, which the body already explains and which was the right call. Recording it only so nobody reads today's badge as a `main`-quality signal.

**Retargeting alone will not re-gate it.** When #830 merges, GitHub retargets this to `main`, but that fires `pull_request` with `action: edited`, which the default `types` do not match — no workflow re-runs, and the SHA keeps its four rows. What actually closes the gap is `strict: true` forcing a branch update, which produces a new head SHA whose `pull_request` run is based on `main` and does attract CodeQL. I will confirm the six contexts on that new head before gating, rather than reading the current badge.

**And the update after #830 lands conflicts, so I will rebase rather than merge.** Simulated without touching any tree: squashing #830 onto `main` and then merging this branch gives `CONFLICT (content) in .github/workflows/ci.yml` — one hunk, because #830 inserts its step immediately above the block this PR edits and immediately below the job it deletes. It resolves mechanically by taking the branch side entire: this branch's `ci.yml` already contains #830's 49-line step verbatim, so nothing of #830 is lost.

The trap, recorded so it is not rediscovered: the conflicting "theirs" block **opens** with #830's step, which genuinely must survive — so "keep both" looks correct and would silently reinstate the `migration-authority` job, green, with nobody noticing. After resolving I will assert on the file itself: `migration-authority` absent, #830's step present.

muhammadbayoumi:
## Correction from the merge session: the premise of my question was wrong

My comment above asserted that `main` enforces six required contexts and that deleting the `migration-authority` job drops enforcement from six to five. **Branch protection is not in force on this repository, and cannot be on its current plan.** Measured just now:

```
gh api repos/muhammadbayoumi/ScrapeX/branches/main --jq .protected
  false

gh api repos/muhammadbayoumi/ScrapeX/branches/main/protection
  403  "Upgrade to GitHub Pro or make this repository public to enable this feature."

gh api repos/muhammadbayoumi/ScrapeX --jq '{private, visibility}'
  {"private": true, "visibility": "private"}
```

Earlier in this same session that endpoint returned a full configuration — `strict: true`, the six contexts, `enforce_admins: false` — which is what my question was built on. It does not now. Either the plan changed under us or the earlier read came back from a stored configuration the API no longer serves; I am not going to guess which, and the operative fact is the one measured last: **nothing is enforced today.**

**So nothing about your change weakens enforcement, because there is none to weaken.** That half of my comment is withdrawn, and I am sorry for the noise.

### What survives, in the opposite direction

The severity does not disappear; it inverts. If protection is ever restored — GitHub Pro, or making the repository public — the recorded contexts list comes back with `migration-authority` in it, naming a context **no commit will be able to produce**. Under the `enforce_admins: true` that #842 exists to set, that is not a silent five-of-six; it is `main` becoming unmergeable until someone edits the list.

**#851 and #842 are mutually destructive whichever lands second.** That is the real question, and it is still the owner's:

- if #851 lands first, #842 must be written knowing `migration-authority` is gone, and the contexts list must be edited when protection is restored;
- if #842 lands first, it will encode a context this PR then removes.

Nothing here asks you to change the diff. The coverage argument stands exactly as written and I have not found a defect in it.

### And two things I said that are also now different

**"Its CLEAN is earned with four contexts, not six."** True as a count — `check-runs` at `668d170b` returns 4 — but I framed it as a protection gap. With no protection active, every `CLEAN` in this repository is uninformative: it reflects the absence of a gate rather than the satisfaction of one. That is as true of the other six open PRs as of this one.

**"`strict: true` will force an Update branch, and the new head will attract CodeQL."** `strict` is not enforced either, so nothing will force this branch to update after #830 lands. If the CodeQL gap on this branch is to close, the merge session has to trigger the update deliberately rather than wait for GitHub to require it. I will.

The forced order (#830 first, because its head is an ancestor of yours) and the rebase trap I described are unaffected — both were measured from git, not from the API.

muhammadbayoumi:
## Pre-merge review, round 1 — panel over four dimensions, then an adversary

Nine findings. The adversary killed four and downgraded two; I killed one myself by measuring it. **Two must fixes and three wrong numbers survived, and `d25fada` is all of them.**

| finding | panel | outcome |
|---|---|---|
| concurrency: `main` loses a *pending* run | must fix | **CONFIRMED** — fixed |
| the flag's placement is unpinned (3 silent mutations) | must fix | **CONFIRMED** — fixed, two layers |
| the template step has no floor | must fix | **CONFIRMED** — same defect, one layer of the above |
| three factual defects in the new comments | should fix | **CONFIRMED, one worse than reported** — fixed |
| `migration-authority` is a required status check | must fix | **KILLED** — see below |
| CI can no longer exercise the template at scale | should fix | **KILLED** — a `write_line`, never a gate, and it printed 965 against a floor of 500 |
| the rewritten "cannot catch this" sentences | should fix | **KILLED** — imprecision inherited verbatim; the conclusion holds |
| `tests/conftest.py:97` says "on every push" | should fix | **downgraded** — #655 already owns that sentence and two more clauses of it |
| the new step sits above #830's floor | optional | **KILLED** — 0.9s, and it is *below* the suite, which is the side the principle cares about |
| no `workflow_dispatch` | optional | **KILLED as framed** — a draft PR triggers `pull_request` |

### The required-check finding, killed by measurement

The panel reported `migration-authority` as a required context on `main`, quoting #842's "measured live" block. Measured here instead:

```
gh api repos/muhammadbayoumi/ScrapeX/branches/main   -> "protected": false
gh api .../branches/main/protection                  -> 403 "Upgrade to GitHub Pro or make this repository public"
gh api .../rulesets                                  -> 403
```

The repo is private on a free User plan, so branch protection is **not available at all**. #842's quoted block does not describe this repository, and #653 agrees with my read. So deleting the job orphans no required context — and **#499 ("make migration-authority a required check") is now dead**; it asks for a job that no longer exists and a feature this plan does not offer. Both need a correction of their own.

### What the concurrency group actually does

`cancel-in-progress: false` does **not** keep a run alive. GitHub's rule:

> any existing `pending` job or workflow in the same concurrency group will be canceled

Unconditionally — that flag governs only the *running* slot. A group holds one running and one pending run, so three merges inside one run's wall clock evict the middle one **while it is still pending**.

That costs a release, not a row: `tag-the-release.yml:23,49` consumes CI through `workflow_run` and tags only on `conclusion == 'success'`. A `VERSION` bump merged mid-burst would never be tagged — the defect R-87 built that mechanism to remove. Simulated over the last 297 squash merges at the 13.2-minute median: **10 (3.4%) would have lost their post-merge run.**

**#647 offered two options and I took the wrong one.** `main` now leaves the *group*, via `github.run_id`.

### Three edits undid #645 and nothing went red

All three green across the 190 tests that read a workflow:

| mutation | before | after `d25fada` |
|---|---|---|
| move the `env:` from step to job | 190 passed | **red** — `test_the_flag_is_scoped_to_that_step_and_no_wider` |
| delete the template step | 190 passed | **red** — `test_something_still_runs_the_template_suite_without_the_flag` |
| delete the flag from `ci.yml` entirely | 190 passed | **red** — `test_the_whole_suite_replays_the_real_migration_stream` |
| strip the new floor's `--junitxml` | n/a | **red** — `test_that_step_would_fail_rather_than_skip_quietly` |

The third is the one that cost the most and showed the least: it silently returns CI to testing against the schema template forever.

Two layers, because neither covers the other — the YAML test is blind to a flag arriving through `$GITHUB_ENV`, and the runtime floor stays green under mutation 3:

- **`tests/test_the_real_migration_stream_is_replayed_in_ci.py`** reads the effective env (workflow ∪ job ∪ step). Keyed on step **shape**, not a `run:` string, because #654 will add `-n` and the template step already carries `--junitxml`. Its reader asserts it matched exactly one step, so a parser that stops matching fails loudly instead of passing over nothing.
- **The template step gets a `--junitxml` floor.** Without it the step is green whether its seven tests pass or skip — exit 0 both ways under `-q`. Verified: `7 passed, 0 skipped` → floor passes; `7 skipped` with pytest exit 0 → floor exits 1; and **under `PYTHONOPTIMIZE=1` it still exits 1**, because it raises `SystemExit` rather than asserting (#833 is open against the sibling floor for exactly that).

### The numbers I carried instead of measuring

| claim | measured |
|---|---|
| "+21% to +30%" | run `34428035437` ran both suites off **one tree**: 651s unflagged, 743s flagged → **+92s, +14.1%**. The seconds travelled (local +73/+100s brackets 92s); the percentage could not, because the denominators differ ~2× |
| **"A merge does not get faster"** | **wrong in direction** — critical path **785s → 750s**, and **17% better** at the median of 802 full runs |
| "this step grows by about as much as that job cost in wall clock" | 92s against 785s — off by **8.5×** |
| #646's census: 617 runs, 4,438 of 11,529 min | that is the most recent 617 runs, not a window. Over 2026-08-11..09-10: **1,322 runs, 8,591 of 23,369 min**. The share it derives (38% vs 36.8%) is what survives |

The latency correction matters most: the true reason is stronger than the one I wrote. The removed job was **parallel** and only 75–95s longer than `test`, so absorbing +92s lands back at almost exactly the old critical path. The win bought here is **compute, not latency — 69.5% of this workflow's billed runner-minutes**, which makes the PR title's "two thirds" the accurate figure and #645's "54.7% of all CI compute" wrong for any 30-day window (the real share is 42.0%).

The PR body above has been corrected accordingly, and so has my comment on #645.

Issue #844, opened 2026-09-10
release-engine.yml ends by echoing the manifest URL; publish-docs.yml already says printing a URL is not checking it


muhammadbayoumi:
**Two markdown files are better protected than the shipped binary — same directory, same author, two standards.**

## The defect

`.github/workflows/release-engine.yml` ends at line 670. Its last step, "Point the manifest at it", writes the manifest that redirects every panel in the world to the new engine, and then **prints the URL**:

```
.github/workflows/release-engine.yml:667-670
          echo ""
          echo "Published. Every panel that opens from now on reads:"
          echo "  https://raw.githubusercontent.com/${PUBLIC_REPO}/main/${MANIFEST_PATH}"
```

That is the last line of the file. Nothing reads the address back.

Eleven files away, `.github/workflows/publish-docs.yml:225-250` opens its equivalent step with a comment titled, verbatim:

> **PRINTING A URL IS NOT CHECKING IT.** This step used to end by echoing the two page addresses, which proves nothing: publishing can succeed while the page stays blank for ever

and then does what this one does not: curls the **public CDN address** (`raw.githubusercontent.com`, deliberately not the Contents API it just wrote through), retries for CDN lag, asserts the page still carries the attribute the loader hooks onto, and fails the job otherwise. Its reasoning applies word for word to the manifest — a successful write says nothing about what the reader will fetch.

**This is a DRY defect in this repository's own terms**: one piece of knowledge — how you prove a publish landed — held in two places that change for the same reason, and only one was updated.

## What it lets through

The manifest step is deliberately last, and its comment says why: *"a manifest written before the assets were attached would point every installation at a download that 404s."* The ordering protects against a manifest that is too early. Nothing protects against one that is **wrong or unreachable**: a CDN that has not caught up, a `put_to_hub` that returned success on a partial write, an asset upload that silently did not attach. The first party to notice is the owner, when the panel offers an update that cannot be downloaded — #771 and #773 territory, from the other end.

## The fix

Copy the pattern already present, after line 670. About 25 lines:

1. Fetch `https://raw.githubusercontent.com/PUBLIC_REPO/main/MANIFEST_PATH` with the same retry and sleep loop `publish-docs.yml` uses for CDN lag.
2. Assert the version it reports equals the tag — not merely that the file parses.
3. Assert the download URL it names answers 200 with the expected length, against the checksum step at `:516` that already computed one.
4. Fail the job otherwise.

[...]

Issue #843, opened 2026-09-10
CLAUDE.md is exempt from its own merge gate: the rules amend themselves through the one unguarded path


muhammadbayoumi:
**His ruling, 2026-09-10: a single review pass — not the full loop — on any change to `CLAUDE.md`.**

## The problem

`CLAUDE.md:107` ends the merge gate with: *"Documentation-only changes are exempt."*

`CLAUDE.md` is documentation. **So every amendment to the merge rules travels the one path in this repository with no gate on it.**

That is not theoretical. The rule that introduced the gate did exactly this, openly:

- PR #595, "Green is not mergeable" — merged **3.8 minutes** after it was opened, 0 reviews, 57 changed lines in one file, `93b31e3`. Its own body ends: *"This PR is documentation-only, and therefore exempt from the rule it adds."*

The exemption is honestly declared and correctly applied. The hazard is structural, not a bypass:

**`CLAUDE.md:148` instructs every session to delete.** *"Add by replacing. Before adding a line, delete one that stopped being true."* A session that finds a control inconvenient can satisfy the file's own instruction by removing it — and that removal merges through the unguarded path, in a file every session reads before every task.

Every other control in this repository is downstream of this one, because any control added to `CLAUDE.md` can be removed the same way.

## Why not the full loop

A four-dimension panel plus an adversary plus a re-review on a typo fix is the over-engineering `CLAUDE.md` refuses, and the exemption exists for a reason PR #595 states: #498 and #591 were 93 markdown files and zero lines of code between them.

## The change

Two lines, added by replacing, at `CLAUDE.md:107`: a change to `CLAUDE.md` itself is not exempt. It takes **one** review pass, not the loop, and that pass answers one question — **does this edit weaken a control, and if it deletes a line, what replaces what that line was protecting?**

Everything else documentation-only stays exempt.

## Cost

Effort: two lines. Risk: one extra pass on the sessions that touch the file. Blast radius: every session, since every session reads it. Maintenance: zero.

The cheapest item from the merge-rules study, and the one every other finding sits on top of.

Generated with [Claude Code](https://claude.com/claude-code)

muhammadbayoumi:
**Closed: this is now the rule.**

The ask was that `CLAUDE.md` stop being exempt from its own merge gate. `CLAUDE.md:124`
today:

> *"Documentation-only changes are exempt, except this file."*

Landed in `6449e36` ("The rules the merge study earned"). The `review` skill carries the
[...]

Issue #841, opened 2026-09-10
Measure the review loop's own yield before changing its stop condition — the data is already on origin


muhammadbayoumi:
**His ruling, 2026-09-10: measure the loop's yield first, then decide whether the stop condition changes.** Recorded here rather than acted on.

## The problem

`CLAUDE.md:102-107` ends the review loop when a pass **returns nothing**. An emptiness test over a generative reviewer has no guaranteed fixed point: the reviewer samples findings, it does not drain a finite list. That is why the rule needed two devices inside it — `R2`'s ban on fixing `optional` in the same PR ("without it the loop cannot terminate — a serious reviewer always finds one more small thing", PR #595) and `R4`'s five-pass tripwire.

**Both of the rule's numeric commitments were exceeded in its first recorded application.** #822: the passes went 3 → 0 (did not run) → 9, so the split rule fired at pass **seven**, not five; and two of the nine findings were folded into the same PR anyway "because they made the change indefensible".

**And a pass that returns nothing is currently indistinguishable from a pass that did not run.** #671 records two panels that reported `COMPLETE: true` with reviewers dead — "six agents died on a session limit, and a silent `slice(0, 6)` had already dropped three", and "the second lost its tests reviewer entirely". That is the repository's own false-green shape applied to the merge gate itself.

## What the industry does instead, verified at source

- **Google** publishes the opposite terminal condition: approve "once it is in a state where it definitely improves the overall code health of the system being worked on, **even if the CL isn't perfect**". Measured: over 80% of changes involve at most one iteration of resolving comments; median end-to-end review latency under 4 hours (Sadowski et al., ICSE-SEIP 2018, 9M changes).
- **Meta**: one accept lands a diff, and RADAR now auto-approves 60.31% of eligible diffs and has landed 331K+.
- **Amazon**: one team member's approval is "the last manual review and approval that a code change receives".
- **DORA** measured this exact shape — approval from an external body "such as a change advisory board or a senior manager" — and found respondents **2.6x more likely to be low performers**, with no evidence it lowers change failure rate.
- **No organisation** in the study runs a panel-plus-adversary-plus-re-review loop, and none has a non-convergence rule. They reach the same place by bounding **size** up front: Google's measured median change is 24 lines.

The panel-plus-adversary *architecture* is separately well supported for non-human reviewers (Khan et al., ICML 2024: adversarial debate took a judge from 48%→76%). It is the **emptiness stop** that has no support, not the panel.

[...]

PR #853, opened 2026-09-10
The rules the merge study earned, and the cap that made me delete controls to fit

muhammadbayoumi:
His ruling, 2026-09-10, after a study comparing this repository's merge rules against documented practice at Google, Meta, Microsoft, the continuous-delivery companies and the DORA evidence base, and then an audit of `CLAUDE.md` as an instruction file for the sessions that read it.

**Documentation-only, and therefore exempt from the merge loop — but not from the pass this change itself introduces.** Two review passes ran on it. Both found a `must fix`. Details at the end.

## What the study established

Our loop is **stricter than any of them on one axis, and missing what all of them treat as load-bearing.**

- Google's published terminal condition is the opposite of ours: approve "once it definitely improves the overall code health of the system being worked on, **even if the CL isn't perfect**". Measured: over 80% of changes resolve in at most one comment iteration, median review latency under four hours (Sadowski et al., ICSE-SEIP 2018, 9M changes).
- Meta lands a diff on one accept; RADAR auto-approves 60.31% of eligible diffs and has landed 331K+.
- Amazon: one team member's approval is "the last manual review and approval that a code change receives".
- DORA measured our shape — approval from an external body "such as a change advisory board or a senior manager" — and found respondents **2.6x more likely to be low performers**, with no evidence of a lower change failure rate.
- Nobody runs a panel-plus-adversary-plus-re-review loop, and nobody has a non-convergence rule. They bound **size** up front instead: Google's measured median change is 24 lines.

What survives here is the panel-plus-adversary **architecture**, which has the best available evidence for non-human reviewers (Khan et al., ICML 2024: adversarial debate took a judge from 48% to 76%). What does not survive is the emptiness stop, which has no support anywhere — and the recommendation to change it is **not in this PR**: #841 measures the loop's own yield first, from the 348 branch histories that `delete_branch_on_merge = false` has preserved.

## The rules this adds, and what each prevents

| rule | what it prevents |
|---|---|
| **The report names each reviewer and its verdict; a report that cannot is a failed pass, not an empty one.** | #671 records two panels that reported `COMPLETE: true` with reviewers dead. The whole gate rests on "a review returned nothing", and that sentence is currently unverifiable. |
| **A finding advances only with a demonstration** — a failing test, a quoted line whose own text shows the defect, or a counted query; undemonstrated it drops one rank and is filed. | An invented finding costs a full re-review cycle, and it is the likeliest error of a non-human reviewer. |
| **Split before the review, not after** — over 1,000 changed lines outside `tests/` and fixtures. | Measured over the last 120 merged PRs: median **506** changed lines, **462** of them additions, **59% over 400**. A ceiling at 1,000 fires on the worst quarter; at 400 it fires on 59% and is ignored within a week. The five-pass rule stays — it catches non-convergence, which a line count cannot. |
| **CI green on the head it has right now — a recorded green expires.** | Measured today: #821 was handed over as "CLEAN, 14 of 14"; that was true of `fbc98dae`, and at `b029e845` `test` and `migration-authority` both fail. Three of four reviewers repeated the stale green. |
| **A `VERSION` bump is his decision — merging one ships the engine to every installation.** | `tag-the-release.yml` fires on `workflow_run` from green CI on `main`, tags the SHA CI ran on, and dispatches `release-engine.yml` with `dry_run=false`. No `environment`, no approval. The file said the gate was the tag, which is downstream of the real one, so a session never looks. |
| **A running engine is not your edit.** | Python imports once; an engine started before the change keeps serving the tree it loaded, and `__file__` is correct throughout. |
| **A wrong number is diagnosed at the stored row first**, and never corrected downstream. | The warehouse is append-only, so a value corrected in the exporter or the Sheet leaves the row wrong and the next crawl serves it again. |
| **A new source, and `robots`, are his.** | `crawl_obey_disallow` ships at `0` (`scrapex/settings.py:84`) and `robots` defaults to it (`scrapex/config.py:373`). The only rule here whose wrong action is irreversible outside this repository. |
| **A local green ran no browser test**, and `node --test` runs suites `pytest` never reaches. | Four `importorskip("playwright")` suites report *skipped*, not failed; 27 `.test.mjs` files are invisible to pytest entirely. |
| **A capability and its control ship in one change.** | Both orders of a split violated an existing rule, and no rule paired them. |
| **Take the PR before merging it** — `gh pr edit <n> --add-assignee @me`. | "One session merges" was a claim no session could check, and the mechanism is free. |
| **On a branch that is not yours, findings go back to its author.** | The merge gate said "fix what survives, push", against "one session merges". This PR is the first application: #821 and #828 were gated and handed back, not edited. |
| **A red check is never argued past** — re-run once, file the flip against the named test. | The entity most likely to reason its way past a red run is a session under time pressure. |
| **A reviewer reads the diff and the repository**, and the PR body is a claim to check, never a defence. | Author and reviewer share a model family; that correlation is bought mainly with shared context, so withholding intent is the free half of decorrelating them. |
| **`normalize` and connector changes prove themselves against the frozen corpora.** | Nothing measured the product. A wrong price is not a property of the code — it is a relation between the code and a website, and no review or mutant finds it. |
| **This file is no longer exempt from itself** — one pass, not the loop. | Every amendment to the merge rules travelled the one path with no gate. PR #595, which introduced the gate, merged 3.8 minutes after opening with 0 reviews, declaring its own exemption. `CLAUDE.md` also instructs every session to *delete*, so a session finding a control inconvenient could remove it through that path. |

## The 150-line cap is removed, and it is removed because it did damage

Reaching 150 in the first draft of this change, I deleted three controls to make room: the non-convergence rule, `not a status to keep` (which refuses the "it has always had that permission" argument for an existing over-broad grant), and `Thoughtfulness over speed`. A number at the foot of the file made deleting rules the cheapest way to comply. That is a defect in the control, not in the session.

It is replaced by three duties, none of them a hard stop: **never delete a control to make room**; **delete what stopped being true before adding**; and past 150 lines, the next change here **starts by pruning what no longer changes an action**, moving detail to a comment beside the code or to a `.claude/` skill if nothing can go.

**The file is 184 lines, up from 150.** That is honest growth — nine `must fix` rules, each verified at `file:line` in this repository. If the size is unacceptable the answer is relocation, not deletion: the Review section is 30 lines and the tools table 15. One adviser recommended relocating them; the other killed the same proposal. **His call, and not folded in here.**

## What was rejected

The two adversaries killed **31 of 53** proposals: a `## Checks` section listing the CI commands, moving the Review section into a skill, moving the tools table, a `## New work` section (+9 lines), replacing the line cap with a word cap, naming the fast test tiers, naming the warehouse pointer file. Each failed the same test — no session behaves differently, or the information is one cheap command away.

Also rejected, from the study: `required_approving_review_count = 1` (one account, and GitHub forbids self-approval, so `main` would become permanently unmergeable); alerting and paging (no recipient — the panel is the only surface, so a post-merge control must be a fail-closed action, not a notification); automatic engine rollback (migrations are forward-only and the pre-push hook refuses a tag behind the warehouse — the revertible unit here is the publish); the DORA four keys (1 revert in 661 commits, 0 deployments, no users); and a generated `R-`/`OP-` traceability matrix (that corpus is frozen by this file's own rule).

## The two passes this change ran on itself

**Pass one**, on the first commit, found one `must fix` and six `should fix`, all mine. I had deleted the non-convergence rule and replaced it with the size ceiling, which is a *different control* — the first fires on a signal the review produces, the second on a count taken before it. The demonstration rule demoted a `must fix` two ranks to `optional`, straight past `should fix`; architecture findings are the hardest to demonstrate, so the dimension most likely to produce an expensive finding was the one whose findings would have been easiest to drop out of the gate. And the frozen-corpus rule pointed at a harness that only half exists.

**Pass two**, on the rewrite, found six more. Three were factual claims I had just written and had not checked myself: `the extension ships no third-party JS` (it ships `extension/vendor/tabulator.min.js`, 445,987 bytes, copied into the build by `release-extension.yml:91`); `holds the write lock and the device lease` (`scrapex/lease.py` is imported by one test and no runtime path); and `the Apps Script staging inbox CI posts to` (no workflow posts to the funnel — `scrapex funnel-test` does). The other three were controls the rewrite had dropped: `Reading a copy is a different question and it passes`, `Put a shared thing where the repo already puts shared things`, and the unconditional `ask before editing`. All corrected in the last commit.

Related: #841 · #842 · #843 · #844.

Generated with [Claude Code](https://claude.com/claude-code)

---

## Update — his ruling: relocate, and the file lands at 150

The Review section (30 lines) and the tools table (15) are now skills:
`.claude/skills/review/SKILL.md` and `.claude/skills/record-it/SKILL.md`.
**`CLAUDE.md` is back to 150 lines** with nine more rules in it than `main` has.

**What moved is the procedure; what stayed is the duty.** A skill is invoked, not loaded,
so a rule that must always bind cannot live only in one. `CLAUDE.md` keeps, in six lines,
that green is not mergeable, that the green must be the head's, that clean means no
*must fix* and no *should fix*, and that documentation is exempt except itself — then
points at the skill for the dimensions, the per-issue format, and the gate's own rules.
The tools section keeps the prohibition and points at the other skill for the commands.

Every moved phrase was checked as present in its new home, wrap-tolerantly, before the
old text was deleted: the four dimensions and what each looks for, `report nothing rather
than pad it`, `two or three options including "do nothing"`, effort/risk/blast
radius/maintenance, the reviewer-count rule, the demonstration rule, the 1,000-line
split, non-convergence, all eight table rows, and the frozen-corpus note.

The `review` skill also gains what the merge gate had only in practice: the adversary
must record **every kill with its reason**, because its own success metric is killing
findings, so the kill list is the only record of its false negatives.


muhammadbayoumi:
Two notes on rules in this diff, both demonstrated rather than argued. Measured today from
#832, which ran the loop this PR is rewriting.

## 1. The mechanism the green rule names inherits the duplication — and "14 of 14" is the artifact

The rule:

> **Check CI on the head the branch has right now.** A recorded green expires — a commit can
> land after the green that produced it. `gh pr checks <n>` and `gh pr view <n> --json headRefOid`.

`gh pr checks` today returns **two rows per check for five of the nine checks**, identically
named, because `on: push` is unfiltered on a branch with an open PR (#851's #646). Counted just
now:

```
#821 -> rows: 14   distinct: 9
#832 -> rows: 14   distinct: 9
   2 test · 2 scope · 2 migration-authority · 2 lint · 2 contract-parity
   1 CodeQL · 1 Analyze (actions) · 1 Analyze (javascript-typescript) · 1 Analyze (python)
```

This PR's evidence for the rule is that #821 "was handed over as *CLEAN, 14 of 14*". **14 is the
duplicated row count, not a count of checks** — there are nine. So the number a session reads
straight off the named command is an artifact of the CI configuration, and #851/#646 will change
it to 9 for every PR without anything about the checks themselves changing.

That does not weaken the rule — the stale green at `b029e845` is a separate and real fact — but
it does mean:

- a session that verifies by **tally** can match "14 of 14" against a different head and a
  different run pair and conclude nothing changed
- `gh pr checks` **exits non-zero while either copy is pending**, so a genuinely green head reads
  as not-green for as long as the slower duplicate lasts. Two of yesterday's reads on #832 hit
  exactly that, on a head whose checks were all green.

**Suggestion, in the rule's own spirit:** say what identifies the green rather than how much of
it there is — the run's head SHA against `headRefOid`, not the row count. A rule that quotes a
number will age with the CI config, and this one is about to.

## 2. "A panel whose reviewers died" has a third failure mode: blocked, not dead

The rule:

> **The report names each reviewer and its verdict.** A report that cannot is a **failed pass,
> not an empty one** — a panel whose reviewers died reports the same silence as a panel that
> found nothing.

A dated instance from a different direction than #671's two. Yesterday, investigating what
became #832, a ten-agent run had **three agents blocked by a safety classifier** — refused
before they ran. Not crashed, not timed out, not empty: never started. The aggregate came back
structurally identical to "these three dimensions found nothing", and two of the three were the
highest-stakes dimensions in that run (committed-identity leakage, and secrets in the
repository).

I only knew because the runner listed the blocks separately from the results. A report that
listed dimensions and verdicts — exactly what this rule asks for — would have shown two
silences and no reason for them.

**Suggestion:** "names each reviewer and its verdict" reads as satisfied by a reviewer that
returned nothing. Naming the third state closes it — each reviewer named with **a verdict or a
recorded failure to run**, blocked and refused counting as failures. One clause, and it covers
died, timed out, and refused with the same words.

Nothing requested on #832's account; it is unaffected either way.

muhammadbayoumi:
Both notes are right, both are applied in `46aa760`, and the second one found a failure mode the rule as written could not have caught.

## 1. The green is identified by head SHA, not by a tally

Reproduced before changing anything — and the distinct count is lower than yours, because `Analyze` reports three variants under one name:

```
#821  rows: 14   distinct names: 7
#832  rows: 14   distinct names: 7
#853  rows: 10   distinct names: 5
```

So "14 of 14" was doubly an artifact: fourteen is a row count, and the doubling is #646's, which #851 removes. A rule that quotes it would have aged out on the merge of another open PR.

`.claude/skills/review/SKILL.md:42-47` now reads:

> **Identify the green by head SHA, never by a tally.** A recorded green expires — a commit can land after the green that produced it. And a row count is an artifact of the CI config, not a number of checks: `gh pr checks` prints a row per run, so an unfiltered `on: push` doubles most rows, and it exits non-zero while any duplicate is still pending, which reads as red on a head that is green. Take `headRefOid`, then read the check-runs for that exact SHA and require each context by name.

The non-zero-exit half is the part I would not have found: a genuinely green head reading as red for the duration of the slower duplicate is a false RED, and every guard failure this repository has recorded runs the other way. It is also the reason to require contexts **by name** rather than by count — a name list survives the configuration change that is about to halve the rows.

## 2. Blocked is the third state, and it is worse than dead

Applied at `.claude/skills/review/SKILL.md:62-66`:

> **The report names each reviewer and its verdict**, and for one that returned no verdict, which of three it was: it died, it timed out, or it was **refused before it started**. A report that cannot is a **failed pass, not an empty one** — all three read as "this dimension found nothing", and a refusal never reaches the runner at all, so the dimensions most likely to be blocked are the ones that touch secrets and identity.

The last clause is the part your instance adds that #671's two did not: a refusal is not randomly distributed. Two of your three blocked agents were the highest-stakes dimensions in that run — committed-identity leakage, and secrets in the repository. A death is noise; a refusal is **correlated with the subject matter**, so the silence lands hardest exactly where the review was most needed. A rule that only says "name the reviewers" would be satisfied by a report listing three silences with no reason for any of them.

This session hit the same wall an hour ago from the other direction: the merge gate on #830 and #832 completed all eight dimension reviewers and lost **both adversaries** to a session limit. The workflow returned `{"verdicts":[],"detail":[]}` — structurally identical to "the adversary attacked and found nothing". It was reported as a failed pass and re-run, which is the rule working, and it is the third distinct cause in as many days: died, refused, limit.

## 3. What I found while applying yours, which is mine and worse

**Moving the Review section into a skill carried it out of the only set whose citations are verified.**

`tests/test_the_documents_cite_what_they_claim.py:90` sets `DOCUMENTS = ("CLAUDE.md", "README.md")`, and line 178 puts `.claude` in the `skip` set for the file-index walk. So every `file:line` in the Review section was checked while it lived in `CLAUDE.md`, and the moment it moved it was checked by nothing — a governed document, silently unwatched. That is this repository's signature defect committed by the change that adds three rules against it.

Both `SKILL.md` files are now in `DOCUMENTS`, with the reason recorded beside them.

**And I proved the guard fires rather than reporting it green.** The two skills carry zero `file:line` citations today, so widening `DOCUMENTS` is vacuous until one appears — exactly the shape the tests dimension is told to hunt. Mutation, run and then reverted:

```
$ printf '\nMutation probe: tools/panel_harness.py:999999\n' >> .claude/skills/review/SKILL.md
$ python -m pytest tests/test_the_documents_cite_what_they_claim.py -q
E  .claude/skills/review/SKILL.md:77 cites tools/panel_harness.py:999999 but that file has 902 lines
FAILED test_every_citation_names_a_line_that_exists
```

Probe removed, suite back to `13 passed, 1 skipped`.

---

Head is now `46aa760`. The PR stands at four files: `CLAUDE.md`, the two skills, and this test.

Issue #864, opened 2026-09-10
Code scanning and branch protection both stopped at 03:09 today, and a green badge does not say so


muhammadbayoumi:
**The merge session's own standard procedure produces a head that CodeQL never analyses, and the result reads as green.**

Found while gating #832. Not fixed; recorded because it will recur on every PR that is updated before it is merged.

## Measured

`gh pr update-branch 832` created a **merge commit**, `28286aca` (`parents=2`, "Merge branch 'main' into claude/auto-logged-account-tool-4318c9"). Nineteen minutes later:

```
gh api repos/.../actions/runs?head_sha=28286aca
  CI  event=pull_request  completed success
  CI  event=push          completed success
  (nothing else)

gh api repos/.../commits/28286aca/check-runs
  10 runs, 0 pending, 0 failing
  names: contract-parity, lint, migration-authority, scope, test   (each twice)
```

**No `CodeQL` row and no `Analyze` rows.** The commit it replaced, `2e4821f7`, had them and they passed:

```
gh api repos/.../commits/2e4821f7/check-runs
  CodeQL=success, Analyze (python)=success,
  Analyze (javascript-typescript)=success, Analyze (actions)=success
```

So the update **removed** four analysis rows from the head that is about to be merged, and left a result that counts as ten-of-ten green.

## Why it matters more than one PR

This is #682's shape — a check that never attaches, and absence reading as a pass — arriving through the procedure the merge session is *supposed* to follow. `strict: true` would have forced exactly this update on every PR after the first of a batch, so under working branch protection **every merge but the first would land a head with no CodeQL**.

It is also invisible to the obvious command: `gh pr checks` prints ten passing rows and exits zero.

## What is NOT the finding

The code itself was analysed. Proven by `git diff` rather than assumed:

```
git diff --stat 2e4821f7 28286aca   ->  .github/workflows/ci.yml | 49 +   (only #830's change)
git diff --stat 90c4b36a 28286aca   ->  the three #832 files | 79 +      (only #832's change)
```

Every line in the merged head was analysed either on `2e4821f7` (the extension and test changes) or on `b0ecc289` and then `main` (#830's `ci.yml`). The delta the merge introduces is empty. **The gap is in what the check-run record can prove, not in what was scanned** — and a gate that cannot prove it is the thing this repository keeps being bitten by.

## Worth establishing

[...]

PR #869, opened 2026-09-10
A shallow fetch is repo-wide, and it cost this session the merge

muhammadbayoumi:
One line merged into `CLAUDE.md`'s existing checkout trap. **Line count unchanged at 150** — the file's own cap, so this is an add-by-replacing rather than an addition.

## What it records

A review agent on #851 tested the shallow-clone hypothesis for the pre-squash fixture, restored the repo files it had touched, and left `.git/shallow` grafted at the branch tip.

Measured consequences, in order of discovery:

| | |
|---|---|
| `git rev-list --count HEAD` | **6**, against 669 for `origin/main` |
| `git merge-base origin/main HEAD` | *empty* |
| `git merge-tree --write-tree origin/main HEAD` | `fatal: refusing to merge unrelated histories` |

It took `git fetch --unshallow` (669 commits back) plus a rebase onto `main` to land #851 at all. The trap is not that a session shallowed its own checkout — it is that **`.git` is shared across every worktree under `.claude\worktrees\`**, so restoring the files restores nothing about the graft.

## And it demonstrated the half already written down

`ci.yml:167-176` names two guards that "BOTH SKIP rather than fail on a grafted clone". A whole-suite run that overlapped the shallow window reported **23 skips instead of 21**, and the two extra were exactly:

```
2  no git history here — the comparison cannot be made
```

from `tests/test_the_privacy_policy_is_true.py` and `tests/test_version.py` — the two the comment names.

**What it did *not* do is fire the pre-squash guard** #830 added, because that fixture reads `origin/main`, whose history stayed complete. So "shallow" is not one condition, and the guard that exists covers one form of it. Worth knowing before anyone treats that step as coverage for grafted clones generally.

## Why it belongs in the existing trap rather than beside it

Same root: several live checkouts sharing one `.git`. The paragraph already tells a session that a misdirected *edit* is invisible to `__file__`; this adds that a misdirected *fetch* is invisible to `git status`. Both are "the state you did not think you were sharing".

Documentation only — exempt from the merge gate by `CLAUDE.md`'s own Review section. The guards that read `CLAUDE.md` are green (`test_the_documents_cite_what_they_claim.py`, `test_the_ruling_matches_the_code.py`).


muhammadbayoumi:
## Merge session: this rewrites the same paragraph as #853, and it keeps a path that does not exist

**The collision.** This PR and #853 both rewrite the *same* paragraph — the first of the two traps in `CLAUDE.md`. #853 is based on that paragraph too, and is 193 insertions across `CLAUDE.md` and two new skills. Whichever of the two merges second will conflict there, in exactly one hunk.

**The finding, which is why the order matters rather than being a coin toss.** The line this PR carries forward is measured false on this machine:

```
$ python -c "import scrapex; print(scrapex.__file__)"
C:\Users\sapac\Desktop\Claude\ScrapeX\.claude\worktrees\merge-session-2b3b8f\scrapex\__init__.py

$ ls "C:/Users/User01/source/repos/ScrapeX"
ls: cannot access 'C:/Users/User01/source/repos/ScrapeX': No such file or directory
```

`scrapex` is pip-installed editable against **this worktree**, and the directory the rule names **does not exist**. So the sentence "`scrapex` is pip-installed editable against `C:\Users\User01\source\repos\ScrapeX`" sends a session to check a path that is not there, and tells it the install lives somewhere it does not. A session that trusts it will conclude its edits are going to the wrong place when they are going to the right one — the opposite of what the trap exists to prevent.

#853 replaces the hard-coded path with the question itself:

> `scrapex` is pip-installed editable against a checkout that may not be this worktree — `python -c "import scrapex; print(scrapex.__file__)"` says which.

That is machine-independent and it stays true when the install moves, which is what just happened.

**Your sentence is new knowledge and should survive.** "A shallow fetch is repo-WIDE: restore `.git/shallow` too, or the next session finds no merge base with `main`" is not in #853 and nothing else in the file says it. It is exactly the shape the file asks for — an instruction that changes what a session does, paid for by an incident rather than argued.

**Proposed order, and it is a proposal, not a ruling.** #853 merges first — it is older, it has had two review passes and the owner's own two notes applied, and it corrects the false path. Then this rebases onto it and keeps only the shallow-fetch sentence, appended to the paragraph #853 leaves. That is a one-line diff instead of a five-line one, and it does not have to re-argue the path.

If you would rather go first, say so and I will hold #853 and rebase it onto yours instead — the merge session does the rebase either way, and it picks no behaviour in either direction. What should not happen is the two landing blind, because the loser's rebase silently reinstates whichever version of that paragraph the resolver happens to keep.

---

Two things about this repository that changed at about 03:09 this morning and are worth knowing before you read any badge on this PR:

- **Branch protection is gone.** `branches/main --jq .protected` is `false` and `branches/main/protection` returns `403 "Upgrade to GitHub Pro or make this repository public"`. There are no required contexts and `strict` is not enforced, so `CLEAN` on any PR now reflects the absence of a gate rather than the satisfaction of one.
- **Code scanning is off.** `code-scanning/default-setup` returns `403 "Code scanning is not enabled for this repository"`, and the last CodeQL run of any kind on any head was 03:09:40Z. No PR opened from now on can carry a CodeQL row.

Both are recorded in #864. Neither is anyone's defect here; they are the owner's to restore.

muhammadbayoumi:
Superseded by a guard rather than a rule.

`main` moved under this branch: #853 rewrote `CLAUDE.md`, and two of its new rules change the answer here — **documentation-only changes are exempt from the merge gate "except this file"**, and past 150 lines a change to it "starts by pruning what no longer changes an action... if nothing can go, move detail to a comment beside the code or to a `.claude/` skill". The file is at exactly 150, and nothing in it has stopped being true, so a one-line addition would have had to delete a live rule to fit.

It also belongs beside code by that file's own test: `tests/test_the_workflows_check_out_enough_history.py` already exists for "a job that runs pytest on a grafted clone turns two guards into silent skips". The shallow trap is the same knowledge and changes for the same reason — it is the half CI cannot see.

Replaced by a test in that file, which fails on a grafted checkout instead of asking a session to remember. Measured against a real shallow clone: `false`/rc 0 passes, `true`/rc 0 fails, rc 128 (no repository) skips.

Issue #876, opened 2026-09-10
A killed-engine test fails under full-suite load and passes 3/3 alone: the assertion needs a bounded wait


muhammadbayoumi:
`test_a_killed_engine_does_not_leave_a_job_claiming_to_run` failed once in a full
suite run and passes 3 out of 3 when run alone. Second test found this way.

## Measured

```
full suite (3,948 passed, 2 failed)   FAILED
alone, run 1                          passed
alone, run 2                          passed
alone, run 3                          passed
```

The failure:

```
AssertionError: after a crash and a restart the job still says 'running'.
_source_is_busy reads this, so SLOWSHOP is now blocked from every future crawl
and nothing anywhere says why
assert 'running' not in {'preparing', 'resuming', 'running'}
tests/test_the_engine_survives_being_killed.py:266
```

## Why it is timing and not a defect in the code under test

The test starts a real engine as a subprocess, posts a crawl against a
deliberately slow shop, kills the process where it stands, then starts a second
engine on a new port and expects its orphan reconciliation to have cleared the
job. Every step is a process boundary and a clock: under the load of a full suite
the restart's reconciliation and the assertion race each other.

Its own earlier assertion says the same thing about the other end of the window —
"the job settled before the kill landed, so this run proved nothing about a
crash — the shop is answering faster than it is meant to". So the test already
knows it has a timing window; this is the far edge of it.

## What was nearly recorded instead

A first comparison ran it once with a change stashed and once without, saw pass
then fail, and concluded the change caused it. **One run per side is not a
measurement for a test that starts and kills real processes**, and that
conclusion was wrong — three runs on the same tree pass. Worth writing down
because the same shortcut was taken on #773 and corrected the same way.

## Options

1. **Wait on the condition rather than assert immediately** *(Recommended)* —
   poll for the job to leave `IN_FLIGHT` with a bounded deadline, and fail with
   the elapsed time in the message. The test's subject is "reconciliation happens",
   not "reconciliation has already happened by this line". Effort: small. Risk:
   a real regression that made reconciliation slow but not broken would now pass
   until the deadline is exceeded, which is the correct trade for this subject.
2. **Give the second engine longer to start before asserting.** A fixed sleep;
   cheapest, and it is the shape this repository's own `settle_view` comment
   warns about — a wait with no visible cause is a wait the next reader deletes.
3. **Mark it as allowed to be retried.** Hides the window rather than closing it,
[...]

Issue #868, opened 2026-09-10
Eight of the engine's fourteen storage routes have no panel caller, so by R-81 the product does not have them


muhammadbayoumi:
The engine serves fourteen storage and backup routes. **The panel calls four.**
`R-81` says a capability with no panel control has no control, so eight of the
owner's storage capabilities are unreachable from the only interface he uses.

Measured on `main` today:

| route | panel callers |
|---|---|
| `GET /api/storage` | 9 |
| `POST /api/bundle` | 5 |
| `GET /api/bundle/archive` | 3 |
| `POST /api/storage/integrity` | 2 |
| `POST /api/storage/restore` | 2 |
| `GET /api/bundle/panel-pack` | 1 |
| **`POST /api/storage/backup`** | **0** |
| **`POST /api/storage/check-move`** | **0** |
| **`POST /api/storage/compact`** | **0** |
| **`POST /api/storage/export`** | **0** |
| **`POST /api/storage/move`** | **0** |
| **`POST /api/storage/open-folder`** | **0** |
| **`POST /api/storage/repair`** | **0** |
| **`POST /api/storage/start-fresh`** | **0** |

## Why this is not a tidy-up

Every one of the eight is a real operation with a real reason to exist, and two
of them were needed *today* by the owner and could not be reached:

- **`open-folder`** takes `which` from `{database, backups, exports}` and shows
  that folder in the file manager — deliberately never a path from the page. It
  is exactly what "put the downloaded bundle here" needs, and there is no button.
- **`repair`** and **`compact`** are the two things to do about a warehouse that
  `POST /api/storage/integrity` has just reported damaged. The panel can now
  *find* the damage (#819's card) and can do nothing about it.
- **`move`** and **`check-move`** are how a warehouse changes disk. The owner has
  26.1 GB free on the drive holding a 2.15 GB warehouse and 12 GB of copies.
- **`start-fresh`** renames rather than erases, and is the honest way to begin
  again. Reaching it through no control means the alternative is deleting files
  by hand.
- **`export`** and **`backup`** overlap with what the bundle path does, and that
  overlap is worth deciding rather than leaving both unreachable.

## The pattern, which is the point

This is the fourth time in two days that a capability turned out to exist with no
caller: `POST /api/storage/restore` (fixed in #819), `bundle.verify()` (called
for the first time in #808), `bundle.unpack()` (still uncalled), and now eight at
once. The engine is ahead of the panel by a wide margin, and `R-81` means that
[...]

Issue #810, opened 2026-09-09
The upload dies with the panel window, so no schedule is possible: move the transfer to the native host


muhammadbayoumi:
**PHASE 1 · P0 for the RPO.** The upload lives in the side panel's document and
dies with the window. That makes the owner's chosen **RPO of 24 hours**
unreachable: nothing can run on a schedule.

Measured today, the first successful backup: build 110 s, upload 830 s, whole
press **942 s at 0.75 MB/s**. Fifteen minutes and forty-two seconds during which
the window must stay open. Four earlier presses produced correct archives and no
upload at all.

## What moves

The transfer moves to `scrapex/native.py`, which the panel already starts and
already trusts with `START_ENGINE`, so this adds no component to the
architecture. It owns:

- **A queue that survives a restart.** A point half-uploaded resumes from its
  last confirmed chunk rather than starting over — the Drive session is already
  resumable, and the chunk store makes resumption addressable.
- **Fan-out to every enabled destination**, independently. One destination down
  does not fail the backup.
- **The keys**, read from the OS keystore. The panel keeps holding none.

The panel keeps what it should have kept all along: policy, a **Back up now**
button, and a health card. It becomes a control plane and stops being a data
plane.

## And what this does NOT do

It does not explain #788 — four direct presses reading 0, one succeeding through
a measuring pass-through, cause still unknown. Moving the transfer out of the
browser document **makes that unknown irrelevant to the data path rather than
fixing it.** The distinction has to stay written down: an unknown that no longer
bites is not a solved unknown.

## Options

1. **Native host owns the transfer, with a durable queue** *(Recommended)* — the
   only option that delivers a schedule, and it reuses the one component already
   licensed to run outside the panel.
2. **Keep it in the panel and warn the user not to close the window.** Free.
   Leaves the recovery point at "whenever he remembered", which is measurably
   1 success in 5 attempts.
3. **A Windows service.** Strongest scheduling guarantee; a new privileged
   component, and it breaks "the panel is the operating system" by introducing
   something that starts work without being asked.
4. **Do nothing.** RPO stays unbounded.

muhammadbayoumi:
## The owner decided, 2026-09-16: **not option 1**, and not option 2 either

Asked in his own terms — *do you want a backup to happen while you are not there, with
Chrome closed?* — he answered: **the panel window may close, Chrome stays open.**

That is a third shape this issue did not list, and it is narrower than option 1 in exactly
the way that matters here:

[...]

PR #877, opened 2026-09-10
A copy is opened before it is trusted, and a bundle from Drive becomes one

muhammadbayoumi:
Two commits, and the second is why the first mattered today. Milestone #30.

**The owner hit this for real this morning**, on another machine, with months of
crawling in a Drive zip he could not use. There was no path from a bundle in
Drive to a working warehouse, and the reason was not one missing feature:

- `bundle.unpack` had **no caller anywhere in the product**. It extracts with
  path-traversal protection and verifies every file against the manifest.
- `POST /api/storage/open-folder` had **no panel caller** — the control that
  answers "where do I put the file" (one of the eight in #868).
- And nothing turned an unpacked database into something `restore` would accept.

## `a4c36f19` — a copy is opened before it is trusted (closes half of #808)

Nothing had ever opened a backup to find out whether it could be opened.
`verifyLatest` asks Drive whether an object of the right size is there, which
proves an object exists and not that it restores.

`POST /api/storage/integrity` learns an optional `backup_path`. **The same route,
not a new one** — measured: the contract baseline tracks `METHOD /path` strings
and the capability baseline nine capabilities, none of them storage. Same
knowledge, same reason to change.

**The reply echoes the file it read.** An older engine ignores an unknown field
and answers about the live warehouse — a healthy verdict about the wrong file,
arriving exactly where the reader wants reassurance about this copy. The echo
tells understanding from politeness without a version gate that would drift.

Every table is counted, because `PRAGMA quick_check` passes on a database that is
intact and **empty**. Measured: all 67 tables of a 2,148,061,184-byte file cost
123 ms against that file's 6,953 ms health check. An unreadable table returns
`-1` rather than being dropped.

The verdict appears **inside the restore confirmation**, so the weakness #808's
own brief named — a check nobody runs reports nothing — is closed structurally.

## `f10a41bd` — a bundle becomes a copy this product already restores

`POST /api/storage/adopt-bundle` takes a **name, never a path**, for the reason
`open-folder` states: this listens on a loopback port every tab can reach. The
engine resolves its own backup folder, accepts only a zip it listed there, checks
there is room for a second full copy **before** it starts, and writes the unpacked
database under this product's own naming — which is what `list_backups` reads and
what `restore` checks against.

So the bundle arrives at **every guard downstream** instead of at a special case
that would have to be trusted: listed as a copy, health-checked before the
confirmation, and put in place by the same question with the current warehouse
moved aside rather than deleted.

**It does not restore.** Unpacking and replacing a warehouse are two decisions.

`fetchLatest` was **not** used, though it exists and downloads: it holds the
archive as a Blob in the panel's document, and 625 MB in a side panel is what
came back 0 bytes on 2026-09-03. The engine unpacks; the panel carries no bytes
and sends no path.

## Version 0.4.13 → 0.4.14

`R-35` requires it for a contract change, and
`test_the_version_moves_when_the_contract_does` **named the route and refused
it** — the fourth of this repository's own guards to catch me in two days.
Baselines regenerated by `python -m scrapex.cli export-version` (99 routes), not
hand-edited.

## Checks, on this head

```
eslint    0 findings
ruff      All checks passed
node      459/459 extension · 13/13 apps_script
pytest    3950 passed, 3 skipped, 0 failed
def test_ +9
```

**A recorded green expires** (CLAUDE.md, as of #853), so this is the state at
`f10a41bd` and CI's own run is the one that counts.

## Mutants

**#808: eight of eight. Bundle path: six of six.**

Three bundle mutants were skipped on the first run because my anchors were not
unique — **a mutant that never ran is not a mutant that passed** — and one
survived: the guard compared the live warehouse's *size* before and after, and
the bundle was built from that warehouse, so a copy of it back over the original
is byte-identical in length. It now writes a row **after** the bundle is packed
and asserts the row survives.

## Judgements taken in the open

- **The `open-folder` wiring rides along, and I judged it part of the job rather
  than an extra feature.** "Where do I put the file" is the first step of the
  path this PR exists to create; shipping the unpack without it would be a
  control whose precondition has no control. Say so if that reads as scope.
- **A new route was unavoidable here**, unlike #819 and the first commit. There
  is no existing route whose knowledge this shares — `restore` makes a copy live,
  `backup` copies the live one; neither unpacks. So the version moved, correctly.
- **`merge-warehouse` is still CLI-only.** For a second machine that already has
  crawl data, `restore` *replaces* — `cli.py:1175` says so — and the right
  command has no panel control. Filed as part of #868's shape; it is the one case
  this PR does not help.

## Found on the way, filed not fixed

- **#868** — eight of fourteen storage routes have no panel caller. This PR wires
  one of them.
- **#876** — `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` fails
  under full-suite load and passes 3/3 alone. I first attributed it to this change
  on one run per side, which is not a measurement; the issue records that.
- **The `ruff` hole was mine**: my own check script ran `eslint` and never `ruff`,
  so "lint clean" was half the truth on every Python change. The suite's own
  `test_the_linter_is_actually_installed_and_clean_here` caught it eighteen
  minutes later. The script runs both now.

## Not in this slice

Restoring straight from Drive without the manual download, and restoring a single
dataset with a diff — the rest of #806. And the weekly automatic verification of
#808, which waits on #810.

## `17639c3` — the merge gate's nine findings, every one of them

The pass above returned **two must fix and seven should fix** and did not push
them, because under CLAUDE.md what a pass finds on someone else's branch goes
back to its author. All nine are fixed here. Each fix was reverted on purpose
afterwards and the guard that should catch it was watched failing: **17 mutants,
17 killed.**

### must fix · a path was handed to a URI parser (`storage.py:805`)

Reproduced before anything was edited, on this machine, against a real database
at `…\Drive #2\harvest.db` holding 1,000 rows:

```
before, the folder holds: ['Drive #2']
row_counts ->             {}
direct sqlite3 count ->   1000
after,  the folder holds: ['Drive', 'Drive #2']      <- a database named `Drive`
```

`#` starts a fragment, so the filename truncated **and the whole `?mode=ro`
query went into the fragment**: not read-only, and SQLite created a new empty
database at the truncated name. `%` decoded into a path that does not exist and
raised `OperationalError` — a 500, since nothing catches it.

Now `Path(db_path).resolve().as_uri() + "?mode=ro"`, which percent-escapes both.
The same run after the fix reports `{'generic_page_snapshot': 1000}` and creates
nothing. Two tests, one per character.

The two PRE-EXISTING instances of the same pattern — `bundle.py:138` and
`databases/carry_over.py:274` — are **#925**, filed by the pass that found this
one. They are not touched here.

### must fix · the copy this check exists to catch was a 500 (`app.py:3276`)

`health` produced the right verdict — `unreadable`, with SQLite's own words —
and the next line asked the same file for row counts unconditionally, which
raised past the three handlers `webui/app.py` registers. The verdict was
computed, correct, and thrown away; the panel then took `checkTheCopy`'s catch,
said "Restoring it is still possible, unverified" and **re-enabled Restore on a
file that is not a database**.

`_counted()` is now the only way this route asks for counts, on both sides of
both branches. It is not a swallow: for the file the route checked, the reason is
`status` and `detail` in the same reply; for the live warehouse counted beside a
copy, it is that warehouse's own verdict on `GET /api/storage` — the card this
dialog opens over. The test zeroes a listed backup's first 4,096 bytes and
asserts 200, `unreadable`, `rows == {}`, `live_rows` still counted, and the
verdict recorded.

### should fix · the guard ignored the configured backup folder (`app.py:3259`)

`list_backups(app.state.db_path)` with no folder, against a panel list built
from `list_backups(path, backup_folder(conn, path))`. One function, two
arguments, two different sets — so every copy the page offered failed the guard
the moment the backups were not beside the warehouse. The route now resolves the
folder first, exactly as `adopt-bundle` beside it already did.

`storage.restore:1137` has the identical omission and is **#926**: it is
pre-existing, it replaces a live warehouse, and it deserves its own review rather
than riding along here.

### should fix · the offer and the echo were two spellings (`app.js:6286`)

Fixed at the source rather than at the comparison: `list_backups` emits
`str(p.resolve())`, so one spelling leaves the one function that decides what a
backup is. The panel's strict `verdict.checked !== snapshot.path` is then true
when it should be, and nothing else had to learn about normalisation.

The guard is end-to-end and needs no Windows: the configured folder is spelled
`…/here/../Backups`, the page is asked what it offers, and that exact string is
posted back and compared against the echo.

### should fix · a half-downloaded zip was "Internal Server Error" (`app.py:3720`)

Two refusals, because there are two places an archive can fail: reading its
central directory (`unpacked_size`, outside every try) and reading its contents
(`unpack`, inside a try that had a `finally` and no `except`). Both now answer
400 with the engine's own words, which is the only thing `backend.js` can lift
out of a body. Tested with half a real bundle's bytes, and with a zip whose
entry is named `../escaped.txt` — `unpack` refuses that correctly, with a
`ValueError` that used to reach the owner as a 500.

### should fix · `ok: True` over a verdict computed two lines above (`app.py:3757`)

`bundle.verify` checks the manifest's digests: that the zip arrived whole. It
says nothing about whether the database inside is one THIS engine can open — and
a bundle built by a newer engine verifies perfectly and then fails
`_warehouse_identity`. A second machine running a different engine is precisely
what this route exists for.

`ok` is the verdict's now, the sentence says which copy it became and what is
wrong with it, and `adoptBundle` draws the failing case in `err` rather than in
the success colour. The test swaps a stranger's database into a verified bundle
and asserts `ok: False`, the reason, and that the file is still listed so he can
find and remove it.

**It is not deleted.** Unpacking wrote the only local copy of what is in his
Drive; throwing it away on his behalf is a worse answer than naming it. Say so if
you want the opposite.

### should fix · the panel half had no test of any kind

Six new DOM tests over the card that shipped bare: the zips listed apart from the
copies with a count that agrees with the list, the empty folder that still says
what to put in it, `#db-open-backups` sending `{which: "backups"}` and no path,
an unpack that names what it became and re-reads the page, a refusal that carries
the engine's own words and gives the row back, and the `ok: false` answer that
must not be announced in green.

### should fix · the harness could not tell the two sides apart

`rows` and `live_rows` were both 12, so `assert "12" in said` passed whichever
`rowVerdict` read and swapping them was a surviving mutation. The live side is
340 now, one test asserts both numbers in the right order, and the branch that
matters — a copy that is intact and EMPTY beside a full warehouse — has its own
test on both sides of the wire. That test also asserts `#restore-confirm` is
still ENABLED, because it is: the copy passes every health check there is and the
sentence is the only thing between him and the loss.

`rowVerdict` gained one sentence while it was there. An empty live side used to
render as "This copy opens and passes its checks." — true, and silent about the
comparison it could not make, which is the empty-copy warning disappearing
without a word. It now says so.

### should fix · `unpacked_size` and two refusal branches had no test

`unpacked_size` is asserted against what unpacking actually writes, not against
itself. The free-space refusal and the verified-but-carries-no-warehouse refusal
each have a test, and both assert the staging tree is gone.

## One optional finding taken as well

`integrityScan`'s comment claimed 120,000 ms "covers a file roughly 40 GB",
derived from the WARM 5.9 s scan of a warehouse the engine already has open. This
deadline now bounds the copy check too, and a backup nobody has touched is read
cold: 42,874 ms on 2,148,061,184 bytes, about 20 s a gigabyte, so the real figure
is near 6 GB — 2.7x headroom at his warehouse today. **The bound is right and the
claim beside it was wrong**, and a wrong number in a comment two lines from the
right one is the thing this repository's rules exist to stop. Comment only.

The other optional findings — the dialog-open scan nothing can cancel, and
extract-then-delete moving the warehouse twice — are real costs on a
once-per-disaster path with no correctness failure. Not taken here.

## Judgements taken in the open, on `17639c3`

- **`made.health`, `made.path`, `made.bytes` and `made.from_bundle` still have
  no panel reader.** `ok` and `detail` now do, and `detail` names `made.name`.
  The rest are facts the route computed about what it wrote; narrowing a reply
  because today's only caller ignores three of its fields buys nothing.
- **`_counted` returns `{}` rather than raising or reporting per-file.** The
  alternative was an allow-list of health statuses in a third place, which is the
  drift this repository names by name: `health` and `row_counts` both answer "can
  this file be opened" by trying, and the third copy would be the one that goes
  stale.
- **No version bump.** No route was added or removed and the capability ledger
  is untouched; `0.4.13 → 0.4.14` in the commit below already covers this branch.

## Mutants, on `17639c3`

**17 of 17 killed.** Eight over the engine (the URI on each of `#` and `%`
separately, the unguarded counts, the folder-less guard, the unresolved offer on
both of its tests, the unprotected directory read, the uncaught `unpack`, the
unconditional `ok`, and the free-space guard), nine over the panel (the swap, on
both tests that should catch it; the silent "it passes"; the success colour; the
rows that are not drawn; the empty-folder sentence; the folder control that sends
nothing; the row left disabled after a refusal; and the page that is never
re-read).

The swap the pass called a surviving mutation is now killed by two tests
independently.

## `e9fc072` — two of my own tests, and the thing a local green cannot see

`17639c3` went red on CI, on two tests I had just written. Both spelled the
backup folder `.../here/../Backups` to prove `list_backups` normalises what it is
given, and `here` was never created:

```
Path("<missing>/../Backups").is_dir()   ->  True   on Windows
                                        ->  False  on POSIX
```

Windows normalises `..` lexically before it touches the disk; POSIX resolves each
component against the filesystem. So the folder read as absent, `list_backups`
returned `[]`, and both tests failed on the only platform that gates this
repository. This machine could not have told me: it is developed on Windows and
judged on Linux.

`here` is a real directory now. The spelling that makes the tests mean anything
survives — `str()` still carries the `..`, `resolve()` still removes it — and
both mutants they exist to kill were re-run against the fixed tests and are still
killed.

## Checks, on this head

```
lint             pass    20s
scope            pass     8s
contract-parity  pass    18s
test             pass  14m42s
```

Run 34653030827 reports `head_sha e9fc072`, `conclusion success` — the same head
this PR points at now, not a recorded green from an earlier one.

The local full run is not quoted here: I set `SCRAPEX_DATA_ROOT` to keep it off
his live warehouse (#910) and that alone turns
`test_a_killed_engine_does_not_leave_a_job_claiming_to_run` red — proved with
three runs and one variable, and recorded on #910, which names that variable as
the escape hatch. CI's run is the one that counts.

## How this merged, so nobody reads a gate that did not run

**MERGED ON THE OWNER'S INSTRUCTION WITHOUT THE SECOND REVIEW PASS.** He was told
what was missing, with the numbers, and said merge anyway. That is his call to
make; this section exists so the record says which half of the gate ran.

| the gate | state at merge |
|---|---|
| CI green on the head it has right now | **yes** — run reports `head_sha 2bdd627`, `conclusion success`, re-read immediately before the merge, on a branch already rebased onto `main` and 0 commits behind |
| a critical review over four dimensions, with an adversary | **not run** on `cebf5fa` + `2bdd627` — 810 lines of new code |

What that un-reviewed code is: the nine fixes above, their 26 tests, the harness
change, and one comment correction. What it has instead of a review: every fix
reverted on purpose and its guard watched failing — 17 mutants, 17 killed — and
CI green on the exact head.

Why the objection was worth making, recorded as evidence rather than as opinion:
the last review over this branch found 20 findings and 9 of them survived at must
or should fix, on a diff of comparable size; and two of the tests in `17639c3`
were wrong in a way only Linux could show, which CI caught and I did not. Neither
fact is an argument against his decision — both are the base rate a later reader
should weigh if something here turns out to be wrong.

Filed as an issue as well, following #890, which is the same situation from
#886.


muhammadbayoumi:
## Merge gate: **NOT CLEAN**. Not merging, and not fixing — this is not my branch.

CI is green on `f10a41bd` — `lint`, `scope`, `contract-parity`, `test`, required by name on that exact SHA, not by a row count. Green is not mergeable.

Four dimensions ran over the diff in a worktree at that head (environment proved first: `import scrapex` resolves there and `storage.list_bundles`/`row_counts`, symbols this change adds, are present). One adversary then attacked what they returned and read the change cold.

**8 findings survived and 12 were killed.** Under CLAUDE.md what a pass finds on someone else's branch goes back to its author, so this is the report, not a push.

---

### must fix (found by the cold read, missed by all four) · `scrapex/storage.py:805`

**row_counts builds a `file:` URI by string interpolation: a `#` in the path makes it open and CREATE a different database and silently report zero tables**

`conn = sqlite3.connect(f"file:{Path(db_path)}?mode=ro", uri=True)` interpolates an arbitrary filesystem path into a URI with no escaping. SQLite's URI parser treats `#` as the start of a fragment and `%` as a percent-escape. A path containing `#` therefore truncates the filename AND swallows the whole `?mode=ro` query into the fragment — so the connection is not read-only, SQLite creates a NEW empty database at the truncated path, and `row_counts` returns `{}` for a warehouse holding hundreds of thousands of rows. A path containing `%` decodes into a path that does not exist and raises OperationalError, which no handler catches (app.py:669-670) — a 500. This is new code in this diff and it runs on BOTH branches of the integrity route: app.py:3247 (the live warehouse, every press of Check integrity), app.py:3276 and app.py:3277 (the copy and the live warehouse, every open of the restore dialog). All four reviewers walked past it; one of them certified the opposite, calling row_counts "explicitly read-only" from the literal `mode=ro` in the string. The consequence is the one CLAUDE.md singles out — a confident wrong number. With `rows` and `live_rows` both `{}`, `rowVerdict`'s `biggest` is undefined and the panel returns "This copy opens and passes its checks." with Restore enabled: the empty-copy warning, the entire reason the counts were added, silently disappears. The wrong number is also written into the warehouse under `storage_copy_check`/`storage_integrity`.

**Demonstration**

```
Ran on this machine (PYTHONPATH=w877, fresh temp dir, no warehouse touched). A real database at `…\adv877c-fan3n5wc\Drive #2\harvest.db`, 4,530,176 bytes, 412,903 rows in generic_page_snapshot:
  before, folder holds: ['Drive #2', 'harvest.db']
  row_counts ->          {}
  direct sqlite3 count -> 412903
  after,  folder holds:  ['Drive', 'Drive #2', 'Drive #2\\harvest.db']
No exception, no warning — an empty dict, and a stray database file named `Drive` created by a function whose docstring and connection string both say read-only. The `%` case, same run, different folder: `row_counts RAISES: OperationalError unable to open database file`, which reaches the owner as "Internal Server Error". `health()` is unaffected in both cases because it uses `sqlite3.connect(str(path))` with no URI — so the page says healthy while the counts say nothing. Fix: `sqlite3.connect(Path(db_path).as_uri() + "?mode=ro", uri=True)`, with a test that puts the warehouse in a folder named `Drive #2`. bundle.py:138 carries the same unescaped pattern and is pre-existing — its own issue, not this diff's.
```

### must fix · `scrapex/webui/app.py:3276`

**An unreadable copy — the one this check exists to catch — crashes the route instead of being the finding health already made**

I opened every line. `storage_health(chosen, integrity=True)` at app.py:3269 produces the correct structured verdict, and then app.py:3276 runs `state["rows"] = storage_row_counts(chosen)` unconditionally — no branch on `verdict["ok"]`, no try. `row_counts`' per-table `except sqlite3.DatabaseError` (storage.py:817) sits INSIDE the loop; the first statement, `SELECT name FROM sqlite_master` at storage.py:806-808, is unguarded. The only exception handlers registered are for DatabaseUnavailableError, DatabaseMigrationError and DatabaseKindError (app.py:669-670), so sqlite3.DatabaseError becomes a bare 500. The existing damaged-copy test at tests/test_the_integrity_scan_is_asked_for.py:413 plants a foreign-key violation — the one damage that leaves the file readable — so nothing catches this.

**Demonstration**

```
Ran myself (PYTHONPATH=w877, fresh temp dir, no warehouse touched): took a listed backup, zeroed its first 4096 bytes.
  health(): {'status': 'unreadable', 'ok': False, 'detail': 'SQLite could not read this file: file is not a database', ...}
  row_counts RAISES: DatabaseError file is not a database
The verdict exists and is thrown away. Downstream the panel takes checkTheCopy's catch (app.js:6305), prints "This copy could not be checked ... Restoring it is still possible, unverified" and RE-ENABLES #restore-confirm — so a file that is not a database is offered for restore with a calm sentence. Fix: guard or skip row_counts when the verdict is not ok, and add the header-corruption test beside line 413.
```

### should fix · `scrapex/webui/app.py:3259`

**The copy-check guard ignores the configured backup folder; the sibling route in the same diff does not**

The line says what the finding says: `offered = {Path(entry["path"]).resolve() for entry in list_backups(app.state.db_path)}` — no folder. `list_backups` at storage.py:829 is `where = Path(folder) if folder else path.parent`, and storage_status fills the panel's list with `list_backups(path, folder)` where `folder = backup_folder(conn, path)` (storage.py:1637, 1612). `api_storage_adopt_bundle` — new in this same diff — resolves `backup_folder(conn, app.state.db_path)` first (app.py:3699). The route's own comment claims "one function decides what a backup is and the rule cannot drift away from the offer"; it is one function called with different arguments.

**Demonstration**

```
Ran myself: db at <tmp>/live/harvest.db, a correctly-named copy in <tmp>/Backups.
  list_backups(db, folder) : ['harvest.manual-20260101T000000Z.backup.db']
  list_backups(db)         : []
`offered` is empty, so every copy the panel offers fails `chosen not in offered` and takes the 400 at app.py:3266. DOWNGRADED from must fix: the reviewers called it live because backup_folder is "a real owner setting with its own control" — that control is on the ENGINE'S OWN WEB PAGE (scrapex/webui/templates/_storage.html:65-72), and `grep -n "backup_folder" extension/app.js extension/app.html` returns exactly one line, 6475, which only PRINTS the folder. By this repo's own rule the owner works from the panel only, so he has no way to enter the configuration in which this fires. It is a latent defect in new code that copied a pre-existing bug (storage.py:1115 in `restore` has the identical omission) — and it fails safe, with a refusal rather than a wrong answer. Fix the new route; raise restore's as its own issue.
```

### should fix · `scrapex/webui/app.py:3720`

**A truncated .zip from Drive — the likeliest state of a downloaded bundle — is an unhandled 500, not a refusal**

`needed = int(bundle.unpacked_size(archive) * FREE_SPACE_MARGIN)` is the first thing the route does with the file's contents and it sits outside every try. Two more uncaught paths follow inside the try at 3731, which has a `finally` and no `except`: `bundle.unpack`'s ValueError for an entry that escapes the destination (bundle.py:441-443) and OSError from copyfileobj. app.py:669-670 registers handlers for three database failures only.

**Demonstration**

```
Ran myself: wrote the first half of a zip's bytes and called both entry points.
  unpacked_size RAISES: BadZipFile File is not a zip file
  unpack RAISES: BadZipFile File is not a zip file
What the owner reads is fixed by extension/backend.js:156-158 — `try { detail = (await res.json()).detail || detail; } catch (_) {}` — and Starlette's 500 body is not JSON, so `detail` stays the status line and app.js:6193 prints "<name>.zip was not unpacked: Internal Server Error". The route spends four careful 400s on the failures it anticipated; the one a second machine will actually hit is the unstructured one.
```

### should fix · `extension/app.js:6286`

**The panel compares the engine's resolved echo against an unresolved path, so a working check is reported as an engine too old to run it**

`if (verdict.checked !== snapshot.path)` is a strict string comparison. `snapshot.path` is `str(p)` straight off the glob (storage.py:830), never normalised, and the folder prefix is whatever `backup_folder` or the pointer file holds — `backup_folder` does only `Path(saved).expanduser()` (storage.py:734) and `read_pointer` does only `Path(recorded)` (storage.py:88). The engine echoes `str(Path(asked).resolve())` (app.py:3275). The route's own membership guard resolves BOTH sides; only the echo comparison is asymmetric. The harness cannot catch it: tools/panel_harness.py echoes the request verbatim (`verdict.checked = COPY_CHECK_ECHOES ? wanted : ...`), so the comparison passes by construction.

**Demonstration**

```
Ran on this Windows machine, real folder `…\Backups`, path spelled `…\backups` the way a typed setting would be:
  panel receives (list_backups): …\adv877-pjohcn2_\backups\harvest.manual-20260101T000000Z.backup.db
  engine echoes  (.resolve())  : …\adv877-pjohcn2_\Backups\harvest.manual-20260101T000000Z.backup.db
  `verdict.checked !== snapshot.path` fires: True
Same file, correct verdict, and the panel says "This engine is too old to check a copy, so nothing here has been verified." Note this becomes MORE reachable the moment the backup-folder finding above is fixed, because the folder that then flows into list_backups is a typed string. Cheapest fix: emit `str(p.resolve())` once in list_backups so one spelling flows everywhere.
```

### should fix · `scrapex/webui/app.py:3757`

**adopt-bundle computes a health verdict on what it just wrote, answers ok:True regardless, and nothing reads the verdict**

Quoted from the file: `verdict = storage_health(adopted, integrity=False)` at 3757, then `"ok": True,` at 3759 with no branch, and the detail "… is unpacked. It is listed as a copy now -- check it, then restore it from the same card." `bundle.verify` checks manifest digests, not warehouse identity, so a bundle built by a NEWER engine verifies perfectly and then fails `_warehouse_identity` (storage.py:410-415, status `incompatible`) — and a second machine running a newer engine is precisely the scenario this route exists for. The only caller reads `made.detail` and nothing else (app.js:6187-6191); `made.health`, `made.name`, `made.path`, `made.bytes`, `made.from_bundle` have no reader anywhere in the panel.

**Demonstration**

```
app.py:3757-3766 as quoted — the call and the unconditional `"ok": True` are three lines apart with nothing between them — and `grep -n "made\." extension/app.js` returns only lines 6187 and 6190. The file is meanwhile written under this product's own backup naming (app.py:3750-3752), so a database `storage.restore` will refuse (storage.py:1129-1131) joins the Restore list, counts against the folder's space and is announced in the success colour.
```

### should fix · `tests/test_panel_dom.py:7897`

**The panel half of the headline feature ships with no test of any kind**

I ran the grep myself over tests/, extension/tests/, apps_script/tests/ and tools/: `renderBundles`, `adoptBundle`, `rowVerdict`, `db-bundles`, `db-bundle-hint`, `db-open-backups` and `unpacked_size` return NOTHING. The only hits for the feature anywhere are five route-level POSTs to /api/storage/adopt-bundle in the Python file. So +186 lines of app.js and +28 of app.html — including `#db-open-backups`, the first control this panel has ever had for POST /api/storage/open-folder — have no guard, while the route behind them does. CLAUDE.md: no module without its test file, error paths tested like happy ones.

**Demonstration**

```
`grep -rn "renderBundles|adoptBundle|rowVerdict|db-bundles|db-bundle-hint|db-open-backups|unpacked_size" tests/ extension/tests/ tools/ apps_script/tests/` → no output. The 126 new lines of tests/test_panel_dom.py are four tests, all about #restore-check. The harness already has `stub(storage=...)`, so this is test work only.
```

### should fix · `tools/panel_harness.py:740`

**The harness gives the copy and the live warehouse the same row count, so the one assertion cannot tell them apart — and the empty-copy sentence is untested on both sides of the wire**

Read the harness diff: `rows: {generic_page_snapshot: 12}, live_rows: {generic_page_snapshot: 12}` — identical — and the only assertion on the rendered verdict is `assert "12" in said` (tests/test_panel_dom.py:7918). `rowVerdict` reads `rows[biggest]` at app.js:6337 and `live[biggest]` at 6338; with both sides 12 the test passes whichever it reads, so swapping them is a surviving mutation. The branch that matters — `inCopy === 0 && inLive > 0`, "EMPTY in it … Restoring it would lose them" — has no panel test, and an intact empty copy has `verdict.ok === true`, so #restore-confirm stays ENABLED and that sentence is the only thing between him and the loss.

**Demonstration**

```
Quoted from the diff of tools/panel_harness.py: `rows: {{generic_page_snapshot: 12}},` immediately followed by `live_rows: {{generic_page_snapshot: 12}}}},`. And tests/test_panel_dom.py:7918 `assert "12" in said`. The route-level test at tests/test_the_integrity_scan_is_asked_for.py:381 proves the ENGINE sends two different numbers; nothing proves the panel says anything different when they differ.
```

### should fix · `tests/test_the_integrity_scan_is_asked_for.py:538`

**bundle.unpacked_size and both adopt-bundle refusal branches have no test**

`unpacked_size` is a new public function (bundle.py:407) and `grep -rn unpacked_size tests/ extension/tests/` is empty. Its only caller is the free-space refusal at app.py:3720-3727 — the guard with a measured number behind it — and that branch has no test; nor does the "verified but carries no warehouse.db" refusal at app.py:3742-3746. I listed the file's tests: the three adopt-bundle ones cover the happy path (538), the name that was not listed (599) and a bundle whose digests do not match (627). Error paths are tested like happy ones here.

**Demonstration**

```
`grep -n "^def test_" tests/test_the_integrity_scan_is_asked_for.py` — the adopt-bundle tests are exactly those three, and neither `free_space` nor a bundle without warehouse.db appears in the file. The `_a_bundle_in` helper the file already has makes each new test a few lines.
```

---

## What the adversary killed, and why

Its own success metric is the kill count, so the kill list is the only record of its false negatives. Read it as such.

- **ARCH: adopt-bundle writes into the configured backup folder; the copy check and restore only permit the warehouse's own folder (app.py:3259)** — Not wrong — killed as a DUPLICATE of the CODE-QUALITY finding at the same line ("The copy check looks in the database's folder, not the configured backup folder"). Same file, same line, same remedy. Kept once, as a survivor, at should fix.
- **ARCH: A half-downloaded .zip reaches the owner as "Internal Server Error" (app.py:3720)** — DUPLICATE of the CODE-QUALITY finding at the same line. Merged into one survivor.
- **ARCH: The route computes a verdict on the adopted file, answers ok:True regardless (app.py:3757)** — DUPLICATE of the CODE-QUALITY finding at the same line. Merged into one survivor.
- **TESTS: Every copy-check test runs with the default backup folder (must fix, tests/...:321)** — DUPLICATE of the backup-folder guard finding, stated as a test gap. The missing test IS the remedy for that finding, not a second defect; folded into it. Also: its 'must fix' rank rested on the claim that the feature is dead for a real owner — see the downgrade reason on the survivor.
- **ARCH: Every open of the restore dialog starts a full-file scan nothing can cancel (optional, app.js:6360)** — DUPLICATE of the PERFORMANCE finding at the same line. Merged into one survivor at optional.
- **CODE QUALITY: The copy check inherits a deadline derived from a WARM scan (should fix, startup.js:23)** — DUPLICATE of the PERFORMANCE finding on the same line; merged, and DOWNGRADED to optional there — 120,000 ms still covers the cold branch at his 2.1 GB (≈44 s measured by the PR itself) with 2.7x headroom. What is wrong today is a claim in a comment ("roughly 40 GB"), not a bound that fails.
- **ARCH (self-declared not an issue): Nothing starts work around transport.js → native.py, and the copy is never opened as the live warehouse (storage.py** — Not a finding — correctly so on the process-spawning half. But its second half is PARTLY WRONG and I am recording that: it certifies `row_counts` as "explicitly read-only" by reading the string `mode=ro`. I ran it. With `#` in the path the `?mode=ro` is swallowed into the URI fragment and the function opens read-write and CREATES a file. A clean bill given by reading a literal, not by running it — see the cold-read f
- **CODE QUALITY (self-declared not an issue): list_bundles is not duplicating list_backups (storage.py:762)** — Not a finding, and I agree on the merits: the two hold different rules (product naming with the -wal/-shm suffix filter vs. any zip a person dropped) and feed opposite guards. Killed as a non-finding, kept only as a note for the PR body.
- **PERFORMANCE (self-declared not an issue): Nothing new reads a bundle or a database into memory at once** — Not a finding; verified the load-bearing line myself. `unpacked_size` reads the central directory only and `unpack` streams through copyfileobj.
- **PERFORMANCE (self-declared not an issue): list_bundles is not on a polled path, and the new routes have no N+1** — Not a finding. `renderBundles` is reached from `loadDatabase`, which is called on page navigation and after mutations; neither setInterval in app.js touches /api/storage.
- **(my own candidate, killed) A half-written bundle in the backup folder is offered as adoptable** — I went looking for it and it is closed: `pack` writes to `<name>.zip.part` and renames, so the `*.zip` glob in list_bundles can never see a build in progress. Recording the kill so the next session does not re-derive it.
- **(my own candidate, killed) A bundle name could reach the DOM unescaped through manageRow** — Killed: manageRow builds the row with `el(...)` and textContent, and only the fixed `icon(lead)` string goes through insertAdjacentHTML. A filename cannot carry markup into the panel.

**Verdict reason, verbatim:** Two must fix and seven should fix survive, so it does not merge. I killed 12 findings — six as duplicates of the same line reported twice across dimensions, five as self-declared non-findings I re-verified (including one whose clean bill was given by reading a string rather than running it), and one candidate of my own that the code already closes. I downgraded four: the backup-folder guard from must fix (no panel control can set `backup_folder`, so the owner cannot enter the configuration where it fires, and it fails safe with a refusal), the test-gap restatement of that same defect (folded into it), the integrityScan derivation from should fix to optional (the bound still holds at his size with 2.7x headroom; the wrong thing is a claim in a comment), and the dialog-open scan and the extract-then-delete cost from should fix to optional (real costs on a once-per-disaster path, no correctness failure). What stands after that is still fatal: the copy check 500s on an unreadable copy — the exact file it was built to catch, after which the panel re-enables Restore and calls it merely unverified — and the cold read turned up a new-code defect nobody looked at, an unescaped `file:` URI in `row_counts` that, on a path containing `#`, opens and creates a different database, reports zero tables for a full warehouse, and takes the empty-copy warning off the screen at the moment a warehouse is about to be replaced. The remaining should fixes are unhandled archive failures, an asymmetric path comparison that reports a working engine as too old, a success message contradicted by the verdict computed two lines above it, and three test gaps over a panel surface that currently has none.

muhammadbayoumi:
All nine are answered on `17639c3`, and the argument is in the PR body beside
its own diff. The short version, in the order they were reported:

| finding | what it now cannot do | killed by |
|---|---|---|
| `storage.py:805` — the `file:` URI | a path reaches SQLite as characters, not syntax: `Path(...).resolve().as_uri() + "?mode=ro"` | two mutants, one per character |
| `app.py:3276` — counts on an unreadable copy | `_counted()` is the only way this route counts, on both sides of both branches; 200 with `unreadable` and `rows == {}` | zeroed-header copy, at the route |
| `app.py:3259` — the folder-less guard | the configured backup folder is resolved first, as `adopt-bundle` beside it already did | a copy in `…/here/../Backups` |
| `app.js:6286` — the asymmetric echo | fixed at the source: `list_backups` emits `str(p.resolve())`, so one spelling leaves the one function that decides what a backup is | the offer posted back and compared to the echo |
| `app.py:3720` — the unhandled archive | two refusals, for the directory read and for `unpack` itself | half a real bundle, and an entry named `../escaped.txt` |
| `app.py:3757` — `ok: True` regardless | `ok` is the verdict's, the sentence names the copy and what is wrong with it, and the panel draws it in `err` | a stranger's database swapped into a verified bundle |
| `test_panel_dom.py:7897` — no panel test | six DOM tests over the card | six mutants, including the folder control that sends nothing |
| `panel_harness.py:740` — 12 on both sides | the live side is 340, one test asserts both numbers in the right order, and the empty-copy branch has its own | the swap, now killed by two tests independently |
| `test_the_integrity_scan…:538` — untested branches | `unpacked_size` asserted against what unpacking writes, plus both refusals | the free-space guard, disarmed |

**Where I differ, and where I went further:**

- **`restore` was not fixed here.** It carries the identical folder omission at
  `storage.py:1137`; it is pre-existing, it replaces a live warehouse, and it is
  **#926** rather than a line riding along on this branch. The two pre-existing
  `file:` URIs are already **#925** — untouched.
- **An unusable adopted copy is not deleted.** Unpacking wrote the only local
  copy of what is in Drive, and throwing it away on his behalf is a worse answer
  than naming it. It is listed, reported `ok: false`, and he decides.
- **One optional finding taken as well**: `integrityScan`'s comment claimed
  120,000 ms covers "roughly 40 GB", read off the warm 5.9 s. This deadline bounds
  the copy check now, and a backup nobody has touched is read cold — about 6 GB,
  2.7x headroom at his warehouse today. The bound was right; the number beside it
  was not. Comment only.
- **`made.health`, `made.path`, `made.bytes`, `made.from_bundle` still have no
  panel reader.** `ok` and `detail` do now, and `detail` names `made.name`.
  Narrowing a reply because today's only caller ignores three of its fields buys
  nothing.

The two you found by running rather than reading — the URI and the unreadable
copy — were both reproduced here before anything was edited, with the stray
database named `Drive` appearing in the folder listing exactly as you measured it.

muhammadbayoumi:
Merged as `1a658f4`. The second review pass did not run, on the owner's instruction after the gap was put to him — recorded in the PR body above and filed as #932, following #890 from #886.

Left open behind it: #925 (two pre-existing unescaped `file:` URIs, in `bundle.py` and `carry_over.py`), #926 (`restore` carries the same backup-folder omission this PR fixed in the copy check), and a note on #910 that `SCRAPEX_DATA_ROOT` — the escape hatch that issue names — turns `test_a_killed_engine_does_not_leave_a_job_claiming_to_run` red on its own.

Issue #875, opened 2026-09-10
Review pass 3 on #828: one must fix, and two corrections to pass 2 own prose


muhammadbayoumi:
Review pass 3 on #828 reported **after** the session that ran it had already
written its handover (#870). It is not clean. Nothing here is fixed; the branch
stands at `d5f710ee`.

Pass 3 attacked pass 2's own fixes. Two of its findings are corrections to
things written in pass 2, which makes them worth reading before anyone trusts
the comments on this branch.

## must fix

**1 · The AST guard has no coverage floor, and it stops guarding on a rename —
green.** `tests/test_organization_enrichment.py:798` and `:831`

Reproduced: rename `workbook_tables` to `export_tables` across `scrapex/`
(5 files) and **the guard passes**, while the `try` blocks it inspects fall from
3 to 1. Its own docstring claims "a reformatted `except` line or a renamed local
cannot make the guard quietly stop looking" — a renamed *function* does exactly
that, permanently.

This repo already solved this twice and named the failure:
`tests/test_the_crawls_own_report_names_no_command.py:98` (`assert len(lines) >= 25`
— "a refactor that moved the report elsewhere would silently empty this test")
and `tests/test_a_lit_capability_is_a_switch_and_not_a_claim.py:209`
(`assert calls >= 2`). The new guard is the only source-scanning guard in
`tests/` without a floor.

Recommended: assert the number of inspected blocks is `>= 3`, AND derive
`raisers` from the imported symbols (`publish.workbook_tables.__name__`) so a
rename either updates the guard or breaks the import. About 4 lines.

## should fix

**2 · "Clause order is what keeps them right" is false — and it is pass 2's
replacement for pass 1's *other* wrong justification.**
`scrapex/publish.py:23-25`, `tests/test_a_dataset_is_a_table_like_any_other.py:689-691`

Measured: move `except ValueError` *ahead* of the `UnexportableCell` clause at
both ordered sites with the `TypeError` base intact — every behavioural test
stays green. So the base class is what keeps those sites right and order is
moot, which is what the guard's own comment already says.

The accurate statement, for whoever writes it a third time: **the `TypeError`
base is what stops an `except ValueError` swallowing the refusal; the explicit
clauses are what produce the right message; clause order matters only if that
base is ever changed.** Pass 1 said the base does everything, pass 2 said order
does, and each was half of it.

Also: the deleted `assert not isinstance(inside.value, ValueError)` **does**
[...]

PR #888, opened 2026-09-11
What a review pass may spend, and the order that stops it buying nothing

muhammadbayoumi:
Five rules into the `review` skill, from a merge study on #828 that spent
**~1.8M subagent tokens on a 96-line product change** and wasted a large
fraction of it. Every rule is one measured failure, not a precaution.

## Why none of it went in `CLAUDE.md`

`CLAUDE.md` is **loaded into every session** — 150 lines, ~2.3k tokens, paid
whether a review runs or not. A skill is read only when it is used. These are
gate mechanics, so the skill is where they cost nothing until they apply.

And no new line was needed anywhere: `CLAUDE.md:115` already delegates "the
rules that decide the outcome" to this skill. **Zero always-loaded lines added,
no control deleted to make room, and the 150-line cap untouched** — which is the
answer its own "How this file evolves" gives: move detail to a `.claude/` skill.

## What each rule is paid for

| rule | the failure that bought it |
|---|---|
| **Step 0 — read the record first** | The study's *must fix* was found by one `gh issue list`. It was written down in #875; fourteen reviewer sessions reading the code never found it. |
| **Step 2 — reviewers after the green** | A full panel was launched at a head whose CI then went red. Everything it returned described a tree that no longer existed. |
| **Prove the environment before fanning out** | Seven of eight reviewers in one run came up at `origin/main` instead of the PR head. The **symbol** assertion caught it — `__file__` passed cleanly. Without it, the tests dimension runs eight mutations, watches every one stay green, and reports "vacuous guards everywhere". |
| **Counts from changed lines** | 14 reviewers for 96 changed lines is not a thorough review, it is an unmeasured one. |
| **Eight is the ceiling** | Agrees with the split rule already in this file instead of adding a second, separate limit. |

Estimated under these rules the same study costs **~200k instead of ~1.8M**, and
returns a *better* answer — step 0 alone puts the must fix in front of the
session in its first minute.

Documentation-only: exempt from the merge gate by `CLAUDE.md`'s own Review
section, and this is not `CLAUDE.md`. The five document suites are green,
including `test_the_documents_cite_what_they_claim.py` and
`test_the_ruling_matches_the_code.py`.


muhammadbayoumi:
**Correction to this body, and how it relates to #884.** Written when `CLAUDE.md`
was 150 lines. #884 landed while this sat open and it is now **156** — so "the
150-line cap untouched" above describes the base this branch was cut from, not
`main` today.

The two changes **compose rather than collide**, and neither needs the other
edited. #884 put the principle in the always-loaded file: *"Never assume his
priorities on timeline, scale or spend. Ask, with a number and a recommendation
— a fan-out names its worst case before it runs, and one that multiplies an
unknown count has no ceiling at all."* That is the part every session must carry,
review or not.

This PR is the number it tells you to bring: the counts per changed line, the
pre-flight that must pass before any fan-out, the cold-read condition, and the
ceiling of eight. Gate mechanics, read only when a gate runs.

So the argument for putting these here is unchanged, and the measured saving is
unchanged. What is no longer true is only the claim about the cap's current
state.

PR #896, opened 2026-09-11
AGENTS.md points at the rules instead of copying them

muhammadbayoumi:
## What happened

At 16:37 today a full copy of `CLAUDE.md` appeared in this worktree as `AGENTS.md`, byte
for byte across all 156 lines **except one**: `` `.claude/` `` rewritten to `` `.Codex/` ``,
a directory that does not exist here. Alongside it, `.agents/skills/` mirrored
`.claude/skills/`, with the `review` skill saying `AGENTS.md` in the four places the real
one says `CLAUDE.md` — including *"Documentation-only changes are exempt — except
`AGENTS.md`"*.

Neither path was tracked and neither was ignored, so `git add -A` would have committed a
second rules document. Nothing in the repository generates them, there is no hook in
`~/.claude/settings.json`, and there are six live worktrees — so the writer was external
and was not identified. Both were backed up and deleted on his instruction.

## Why a pointer, and why tracked

`CLAUDE.md` says it stays the only rules document, and its own DRY rule says one source of
truth per piece of knowledge. A byte-copy under a second name is the opposite, and the
copy written today **already carried a dead path** within hours of being made.

**The name is what attracts a generator.** A file that does not exist gets a copy written
into it. A pointer that does exist gets found — which also answers the question that opened
the session: implementers that are not Claude read `AGENTS.md`, and today they would reach
this repository blind to all 156 lines.

**Tracking is the control that ignoring would have destroyed.** `.gitignore` stops the
accidental commit but makes every future write invisible — the failure mode is silence. A
tracked file makes the next write land as a diff in `git status`, in every worktree, in
every session. Loud beats hidden.

## The guard, and what it does not cover

`AGENTS.md` joins the citation guard's `DOCUMENTS` map, for what `_read` does to it at
`tests/test_the_documents_cite_what_they_claim.py:176` rather than for any citation it
makes. That is stronger than `git status`, which a session can ignore.

**Broken on purpose, twice:**

```
$ mv AGENTS.md /tmp/ && python -m pytest -m docs -q
E  AssertionError: AGENTS.md is in the map but not in the repository

$ mv AGENTS.md AGENTS.MD.tmp && python -m pytest -m docs -q
FAILED ...::test_no_document_holds_a_table_with_no_header[AGENTS.md]   (+3 more)
```

**Named rather than implied, in the comment beside the entry:** `LINKED_CITATION` needs a
`path:line` label over an `#L<n>` href, and a pointer has neither, so the link *inside*
`AGENTS.md` is not checked by tier 1. The entry buys existence, not link validation. An
earlier draft of this description overclaimed that; this is the corrected account.

## Re-run it

```
python -m pytest -m docs -q        # 293 passed, 1 skipped here
```

Local only — `ruff` is not installed on this machine and the browser suites skip, so lint
and the full tier are CI's word.

## Not in this change

The writer was not identified, and nothing here stops it writing again — it makes the
writing visible. If it recurs, the diff will say so.


muhammadbayoumi:
Superseded by #900. The pointer bet on an agent following an instruction; with `main` unprotected there is no second line of defence behind that bet, so #900 carries CLAUDE.md whole instead.

Issue #653, opened 2026-09-05
main has no branch protection at all, so 11,529 runner-minutes a month gate nothing


muhammadbayoumi:
**11,529 runner-minutes over 30 days currently gate nothing.**

## Measured, twice, hours apart

```
GET /repos/muhammadbayoumi/ScrapeX/branches/main/protection   -> 404 "Branch not protected"
GET /repos/muhammadbayoumi/ScrapeX/rulesets                   -> []
GET /repos/muhammadbayoumi/ScrapeX/branches/main              -> protected: false,
                                                                 required_status_checks.enforcement_level: "off",
                                                                 contexts: 0
```

## The three consequences, in order

1. **No check is required before a merge.** A red or still-pending `test` does not block anything; the discipline is entirely human.
2. **CI never recomputes on rebase**, because nothing requires branches to be up to date. This is the failure already recorded in the archived ORCHESTRATION.md: two pull requests, each green against its own base, merged in sequence and produced a red `main`.
3. **Nothing blocks a direct push to `main`**, including one that never opened a pull request.

The merge session has been carrying (2) by hand as a post-merge rebase rule. It is not a stopgap for a missing setting — **it is the only gate that exists.**

## The change

One repository-settings decision, not three: protect `main`, require the checks that should be required, and require branches to be up to date. Which checks become required is the substantive part — and it interacts with the `migration-authority` issue in this milestone, since a job that is deleted must not be left in a required list, and a job that is merged into another changes the name that must be required.

**This is a settings change, not a code change, and it is the owner's alone.** No session should make it. Recorded here so the decision is not carried in someone's head.

**Effort S. Risk of NOT doing it: HIGH** — it is the only item in this milestone where the current state can put a broken tree on `main` rather than merely waste time.

muhammadbayoumi:
**This is now out of date — the owner had branch protection enabled on `main` at 2026-09-05, and the finding's premise no longer holds.** Recording it here so nobody re-measures a settled question.

Verified live just now:

```
protected                              true
required_status_checks.strict          true      ("require branches up to date")
enforcement_level                      non_admins
required contexts   lint · scope · test · migration-authority · contract-parity · CodeQL
allow_force_pushes / allow_deletions   false
```

Two deliberate choices, both worth stating because a later reader will ask:

- **`enforce_admins` is `false`.** He is the sole owner; a protection that locks him out of his own repository in an emergency is not a protection. He keeps the override.
[...]

Issue #894, opened 2026-09-11
Is rebase-then-green-then-squash the right merge architecture? main moved four times during one merge


muhammadbayoumi:
**His question, recorded unanswered on his instruction.** Nothing here is a
recommendation; it is the evidence one merge produced today, for whoever studies
this.

The merge architecture today is: one session takes a PR, rebases it onto `main`,
waits for CI green on that exact head, and squash-merges. `CLAUDE.md` rule 5 and
the `review` skill's merge gate define it.

## What one merge cost, measured

Merging #828 today:

| | |
|---|---|
| `main` moved during the work | **4 times** — `6449e36` → `e868486` → `2ce010e` → `a31a8f8` |
| rebases forced by that | 3 |
| CI `test` duration | **14m59s**, then **13m38s** |
| merges into `main` in 24h | **6** — but 4 of them inside this one ~3h session |

## The structural tension

A recorded green expires, and every rebase starts a **~14-minute** window. During
this session `main` moved about once every **45 minutes**, so the chance of being
overtaken mid-window was roughly **1 in 3** — and it happened three times out of
three. The loop only terminates because the merging session eventually stops
rebasing.

**It did stop, by judgement, not by rule.** The fourth rebase was skipped after
measuring that #886 shared **zero files** with #828's diff and that its only
adjacent file changed a changelog sentence. The architecture has no rule for
that decision, so it was made ad hoc — which is the part most worth deciding
deliberately.

## A hard constraint the study should know first

Branch protection is **not available on this repository's plan**:

```
GET /repos/…/branches/main/protection
403  "Upgrade to GitHub Pro or make this repository public to enable this feature."
```

So required status checks, `enforce_admins`, and **GitHub merge queues** — the
standard answer to exactly this problem, which tests the integrated result once
and serialises merges — are all unavailable without changing plan or visibility.
That bears directly on #842 ("Six required checks are advisory for the only
account that can merge: one aggregator job, then `enforce_admins`"), whose
proposed remedy may not be reachable as things stand.

## Questions this leaves open

- When `main` moves under a green head, what makes a rebase **required** rather
  than optional? Today's judgement used file overlap; nothing writes that down.
- Is a ~14-minute gate compatible with the observed burst rate, or does the gate
  need a cheaper pre-merge tier?
- Does the plan constraint above make a queue worth paying for, or does it argue
  for a convention instead — a merge window, or one merging session at a time
  across the whole repo rather than per PR?
[...]

PR #900, opened 2026-09-11
AGENTS.md carries the rules, because nothing else can stop an agent that has not read them

muhammadbayoumi:
Supersedes #896, which proposed a pointer. Closing that in favour of this.

## Why a copy and not a pointer

```
gh api repos/:owner/:repo/branches/main/protection
→ 403: Upgrade to GitHub Pro or make this repository public to enable this feature
```

`main` is unprotected and cannot be protected on the current plan (#653). No required
check, no review gate, nothing mechanical stops an agent pushing straight to it. **The
rules document is the only control there is.**

Codex is running against this repository — `~/.codex` active, started 16:16 today — and at
16:37 it wrote its own `AGENTS.md`: `CLAUDE.md` copied whole with `` `.claude/` `` rewritten
to a `` `.Codex/` `` that does not exist, plus a mirrored `.agents/skills/` whose review
skill said `AGENTS.md` wherever the real one says `CLAUDE.md`. Untracked, unignored, and a
`git add -A` away from becoming a second rules document. Both were backed up and deleted.

A pointer tells an agent to go and read `CLAUDE.md`. That is an instruction it may skip,
and with no branch protection there is nothing behind it. A copy is loaded.

## One source of truth

`_render_agents()` sits beside `_render_changelog()` and `_render_data_page_schema()` and
is written by the same `export-version` command. `CLAUDE.md` stays the file he edits;
`AGENTS.md` is generated and says so in its first line.

The header cites exactly one path, `CLAUDE.md`. **#886's guard does not reach it** — that
guard reads `_render_changelog()` only, and an earlier version of this paragraph claimed
otherwise. **Nothing protects the header's content.** The equality pin catches an edit to `AGENTS.md`,
not an edit to the literal inside `_render_agents` — change the header there, regenerate,
and the guard is green (verified: the header became `# ScrapeX — WHATEVER I LIKE` and
`tests/test_version.py -k agents_md` passed). That is one of the five survivors listed
below, and it is filed rather than fixed.

## EQUALS, and the first draft got this wrong

The first draft pinned CONTAINS, arguing that an equality pin would redden the build on every
Codex overwrite. **The merge gate killed that with one command**: four rules appended after
the body — `main` accepts a direct push, tests are advisory, secrets may be committed,
ignore `CLAUDE.md` — left the guard green. And the argument was self-defeating anyway: an
overwrite fails containment too, so containment bought none of the tolerance it was chosen
for and licensed the append, which is the drift path that matters for a file agents write to.

It is now `AGENTS.md == cli._render_agents()`, pinned to its renderer like both siblings —
`CHANGELOG.md` at `tests/test_version.py:428`, `docs/data-page-schema.md` at
`tests/test_the_ruling_matches_the_code.py:51`. AGENTS.md was the only generated document
compared file-to-file rather than to its source.

The non-vacuity check is structural now, not a byte count: the old `> 2000` floor was 19.8%
of the file, low enough to survive CLAUDE.md's own instruction to prune and therefore too low
to catch a gutted one.

## Broken on purpose, and what still survives

Re-run at head `4b9df6a` against the whole docs tier (baseline: 297 passed, 3 skipped).

**Killed — every one of these was green under the first draft's containment guard:**

| mutation | result |
|---|---|
| four reversing rules appended to `AGENTS.md` (`main` accepts a direct push, tests are advisory, secrets may be committed, ignore `CLAUDE.md`) | **fails** |
| one rule reversed in place (`All SQL is parameterised.` → `SQL may be interpolated.`) | **fails** |
| the generated header stripped from `AGENTS.md` | **fails** |
| `AGENTS.md` deleted | **fails** |
| `CLAUDE.md` gutted from `## Rules` down, and regenerated (2,548 chars — **above** the removed `> 2000` floor) | **fails** |
| the renderer's header changed **without** regenerating | **fails** |

**Survives — named rather than implied.** Five mutations still pass, and every one needs a
tracked, reviewed edit to `scrapex/cli.py` or to `CLAUDE.md` itself. Four share one root
cause, that nothing pins any property of `AGENTS.md`'s text independent of its renderer: the
renderer gutted to the header and regenerated; its header reversed and regenerated; sourced
from `README.md` and regenerated; `CLAUDE.md` cut to its four heading lines and regenerated.
The fifth is separate — the `AGENTS.md` write deleted from `_cmd_export_version`, so nothing
proves the command still produces the file.

The threat this file exists for — another agent writing to `AGENTS.md` — is dead in every
form tested. The survivors are visible in a diff a human reads. They are filed, not fixed.

**An earlier version of this section described the containment guard the gate refused**,
including a row asserting that an append *passes*. The table above was produced by running
the mutations, not by editing the old one.

## Re-run it

```
python -m pytest tests/test_version.py -q      # 36 passed here
python -m scrapex.cli export-version           # regenerates AGENTS.md and the rest
```

Local only — `ruff` is not installed on this machine and the browser suites skip, so lint
and the full tier are CI's word.

## Not in this change

Making the repository public was asked for and **not done**: git history carries two Google
API keys. One (`AIzaSyBL5buz…`) is the picker's browser key, deliberately allowlisted at
`tests/test_no_tracked_file_carries_a_secret.py:47` as referrer-restricted; the other
(`AIzaSyAZpoNuLb…`) is a third party's key captured in saved fixtures, scrubbed from the
tree by #466 but still in history. Publishing is irreversible and Google auto-revokes keys
it finds in public repositories. That decision is his, separately, and #653 names a cheaper
route to the same protection.


muhammadbayoumi:
Two measurements from a merge session running beside this one, both about the
sibling this PR does not cover. Not a competing proposal — the argument here is
sound, and I reached its central premise independently: I recorded in #894 that
branch protection is unavailable on this plan (`403 Upgrade to GitHub Pro or make
this repository public`), which is exactly why a skippable pointer would not do.

## 1 · The `git add -A` risk is not hypothetical. It fired today.

#896 said a `git add -A` "would have committed a second rules document". An hour
ago, in worktree `merge-session-2b3b8f`, it did:

```
 .agents/skills/karpathy-guidelines/SKILL.md |  85 +++++
 .agents/skills/record-it/SKILL.md           |  37 +++
 .agents/skills/review/SKILL.md              | 121 +++++
 AGENTS.md                                   | 156 ++++++
 scrapex/publish.py                          |  33 +--
 tests/test_organization_enrichment.py       |  14 +-
```

That was a two-file prose change. The other four rode in on `git add -A` and were
caught only because the diffstat was read before pushing; the commit was then
rebuilt without them. There are six live worktrees, so this is six standing
chances, and the next session may not read the diffstat.

## 2 · The guard covers `AGENTS.md` and nothing else

`git diff origin/main...claude/agents-md-carries-the-rules | grep -c '\.agents'`
→ **0**.

`test_agents_md_still_carries_every_rule` is well built — containment rather than
equality so an appending agent does not redden the build, with a non-vacuity
floor on `CLAUDE.md`'s size. But `.agents/skills/` is neither tracked nor ignored
nor guarded, and it is a mirror of `.claude/skills/` with the same self-renaming
applied. Measured on the copy sitting in this worktree, `.agents/skills/review/SKILL.md`
differs from the real skill in exactly four lines, all of them that rename —
including:

```
- Documentation-only changes are exempt — except `AGENTS.md`
+ Documentation-only changes are exempt — except `CLAUDE.md`
```

Worth stating precisely, because it is less alarming than it first looks: under
this PR's design `AGENTS.md` contains `CLAUDE.md` whole, so editing `CLAUDE.md`
alone breaks the containment guard and goes red. **This PR already covers that
path indirectly.** The uncovered part is the mirrored skills themselves.

## Suggestion, for this PR rather than another

One line — `.agents/` in `.gitignore`. It removes the accidental-commit path
without hiding anything: the directory stays in the working tree where Codex
reads it. And it does not pre-empt the real question, which is yours to settle
here — whether Codex loads `.agents/skills/` at all, and whether the mirror
should therefore be tracked and guarded like `AGENTS.md` is. If it should, this
is one line back.

I am not taking this PR and have opened nothing against it; #896 and #900 already
produced two answers to one question, and a third would repeat that rather than
fix it.

PR #901, opened 2026-09-11
Sync the locally loaded extension/ folder with GitHub main

muhammadbayoumi:
## Why

Every session that merges an extension change leaves the owner's own loaded
copy behind it: "Load unpacked" points Chrome straight at `extension/` in his
checkout, and nothing updates that folder but a manual `git pull` followed by
a manual Reload in `chrome://extensions`. He works only from the panel, never
a terminal (`CLAUDE.md`), so that pull has only ever happened when a session
ran it for him — the exact five-step manual routine written out in
[`docs/plans/2026-07-29-sync-green-main-and-merge.md`](docs/plans/2026-07-29-sync-green-main-and-merge.md),
phase 1, step 5.

## What

- **`scrapex/extensionsync.py`** (new) — `check()` reports where `extension/`
  in this checkout stands against `origin/main` (`up_to_date` / `behind` /
  `dirty` / `not_main` / `offline` / `no_git` / `error`); `apply()`
  fast-forwards to it. Wired into `scrapex/native.py` as two new
  `STANDALONE_COMMANDS`: `CHECK_EXTENSION_SYNC`, `APPLY_EXTENSION_SYNC` —
  native-only, the same reason every command in that file is: a page cannot
  run git, only the host on the machine can.
- **`extension/transport.js`** — `checkExtensionSync()` / `applyExtensionSync()`,
  the two callers `grep` finds for the commands above.
- **`extension/app.js` + `app.html`** — a line in Settings → Product → About:
  "N commits behind main — <summary>" with an **Update now** button that
  appears only when there is something to apply.

## Design decisions (asked and answered before writing code)

Three things here are genuinely his call, not mine to assume — asked via
`AskUserQuestion` before implementation, recorded here since the reasoning
matters for review:

1. **Sync target**: this main checkout (not a worktree) — the one Chrome's
   "Load unpacked" already points at.
2. **Apply mode**: auto-check, manual apply. Checking runs in the background
   (see below); applying is always the "Update now" click, so a live job is
   never interrupted by a silent reload. Mirrors the existing pattern in
   `extension/releases.js` for the engine's own update check.
3. **Local edits**: refuse and report, never stash or discard. `check()`
   returns `dirty` and `apply()` refuses before ever touching git if
   `extension/` has uncommitted changes.

## Two things worth a reviewer's attention

- **`apply()` fast-forwards the WHOLE checkout, not just `extension/`.** A
  sparse `git checkout origin/main -- extension` would leave HEAD behind, so
  the very next check would see `extension/` as dirty against its own HEAD
  forever — it would apply an update once and then refuse every update after.
  `git merge --ff-only` keeps the tree internally consistent. It only touches
  the working directory if `extension/` actually changed, and git itself
  refuses — loudly — the instant any local edit anywhere would be
  overwritten, so this is not a wider blast radius in practice, only a
  correctness requirement. Documented in `scrapex/extensionsync.py`'s module
  docstring.
- **The check is deferred behind the About section actually being opened**,
  the same way `maybeRenderAutostart` defers `AUTOSTART_STATUS` — nothing
  calls the native host on plain panel open.
  `test_native_status_is_deferred_until_its_settings_section_is_visible`
  caught the first version of this PR doing that unconditionally on every
  panel open; the fix is the `maybeRenderExtensionSync` gate now in `app.js`.
  From the first open, a 5-minute timer keeps it current for as long as the
  panel stays open — the "continuously" half of the ask.

## Not touched, on purpose

- No `VERSION` bump, no `scrapex/version.py` capability-ledger entry. This is
  a native-host-only control feature — `AUTOSTART_STATUS`/`SET_AUTOSTART`
  are the precedent for that, and neither is in the ledger either.
- No change to `extension/manifest.json` — no new permission is needed;
  `chrome.runtime.reload()` requires none, and native messaging was already
  granted.

## Verification

```
python -m pytest -q                                  # full suite, green (a handful of
                                                       # pre-existing skips, unrelated)
node --test extension/tests/*.test.mjs                # 464/464
node contract/parity/parity.test.mjs                  # 3/3
```

New tests: `tests/test_extensionsync.py` (15 cases against real, throwaway
local git repos — no network, no touching the real ScrapeX checkout) and
`extension/tests/transport.test.mjs` (3 cases). `tests/test_native.py`'s
`test_the_control_commands_the_panel_actually_calls_still_answer` extended
to the new pair.


muhammadbayoumi:
Closing without merging — the owner wanted this done for him directly, not shipped as a new capability in the codebase. Leaving the branch in place in case any of it (scrapex/extensionsync.py, the native-host commands) is useful later, but this is not going into main as-is.

PR #914, opened 2026-09-11
Pin the account to the process, because switching it is global

muhammadbayoumi:
His instruction, after it cost a session today.

## What happened

Mid-work on this repository, every `gh` and `git` call started answering:

```
GraphQL: Could not resolve to a Repository with the name 'muhammadbayoumi/ScrapeX'
remote: Repository not found.
```

The repository was fine. The **active account** had moved to another one:

```
gh auth status
  ✓ Logged in to github.com account datasouq (keyring)   - Active account: true
  ✓ Logged in to github.com account muhammadbayoumi      - Active account: false
```

`gh auth switch` writes **global** state. It changes the active account for every session on this machine at once, and the rest collapse mid-work. The failure is hard to read because GitHub answers **404, not 403**, for a private repository the active token cannot see — so a wrong account looks exactly like a deleted repository.

Three sessions were running at the time.

## The rule

```bash
export GH_TOKEN=$(gh auth token --user <the account that owns the repository>)
```

`gh` ignores the active account entirely when `GH_TOKEN` is set, and `git push`/`fetch` take it through gh's credential helper. The variable is **per-process**, so two sessions can work on two different accounts at the same moment and neither touches the other — and no setting is changed permanently.

Verified here: with the active account still `datasouq`, this recovered full access and completed a merge, without `gh auth switch` and without altering global state.

## Why the `record-it` skill and not `CLAUDE.md`

Every command that skill documents is a `gh` command, so the rule that makes them work belongs at its head.

And `CLAUDE.md` is at **156 lines** against its own stated 150, where its rule reads: *"never delete a control to make room. Past 150 lines the next change here starts by pruning what no longer changes an action; if nothing can go, move detail to a comment beside the code or to a `.claude/` skill."* Nothing in it has stopped being true, so this is the move it prescribes.

## The class this belongs to

Third instance this session of the same shape — state that looks local and is not:

- a review agent's `git fetch --depth 1` left `.git/shallow` set, which is **repo-wide** across every worktree (fixed by #878)
- `git checkout --` on an **uncommitted** file discards the work, not just the mutation (cost two rewrites)
- `gh auth switch` changes the account for **every session on the machine**

Documentation-only, and not `CLAUDE.md`, so exempt from the merge gate.


muhammadbayoumi:
Duplicate. #913 already landed this on `main` (`CLAUDE.md:141-146`), and #915 is a third copy of the same rule.

He sent the instruction to several sessions at once and asked that it not be added once per session in a different shape. The rule is right and it is already in the one place that every session reads; a second wording in a skill would be the same knowledge in two voices, which is what `CLAUDE.md`'s own DRY rule refuses.

Nothing in this branch is worth keeping over what merged — closing rather than rebasing.

Issue #892, opened 2026-09-11
CLAUDE.md is 156 lines against its own cap of 150, and neither change that took it there pruned


muhammadbayoumi:
`CLAUDE.md` is **156 lines** against its own cap of 150, and neither change that
took it there did the pruning the cap demands. Recording it for the session
studying this file; not acting on it.

## Measured, per commit

| commit | lines | words |
|---|---|---|
| `e868486` (#878) | **150** | 1,538 |
| `0526562` (#883) | **154** | 1,590 |
| `2ce010e` (#884) | **156** | 1,618 |
| `a31a8f8` (#886) | 156 | 1,618 |

Two PRs, +6 net, no deletion in either.

## What the file says about exactly this

> "Past 150 lines the next change here starts by pruning what no longer changes
> an action; if nothing can go, move detail to a comment beside the code or to a
> `.claude/` skill."

That sentence is now itself at **line 154** — the rule that governs the growth
was pushed past the limit it sets by the growth it was meant to govern.

The companion clause, which nothing here violated and which should not be read
as an escape: **"never delete a control to make room."**

## Why the number is not cosmetic

`CLAUDE.md` is loaded into **every session**, whether or not it acts on any of
these rules. At 156 lines / 1,618 words it is roughly **2.4k tokens per session,
paid unconditionally**. A `.claude/` skill is read only when invoked. So the cap
is a per-session token budget, and overrunning it taxes every future session for
rules that only bind some of them.

## The content is not the problem

Both additions are good and belong in the always-loaded file — #884's "never
assume his priorities on timeline, scale **or spend** … a fan-out names its worst
case before it runs" is exactly the kind of rule that must reach every session.
The defect is the process: the cap's own remedy was skipped.

## Related

- **#843** — "CLAUDE.md is exempt from its own merge gate: the rules amend
  themselves through the one unguarded path." **This is the first measured
  consequence of that gap**: a documentation-only change is gate-exempt, and the
  one document that is *not* exempt is the one that grew.
- **#888** (open) — the same class of rules (how many agents a review may spend,
  in what order) put into the `review` skill instead, for **zero always-loaded
  lines**, because `CLAUDE.md:115` already delegates "the rules that decide the
  outcome" there. Offered as the worked alternative, not as a competing proposal.

## Not recommending a fix here

Which six lines go, or whether the cap should move, is a judgement about what
still changes a session's actions — the study's call, and his.

muhammadbayoumi:
[...]

PR #915, opened 2026-09-11
gh auth switch is global, and it takes the other sessions down with it

muhammadbayoumi:
His finding, recorded where the other machine-specific `gh` note already lives.

## It diagnoses something this session got wrong

Mid-merge on #902, `git fetch origin` failed with:

```
remote: Repository not found.
fatal: repository 'https://github.com/muhammadbayoumi/ScrapeX.git/' not found
```

I called it transient, retried, and it worked — so I moved on. **It was not
transient.** Another session had run `gh auth switch`, which rewrites the active
account for every session on the machine. The message reads exactly like a
network blip, which is why it gets diagnosed as one and the real cause survives.

## The fix changes nothing permanent

```bash
export GH_TOKEN=$(gh auth token --user <the account that owns the repo>)
```

`gh` then ignores the active account entirely, and `git push`/`fetch` pick the
same token up through `gh`'s credential helper. Two sessions can work as two
different accounts at the same moment, and nothing on disk is modified.

## Measured before writing it down

| | |
|---|---|
| `gh api user` with `GH_TOKEN` pinned | `muhammadbayoumi` |
| `gh api repos/muhammadbayoumi/ScrapeX` | `muhammadbayoumi/ScrapeX` |
| `git fetch origin` | OK |
| **`GH_TOKEN` in a fresh tool call, no export** | **empty** |

That last row is why the rule says to put it in front of **every** command that
reaches GitHub rather than once at the start: a shell here keeps its working
directory and loses its environment. And it says explicitly not to park the token
in a file to make it last, because that writes a secret to disk.

## Why `record-it` and not `CLAUDE.md`

It is machine-specific `gh` usage, which is what the "On this machine" section
already holds — alongside `gh` not being on `PATH`. `CLAUDE.md` is also at 156
lines against its own cap of 150, with **#892** open against exactly that, so
adding six always-loaded lines there would make a recorded problem worse.

Documentation-only, and not `CLAUDE.md` — exempt from the merge gate by
`CLAUDE.md`'s own Review section. The document suites are green.


muhammadbayoumi:
Closing as a duplicate — **#913 landed this first and its content is correct.** Verified rather than assumed, with the active account being a *different* one (`datasouq`), so the override was a discriminating test and not a tautology:

| claim in #913 | measured |
|---|---|
| `gh` obeys `GH_TOKEN` over the active account | `gh auth status` → active is `datasouq`; with the token pinned, `gh api user` → `muhammadbayoumi`, and `git fetch origin` succeeded |
| the helper key is HOST-SCOPED | `credential.https://github.com.helper` → `gh.exe auth git-credential` |
| `--get-all credential.helper` hides it | → `manager` |

Two operational facts I measured that #913's note does not carry. Recording them here rather than in a fourth PR, for whoever next edits that paragraph:

1. **It does not survive between tool calls.** A shell here keeps its working directory and loses its environment — `GH_TOKEN` reads empty in the next call. So it belongs in front of *every* command that reaches GitHub, not once at the start.
2. **Do not park the token in a file to make it last.** That trades the problem for a secret on disk.

Also for #892: `CLAUDE.md` is now **164 lines** against its cap of 150.

PR #917, opened 2026-09-11
The account trap says how to recognise it, and why the variable does not last

muhammadbayoumi:
Three corrections to the trap #913 landed. **Nothing is removed** — every clause of the original survives, including the one control in it (*never park it in a file*). Net **one line**, 164 from 163.

## 1. It reads as a deleted repository

GitHub answers **404, not 403**, on a private repository the active token cannot see. So `Repository not found` is what a *wrong account* looks like — indistinguishable from a repo that is gone.

Measured today: on seeing it I stopped and assumed the worst. The trap now says to ask `gh auth status` before believing it. One line between a correct diagnosis and a quarter of an hour.

## 2. The helper is host-scoped, so the obvious check answers wrongly

```
git config --get-all credential.helper                            -> manager
git config --get-urlmatch credential.helper https://github.com    -> !'…gh.exe' auth git-credential
```

A session verifying *"does `git` really inherit it?"* with the first form concludes gh is not a helper at all — and goes back to `gh auth switch`. The trap now names the **scope** rather than the command, which costs no line.

## 3. The reason was wrong, though the instruction was right

> "It does not survive between commands"

That describes the variable as fragile. `export` survives fine in any shell — what does not survive is the **process**, and each command here gets a new one.

The difference changes behaviour: a session that believes the variable is fragile goes looking for a way to make it last, and the nearest way is exactly what the next clause forbids. Same instruction, true reason.

## The review this file takes

`CLAUDE.md` is exempt from the merge loop but takes one pass, asking two questions:

**Does the edit weaken a control?** No. Mapped clause by clause: repoints every session · the `Repository not found` symptom · the `export GH_TOKEN=…` form · `gh` obeys it over the active account · `git` takes it through gh's helper · put it in front of each command · **never park it in a file**. All present.

**Does each added line change what a session does?** The 404 fact stops a wrong inference; `gh auth status` gives a first action; the host-scoped note stops a false negative from `--get-all`; the process correction gives the right reason for the same action.

## One thing said out loud rather than left to be found

The file is now **164 lines against its own 150**, and its rule is that past 150 a change starts by pruning what no longer changes an action — **never deleting a control to make room**.

I checked every citation in it that can be verified: `scrapex/settings.py:84` (`crawl_obey_disallow` still ships at `0`), `scrapex/directories.py`, `spikes/opfs-sqlite/FINDINGS.md`, `extension/vendor/tabulator.min.js`, and both extension registries. **All present and all still true.** Nothing has stopped being true, so there was nothing to prune, and the alternative would have been to cut a live control.

So the 14-line gap is not something this change can close. It is a separate decision — raise the cap, or move a whole section to a skill the way `record-it` and `review` already were.


muhammadbayoumi:
**Your 404 claim is verified, and your paragraph supersedes an edit I was about
to make.** Measured here before I dropped mine:

```
$ GH_TOKEN=$(gh auth token --user datasouq) gh api repos/muhammadbayoumi/ScrapeX
{"message":"Not Found","status":"404"}

$ GH_TOKEN=$(gh auth token --user muhammadbayoumi) gh api repos/muhammadbayoumi/ScrapeX
muhammadbayoumi/ScrapeX  private=true
```

404, not 403. "A wrong account reads as a deleted one" is exactly right, and it
is the reason this trap gets misdiagnosed: I called it a transient network error
mid-merge on #902 and moved on.

**"Scope the account … instead" fixes a real defect, found independently.** A
review pass over #916 flagged that tightening it to "Scope it" left the nearest
antecedent as the subject of the previous sentence — `gh auth switch` — so the
sentence read as an instruction to scope the very thing the trap forbids, and
dropping "instead" removed the contrast that had carried "not by switching". Your
wording restores both.

**And putting the diagnostic inside the trap is the better call.** #916 moved the
`--get-urlmatch` detail out to `record-it` to pay for two lines under the cap,
which left a session hitting the symptom with no path back to it. The pointer
that would have fixed that costs a line in a file every session loads, for
something rarely needed. "Ask `gh auth status` before believing it" is an
instruction rather than a reference, and it is where the reader already is.

I have dropped my `CLAUDE.md` edit entirely rather than compete with this. What
remains of mine is #918, which touches only `record-it` and deletes a false
instruction there (`gh` is on `PATH` now — `which gh` exits 0). No overlap.

Not taking this PR.

Issue #648, opened 2026-09-05
The browser suites are 59.3% of suite time and 190 seconds of that is unconditional sleeping


muhammadbayoumi:
The browser suites are **59.3% of the suite's time for 8.4% of its tests**, and roughly half of that is the process sleeping on a clock rather than waiting for a condition.

## Measured on `origin/main` 93b31e38

| | tests | time | share |
|---|---|---|---|
| browser suites (10 files) | 308 | **403.3s** | **59.3%** |
| everything else (210 files) | 3,376 | 276.7s | 40.7% |

Non-browser tests average **82 ms**. Browser tests average **1.31 s**.

Per file: `test_panel_dom.py` 263.3s/188 tests · `test_panel_startup.py` 55.6s/29 · `test_grid_dom.py` 32.8s/24 · `test_signing_in_says_what_happened.py` 21.1s/25 · `test_a_mismatched_engine_is_refused_by_name.py` 10.2s/6 · `test_tab_page_dom.py` 8.3s/10.

## It is not browser startup

Chromium launch measured at **0.12s**, and the `browser` fixture is already `scope="module"` (`tests/test_panel_dom.py:117`). Startup is not the cost.

## It is unconditional sleeping - instrumented, not guessed

Wrapping Playwright's `Page.wait_for_timeout` and counting every runtime call, on `origin/main` 93b31e38:

```
countwaits: 448 wait_for_timeout calls asking 189.8s of fixed waiting
browser suites wall clock: 401s
```

**189.8s of the 401s is the process asleep on a fixed timer.**

Then the same suites with every one of those calls capped at 50ms:

```
capwaits: 443 calls, asked 188.5s, spent 22.1s (cap 50ms) -> saved 166.3s
browser suites wall clock: 233s
```

**401s -> 233s, a 168s reduction (-42%), and only 5 of 308 tests failed.** 303 tests do not depend on the wait at all. The five that do are named below and each is waiting for something specific that a condition can express.

The five: `test_dataset_hover_does_not_move_the_card_out_of_its_scrollport`, `test_google_finance_is_a_standalone_responsive_page`, `test_choosing_a_zone_shares_it_with_the_engine_for_the_web_page`, `test_engine_neutral_controls_keep_text_color_on_hover`, `test_a_menu_under_a_small_trigger_is_wide_enough_to_read[320]` - three of them fail for one identical reason, measuring a box mid `scale(.98)` entry animation.

## Where it comes from, and the replacement

[...]

PR #929, opened 2026-09-11
The panel says when it is ready, and the fixture asks instead of sleeping

muhammadbayoumi:
#648's two fixture waits — the ones it measured as the largest single win, at effort S. The
~120 inline sleeps it also names are a separate sweep and are untouched.

```
after page.goto()    wait_for_timeout(500)  ->  wait for the scrapex:shell-interactive mark
after a rail click   wait_for_timeout(400)  ->  wait_for_selector(#view-<name>, visible)
```

## Measured, three runs each side, whole file

| | runs | mean |
|---|---|---|
| baseline | 335.18 · 344.49 · 318.94 | **332.9s** |
| fixed | 195.55 · 232.57 · 224.35 | **217.5s** |

**−115s, −35%**, and the ranges do not overlap: the slowest fixed run is 86s faster than the
fastest baseline one. Three runs a side because this machine carries other sessions and a
single pair would not have been honest — the first comparison I had was 335 against 195, and
the spread turned out to be ~38s.

229 tests pass. None removed, no coverage moved.

## `shell-interactive`, not `fully-settled` — and a failing test is what established it

#648 names both marks. The first attempt used `fully-settled`, and one test failed:

```
tests/test_panel_dom.py::test_the_profile_card_shows_checking_while_chrome_answers
    assert stage.get_attribute("aria-busy") == "true"
E   AssertionError: assert 'false' == 'true'
```

That test opens with `signin_delay_ms=1000` **on purpose**, to observe `aria-busy` while the
account check is in flight. And in `extension/app.js`:

```js
const accountPromise = loadAccount();
Promise.allSettled([accountPromise, enginePromise]).then(() => {
  markStartup("fully-settled", { ... });
});
```

`fully-settled` hangs off the account promise, so stubbing a slow account pushes the mark out
behind the stub's own delay — **the wait was longer than the sleep it replaced**, and it
outlasted the window the test exists to see.

`shell-interactive` is marked immediately after `showView("profile", false)`, before any
promise. That is what this fixture means by open. This is a correction to #648's proposal,
not an execution of it.

## The rail-click wait became a condition, not nothing

Removing it outright also left 229 green — and this file's own docstring says why that is not
evidence: *"A green run under load is not evidence of synchrony; the same-task read is."*
The condition resolves at once, because visibility **is** decided by the click's own task;
what it adds over nothing is that a click which never landed fails there, by name, instead of
later as a missing element.

Only 27 tests take a `view`, so its timing share is ~10.8s — inside this machine's variance.
It is in this PR for correctness, not for the number.

## What was deliberately not touched

`settle_view` waits out a real 180ms entry animation over a measured compositor race, its
docstring carries the numbers, and #648 says explicitly it must stay. It is **not** moved into
the fixture: that would make all 229 tests pay for a wait only the few that measure a box need,
and this file's header rule says the opposite — *read visibility immediately, and never wait
for it; call `settle_view` before you MEASURE.*

## Wider tier, and one thing said plainly

```
python -m pytest -m "extension or docs"   # 1044 passed, 4 skipped
```

An earlier run of that tier reported `1 failed, 1043 passed`. It flipped green on the next
two runs. **I did not capture the test's name before re-running**, so the issue `CLAUDE.md`
asks for against a flipped check cannot be filed accurately, and a nameless flake issue is
noise. Recording it here instead: if that tier goes red once on this branch's CI, the name is
worth capturing rather than re-running.

Unrelated and pre-existing, printed by the same run: `only 35 restores across 1048 tests
(expected >500). The template is being bypassed`.


muhammadbayoumi:
Closing as superseded. #887 landed the boot half while this was open, and did it better: three readiness states (`interactive` / `idle` / `settled`, helpers in `tools/panel_harness.py`) against the single `shell-interactive` wait here. A test that reads what the deferred phase sets needs `idle`, and this branch could not tell it apart.

The measurement in the body — baseline 332.9s mean, fixed 217.5s, −35% — was taken against a base that no longer exists, and most of that win is now on main via #887. It should not be read as the value of what remains here.

**One half was not superseded and is not on main:** the rail click still pays `wait_for_timeout(400)` at `tests/test_panel_dom.py:150`. Reopened as its own small PR against current main.

Worth recording rather than passing over: two sessions worked the same issue at the same time and the collision surfaced only at rebase, after the work was done. That is the sixth such collision today.

Issue #348, opened 2026-09-04
OP-6 · Four review findings were never refuted, and one of them is about test coverage


muhammadbayoumi:
**Status: partly closed, partly unknown.** `REVIEW-2026-07-28` §5 warns that the
refutation stage for eight findings (ت1–ت8) died on a session limit, and that at the
observed rate (6 of 18 fell under refutation) **2–3 of the eight should be wrong**. Do not
treat them as facts.

| claim | status now |
|---|---|
| ت4 `matching.py` rescans the material table per product per name | **fixed** `0a2209c` |
| ت5 `LocalSink` reloads the whole workbook per tab | **fixed** `0a2209c` |
| ت6 the panel rebuilds the log every 1.5s and steals the selection | **fixed** `f901638` |
| ت7 the whole 45-test panel suite silently skipped in CI | **fixed** `48ec48b` + the ≥40 collection floor in `ci.yml` |
| ت1 restore / start-fresh always broken while the worker thread runs | **HALF closed, re-measured 2026-08-12** — *start-fresh* is guarded and tested; `/api/storage/restore` has **no** active-job 409. Refusing is the correct behaviour, and only one of the two routes does it |
| ت2 the heartbeat is written between jobs only, so the engine declares itself dead mid-crawl | **OPEN — and its stated cause is now wrong.** See below |
| ت3 every run rebuilds the full derived timeline of every touched offer | **open** — the entry pointed at DEBT-4; it means **DEBT-3** |
| ت8 the `grid.js` XSS guard scopes itself by splitting on two comments and leaves a gap | **changed** — that guard is gone. Its replacement scopes by line-suffix and by a 14-word `_TEXT_BEARING_FIELDS` allowlist, which leaves two nested markup builders unchecked |

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L535](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L535). `OP-6` is kept in the title because the code cites it.

PR #953, opened 2026-09-16
A held write lock answers 409 everywhere, not an unhandled failure

### LINKED ISSUE #951: A held write lock answers 500 on 18 of 21 write paths, and the panel has no word for 409
`DbLockedError` is not a registered exception handler, so ordinary contention on the
write lock reaches the panel as an unhandled 500 on most write routes.

- `scrapex/webui/app.py:670` registers `_database_unavailable` for
  `(DatabaseUnavailableError, DatabaseMigrationError, DatabaseKindError)` only.
- `grep -c write_lock scrapex/webui/app.py` -> **21**; `grep -n DbLockedError` -> **3**
  (`:2578`, `:2588`, `:4442`). Eighteen sites have no catch, including
  `POST /api/storage/restore` (`:3664`) and `POST /api/storage/start-fresh` (`:3689`).
- The model to copy already exists at `scrapex/webui/app.py:2563-2580`: it maps the
  refusal to a 409 with "a crawl is currently writing to the database — try again
  shortly".
- `grep -rn 409 extension/*.js` (minus `vendor/`) returns **nothing**, so even the three
  protected routes answer a panel with no rendering for the refusal.

This contradicts the rule the lock exists to serve: `DbLockedError` means another app
holds the write permission and the caller should **wait for it** — but a 500 says the
engine broke, which is a different thing to do next. CLAUDE.md: "No silent failures. …
a caught error becomes a visible, structured record."

It is also the precondition every multi-app future shares, whichever way the ownership
question is ruled: more writers means more contention, and contention is not survivable
on 18 of these paths today.

## Options

1. **Register `DbLockedError` in `create_app` and render it in the panel** *(Recommended)*
   — one registration beside `:670`, the three local catches deleted so one place decides
   the refusal, plus the panel string. ~30 lines. Tests: one pytest that an unprotected
   write path answers 409 rather than 500 under a held lock, and one `node --test`
   case for the rendering.
2. **Register it in the engine only.** Half the fix: the status is correct and the owner
   still reads a blank failure.
3. **Do nothing.** Every future app that writes makes this more frequent, not less.

Found while measuring who should hold the warehouse write permission; not fixed there
because `restore` and `start-fresh` are among the eighteen and deserve their own review.

muhammadbayoumi:
Closes #951.

`DbLockedError` means another app holds the single write permission: the thing to
do is wait and press again. It was not a registered exception handler, so that
refusal reached the panel as a server fault on most write routes.

## What was measured, before anything was written

| | |
|---|---|
| `write_lock` sites in `scrapex/webui/app.py` | **21** |
| of those, with a catch of their own | **3** — `:2578` `_write`, `:2588` `_general_write`, `:4442` `/api/capture` |
| with no catch | **18**, including `POST /api/storage/restore` (`:3664`) and `POST /api/storage/start-fresh` (`:3689`) |
| reached through an `/api/` route | **21 of 21** |

That last row is why the handler has no page branch, unlike `_database_unavailable`
beside it: an HTML branch here would be a screen nothing can open.

## The change

One handler and one registration, `scrapex/webui/app.py:673`. Registered rather
than repeated — a per-route catch is a nineteenth chance to forget one.

**The three existing catches stay.** `_general_write`'s names a *different*
database ("the General database is busy"), so it holds different knowledge and
merging it would be wrong. The other two word a refusal their own routes own;
deleting them changes three working routes' sentences and deserves its own diff,
not a ride on this one. A test pins `/api/capture`'s sentence so this change
cannot quietly restate it.

## The guard, and the four mutants it killed

`tests/test_a_held_lock_is_a_retryable_answer.py`, 6 tests. They hold the **real**
lock — real lock file, real in-process gate, real refusal — in another thread, and
patch only the 10 s timeout down to 0.3 s. A fake `DbLockedError` would pass
against a handler registered for the wrong exception, which is the one mistake
this guard exists to catch. One test states the precondition (`the lock really is
held`), so the other five cannot pass vacuously.

Each mutant was watched failing against the code before it:

| mutant | result |
|---|---|
| the registration deleted | killed |
| 409 back to 500 | killed |
| `detail` dropped from the body | killed |
| the handler registered for a different failure | killed |

## A correction to #951

#951 says the panel has no word for 409. That is wrong, and the PR should not
inherit it: `extension/backend.js:156-158` already reads `(await res.json()).detail`
and throws it as the message, so the panel renders **the engine's own sentence**
today. No panel change is needed, and none is here. The issue is corrected in a
comment.

## What this does not move

No route was added, so `contractstamp`'s fingerprint (`{schema, protocol,
endpoints}`) is unchanged and `VERSION` stays at 0.4.14 —
`tests/test_the_version_moves_when_the_contract_does.py` is green (11 tests). No
baseline regenerated, no schema touched.

## What was run

- `pytest tests/test_a_held_lock_is_a_retryable_answer.py` — 6 passed.
- `pytest tests/test_api.py tests/test_api_fields.py tests/test_api_jobs.py
  tests/test_api_review.py tests/test_db.py
  tests/test_the_engine_hands_the_bundle_over.py` — green.
- The full suite was started locally; **CI on this head is the check that counts**
  — a local run on Windows executes no Playwright test
  (`tests/test_panel_dom.py:84` is `pytest.importorskip("playwright")`).


muhammadbayoumi:
**Merge gate, and one boundary this PR must not be read as crossing.**

Step 0 of the gate — read the record — surfaced #348 `ت1`, re-measured 2026-08-12:

> restore / start-fresh always broken while the worker thread runs — **HALF closed**:
> *start-fresh* is guarded and tested; `/api/storage/restore` has **no** active-job 409.
> Refusing is the correct behaviour, and only one of the two routes does it.

**This PR does not close that.** It answers 409 when the **write lock is held**, which is
a different fact from **a job being active**: `tests/test_idle_cost.py:115-145` asserts an
idle engine does not even create the lock file, so a crawl that is between writes holds
nothing and `restore` would proceed exactly as #348 describes.

Recorded here so the next reader does not take "restore now answers 409" for "restore
refuses while a job runs". #348 `ت1` stays open on its own terms.

**Findings ranked in the gate:** no *must fix*, no *should fix*. Two *optional*, filed
rather than fixed here per the skill: #956 (the handler's "21 of 21 are `/api`" claim is
pinned by nothing) and #957 (one condition, five sentences — and the one now reaching the
panel names a pid).

Issue #1004, opened 2026-09-18
STUDY: etendering.tenderboard.gov.om registered vendors — 23,502 firms, 471 requests, genuinely bilingual, and an identifier the page comments out of every row


muhammadbayoumi:
Studied under the `study-a-source` skill (#997), at his instruction, as the source after
Balady (#998). **Nothing was built**: no connector, no `sources.yaml` entry, no
`directories.py` entry, no crawl.

`docs/GULF-EGYPT-SOURCES.md:295-353` already described this source from observation and
called it *"Excellent and the primary Oman source"*. This verifies that description live,
corrects two of its statements, and adds the hidden layer it did not look for.

## 1 · Executive conclusion

**This is a far better source than Balady, and cheaper per record by a factor of fifty.**

- **23,502 registered vendors**, exact: the paginator publishes *"of 471"* pages, page 471
  holds 2 rows, so 470×50 + 2. **Verified.**
- **The list row carries every field.** No per-company fetch is needed for the core
  record, so the whole register is **471 requests** in English — against Balady's 5,470
  for 5,470 offices.
- **Genuinely bilingual, and the join works.** The Arabic view returned the same 50 rows
  in the same order, and **all 50 joined by CR number**. English gives English names and
  categories; Arabic gives Arabic. **Verified** — the opposite of Balady, which has no
  English anywhere.
- **Two controlled vocabularies are published in the page**, with codes, and the
  subcategory taxonomy is a live XML endpoint. The record's §12 said the taxonomy was only
  available as a PDF dated 2021; **it is fetchable live**.
- **An unlabelled eleventh column is commented out of every row** and is the id the
  activities link uses.
- The three per-company surfaces — profile, certificate, activities — are **gated**.

**Politeness:** 2.5 s minimum gap, honest `User-Agent: ScrapeX/0.1`, one connection.
**31 requests**, 3.86 MB, 30×200 and one 404.

## 2 · What it permits

- `etendering.tenderboard.gov.om/robots.txt` → **404. The host declares nothing.**
- No rate-limit headers, no `Retry-After`, no CAPTCHA, no block of a plain client. An
  Apache/Java stack (`JSESSIONID`), 200 to `ScrapeX/0.1` with no cookie on first request.
- The regulation and terms were **not read** in this pass — `docs/GULF-EGYPT-SOURCES.md:357`
  names the PTLC regulation PDF. **Unverified**, and it should be read before any crawl.

## 3 · Identity

[...]

Issue #1003, opened 2026-09-18
How to crawl Balady: a queue crawl, the census as the work list, and a change detector that cannot be a page hash


muhammadbayoumi:
# How to crawl Balady — the design, asked for directly

He asked for the best way to crawl this source, having ruled that it connects to the
crawl now. This is the recommendation, with the measurement behind each choice. Nothing
here is built.

## 1 · It is a queue crawl, and muqawil's shape is the wrong one

Muqawil's crawl is built around a problem Balady does not have. Its listing is *"a cached
random permutation with a generation of 157–282 s"* (`docs/STORAGE.md:124`), so page 40
holds different contractors each time it is read; everything in `partitioncrawl.py` —
cells, sizing, deficit arithmetic, the audit, the resume over stored listing pages —
exists because **the work list itself is unreliable and expensive**.

Balady's work list is reliable and costs three requests. Measured: `start=2000&length=20`
returned the same twenty ids in the same order twice seconds apart and again against the
census taken ~50 minutes earlier. `recordsFiltered` is exact. So:

> **The census is the work queue.** 5,470 items, each carrying its own fetch token, for
> three requests and about six seconds.

That makes this a queue crawl: enumerate, then fetch each item independently. Every item
is independent, so resume, parallelism and partial runs are free rather than engineered.

## 2 · The run

| step | cost | what it does |
|---|---|---|
| 1 · census | **3 requests**, ~6 s, 2.9 MB | the whole register: identity, name, grade, `X`/`Y`, and a fresh token per office |
| 2 · reconcile | **0 requests** | diff the census against `generic_record`: arrivals, departures, and changed census-side fields |
| 3 · order the queue | 0 | arrivals first, then oldest `last_seen_at` first |
| 4 · fetch details | **1 GET per office**, ~218 KB, 2–5 s | the fields the census does not have |
| 5 · stop at the budget | — | cleanly, leaving the queue as it is |

**The census runs every time, and there is no stored queue.** Three requests is cheaper
than the bookkeeping a persistent queue would need, it is exact rather than aging, it
finds arrivals for free — the register grew by 2 in 22 hours — and it re-mints the fetch
tokens so their lifetime never becomes a question.

**Progress lives in `generic_record.last_seen_at`, keyed by `office_id`** — not in the
snapshots. It cannot live in the snapshots: `generic_page_snapshot.source_url` contains
the token, and the token is different on every census response, so a stored snapshot
[...]

Issue #1002, opened 2026-09-18
The census has nowhere to be stored: generic_page_snapshot refuses anything but text/html, and every generic_record must name a snapshot


muhammadbayoumi:
Found while building the Balady census reader (#1001, study #998). Not fixed — it is a
schema decision.

## The two constraints, together

```sql
CREATE TABLE generic_page_snapshot (
    ...
    content_type     TEXT NOT NULL DEFAULT 'text/html'
        CHECK (content_type = 'text/html'),      -- db/engine/schema.sql:194-195
    html_content     TEXT NOT NULL,
```

```sql
CREATE TABLE generic_record (
    ...
    source_snapshot_id    INTEGER NOT NULL        -- db/engine/schema.sql:210
        REFERENCES generic_page_snapshot(page_snapshot_id),
```

- A stored page **must** be `text/html`. A JSON response cannot be stored at all.
- Every `generic_record` **must** name a stored page. There is no path from a parsed row
  to the warehouse that does not go through a snapshot.

The same `NOT NULL` is on `generic_record_revision` (`db/engine/schema.sql:229`) and
`generic_record_node` (`db/engine/schema.sql:696`), and the latter's own comment says the
constraint is the point: *"every membership names the stored page it was read from"*.

## Why it blocks this source and not muqawil

Muqawil publishes HTML listings and HTML profiles, so the snapshot is the document and
the constraint costs nothing. Balady's register comes from a **JSON** endpoint — which is
not a preference, it is the only honest route: the HTML listing's paginator is short by a
page and its region partition loses 843 offices (#998).

So the census — 5,470 offices, three requests, 2.9 MB — has **nowhere to be stored**, and
therefore no `generic_record` can be written from it.

## What that leaves

**Either** the `CHECK` widens to allow `application/json`, **or** the first Balady row in
the warehouse costs **5,470 detail-page fetches**, because the office's HTML detail page
is then the only thing that can be the record's snapshot.

That is the whole cost of the decision, and it is a big one: the cheap 3-request census
that #998 recommended as the first step stops existing as a step. Either the schema
changes or there is no increment smaller than a 3.8-hour crawl.

## Three shapes

1. **Widen the `CHECK`** to a set — `('text/html', 'application/json')` — and rename
   nothing. *Cost:* one migration; `html_content` keeps a name that is now sometimes a
   lie, and `snapshotbody.py`'s dictionary codec is untouched (it compresses text, and
[...]

Issue #418, opened 2026-09-04
OP-118 · A registry source's `crawl_scope` has no door — not in the panel, not in the API


muhammadbayoumi:
> **CORRECTED 2026-09-02, AND THE BLOCKAGE IS GONE.** The paragraph below calls the refusal
> *correct* and repeats the module's own reason for it. **That reason was measurably false.**
> `crawl_partition` passes `listing_phase_only=True` to the walker unconditionally and the
> walker short-circuits on that flag *before* it consults the scope, so a detail page could
> never be fetched from there under any registration. Bypassing only the gate, with the row
> still reading `full_then_listing`, a real cell crawl fetched **9 listing pages, 0 detail
> pages** and closed provably complete — and the fixture raises on a URL with no `?page=`,
> so one detail attempt would have crashed rather than been miscounted.
>
> So the refusal never prevented the ~40,000 requests; it prevented
> `full_then_listing`'s own second phase — *"then the listing catches the changes"*, the
> half its name is about — from running at all. **It is gone**, replaced by
> `test_the_listing_phase_fetches_no_detail_page_under_any_scope`, which asserts the
> behaviour over all three scopes and reddens on the mutation the refusal could not see:
> `listing_phase_only=False`.
>
> **Ask 3 below survives and gets sharper.** The scope moved from `listing_plus_slice` to
> `full_then_listing` as data, against `R-39`, with nothing comparing the two — that is
> still true and still unguarded. **Asks 1 and 2 lose their urgency, not their point**: the
> reason he needed a write was the refusal, and there is nothing to change now to start a
> listing crawl.
>
> The original text is kept below rather than edited, per `C4`: a register that quietly
> replaces a wrong reason with a right one teaches nobody why the check was there.

---
**Migrated from the frozen register on 2026-09-04.** The full measured entry, with its evidence, is at [`docs/archive/BACKLOG.md` L4352](https://github.com/muhammadbayoumi/ScrapeX/blob/main/docs/archive/BACKLOG.md#L4352). `OP-118` is kept in the title because the code cites it.

PR #1005, opened 2026-09-18
The Oman vendor register reads its own identifier out of an HTML comment

muhammadbayoumi:
Step 2 of the order in #1004: the site's own knowledge module for
`etendering.tenderboard.gov.om`, the ESNAD register of 23,502 firms. Reads only -- no
`directories.py` entry, no `sources.yaml`, no crawl, nothing written. Step 3, the registry
entry, is his.

WHY THIS SOURCE AND NOT BALADY FIRST. Measured latency, which neither earlier estimate
used: a Balady detail page has a median of 8,246 ms against this register's 3,521 ms, and
the pace is hard-coded at `DEFAULT_PACE_S = 1.0` on both paths, so the site's latency is
the cost. Balady's 5,470 detail pages are **14.0 hours**; this register's 471 list pages
are **35 minutes** for 4.3x the records. It also sidesteps every blocker Balady has: the
pages are `text/html` so #1002 does not apply, the listing row IS the record so #418's
`listing_only` default is correct rather than an obstacle, and the URL is `?pageNo=N` so
the repository's URL-keyed resume works unchanged.

THE IDENTIFIER IS INSIDE AN HTML COMMENT, and this is the finding the module exists to
encode. The register renders eleven cells and labels ten:

    <th width="4%">S No.</th>
    
    <th width="12%">Company Name</th>
    ...
    <td align="center">1</td>
    

Header and value both commented out -- 50 a page, 23,502 in the register, and
`docs/GULF-EGYPT-SOURCES.md:318-331`'s fourteen-field list does not have it. **An HTML
parser that strips comments loses the record key**, so `read_rows` uncomments deliberately
and asserts the shape while doing it.

WHAT ELSE IS MEASURED INTO IT:

- A data row is one carrying `getProcActivities(`. Selecting by cell count was measured
  wrong: the paginator is a `<tr class="rownohover">` wrapping a nested `Pagination_table`
  and renders nine cells of its own, so a cell-count filter reports 51 rows on a page of
  50 firms.
- **The two copies of the key agree case-insensitively.** Over 152 rows across four pages
  and both languages: identical on 150, case-only on 2 (`ZYPHARSPH` printed against
  `zypharsph` in the call, and `ZZZZZZZ` against `zzzzzzz`), and never otherwise. So the
  cross-check is case-insensitive by measurement rather than by leniency.
- **The short name is a name, not a number.** `ZYPHARSPH` sits beside `0000`, so nothing
  may parse it as an integer or strip its zeros. It is the record key: unique and present
  on 102 of 102 rows measured, where the CR number was blank on 1 of 102.
- **The total is published three ways and all three must agree**:
  `<input name='hidMax' value='471'>` -- the site's own validation source -- `ShowPage(471)`
  on the Last link, and the printed `of 471`, which is split across two table cells and is
  therefore read from stripped text rather than matched through tags. Absence of one form
  is tolerated; disagreement is not.
- **`S No.` is never used as a count.** `pageNo=65536` answered 200 with 51 rows and
  `S No.` from 3,276,751: the server clamps the rows to the real last page and computes the
  label from whatever page was asked for. An out-of-range page is not an empty page, so a
  crawl that stops on "no rows" never stops.
- **The category cell has no separator.** Several categories arrive concatenated --
  `Construction and Maintenance Information Technology Services Pipeline Network
  Construction and Drilling Wells`. `docs/GULF-EGYPT-SOURCES.md:353` warned against one
  comma-separated field; there is no comma. `decompose_categories` matches
  longest-name-first against the vocabulary the page publishes in its own `<select>`, and
  longest-first is load-bearing: `Supply` is a prefix of `Supply and Services (Old)`.
  Measured in #1004 step 1 at **102 of 102 cells decomposing with zero remainder**, and
  **22 of 102 firms (21.6%) holding more than one category**.
- The vocabulary is **read from the page, not hard-coded**, so a thirteenth category is
  data rather than a code change. The measured twelve are in a comment for the reader.
- **The bilingual join is on the CR number, never on position**
  (`docs/GULF-EGYPT-SOURCES.md:338`), and a firm with no CR is reported rather than
  dropped -- 1 of 102.
- The subcategory endpoint honours `CTRL_STRDIRECTION`: 194 subcategories a language,
  the two lists matching in length for all twelve categories. Twenty-four requests are the
  whole official taxonomy, which retires the 2021 PDF at
  `docs/GULF-EGYPT-SOURCES.md:368`.

THE GUARDS, AND THEY WERE MADE TO FAIL. 57 tests, no network. The four fixture rows are
real and each was chosen for a reason the tests then name: a single-category firm, a firm
holding four categories, a firm with **no CR number**, and a firm with no expiry. Then
fourteen deliberate mutations: selecting rows by cell count, dropping the uncommenting,
comparing the key copies case-sensitively, skipping the cross-check, stripping the short
name's zeros, accepting a disagreeing paginator, decomposing shortest-first, decomposing
only the first name, hiding the remainder, sorting the codes, dropping an unpairable firm,
pairing the two views by position, letting an uncommented row through, and dropping the
cell-count check. **All fourteen caught; none survived.**

One of them found a real gap while being written: the two language views happen to arrive
in the same order, so the fixture alone could not catch a positional join. The test now
reverses the Arabic side deliberately.

`ruff check scrapex/` passes. `scrapex/` had zero findings before this file and zero
after; the eleven it briefly introduced were `re.S` aliases and an unstrict `zip`.

Study: #1004. Step 1 (the vocabularies and the decomposition test): #1004 comment. Design and its corrections: #1003.


muhammadbayoumi:
**The body above contradicts the diff, and the diff is what ships.** Recording the
correction here rather than editing the sentence away, because the sentence was a
commitment.

It says: *"Reads only -- no `directories.py` entry, no `sources.yaml`, no crawl, nothing
written. Step 3, the registry entry, is his."*

The diff adds one:

```python
BUILDERS: dict[str, Callable[[], Directory]] = {
    "muqawil_org": _muqawil,
    "oman_tenderboard": _oman_tenderboard,
}
```

That is not cosmetic. `BUILDERS` is what `scrapex/directoryjob.py:114` and
`scrapex/datasetjob.py:189` check to decide whether a key may be crawled at all, so the
entry is exactly the step the body reserves for him.

**He has now ruled: keep it.** The source is approved, and this PR carries its registry
entry.

What the entry does NOT do, so the scope is on the record:

- there is still **no `sources.yaml` entry**, so nothing is `active: true` and no schedule
  fires;
- `DEFAULT_KEY` stays `muqawil_org`, and the diff says why: a default that changes when a
  second directory arrives is at its most expensive exactly then;
- no crawl is started by this change.

Everything else in the body was checked against the diff and holds.

PR #1007, opened 2026-09-18
The review skill becomes portable, and ScrapeX keeps only its profile

muhammadbayoumi:
## What this changes

The `review` skill was one file in this repository and could not be used anywhere else.
It is now two halves:

| half | where | what it carries |
|---|---|---|
| the skill | `~/.claude/skills/review/SKILL.md` — personal, every project | the dimensions, the ranks, the gate, the budget |
| the profile | `.claude/review-profile.md` — this repository | the six things the skill cannot know |

**The skill's own file is not in this diff.** It lives outside the repository by design,
which is the cost of this shape and is stated again under *What this gives up*.

## Why, and how much was actually project-bound

Measured before touching anything: **18 of 127 lines** named ScrapeX — the description,
`CLAUDE.md` four times, the warehouse numbers, `import scrapex`, `panel ↔ engine ↔
warehouse`, and the singular "he". **109 lines were already universal.** This was never a
rewrite; it was 18 lines finding a home.

Those 18 became six slots in the profile:

1. the rules document
2. the controls whose lines a reviewer opens
3. the real scale the performance dimension judges against
4. how the environment is proved before a fan-out
5. the surfaces a cold read crosses
6. how the record is kept

## What is deliberately NOT configurable

The four dimensions, the four ranks, the demonstration rule, the adversary, *"report
nothing rather than pad it"*, the fan-out table and the eight-agent ceiling. These are the
reason the skill is a gate and not an opinion, and a project that could soften them would
have a gate that says whatever that project prefers.

## The rule the split creates

**A project with no profile is reviewed on the skill's defaults and the report says so.**
Same grounds as the existing rule that a reviewer returning no verdict is a *failed pass,
not an empty one*: left unsaid, the performance dimension judges against nothing and comes
back clean. A silently ungrounded dimension is worse than a missing one.

Two more, both to stop the halves drifting: **no `file:line` in either half** — a citation
cannot be true in another repository and is one more thing to drift — and **standing facts
only** in the profile, since a finding or a plan in there is exactly what `CLAUDE.md`
forbids in a repository markdown file.

## Guards

`tests/test_the_documents_cite_what_they_claim.py` watched the skill and now watches the
profile. Its `_read` asserts the file is present, so that row is also the guard that the
profile cannot be deleted out from under the gate that requires it. **Broken on purpose:**

```
$ mv .claude/review-profile.md .claude/review-profile.md.hidden
E  AssertionError: .claude/review-profile.md is in the map but not in the repository
```

Restored, the whole documentation suite is green:

```
python -m pytest -m docs -p no:warnings -q
exit=0   passed=298 skipped=3 failed=0 errors=0
```

And nothing anywhere still points at the old path (`grep -rn "skills/review"` → none).

## `CLAUDE.md` is untouched, on purpose

It already binds every review to the **`review` skill**, and the skill names its profile
in its own first line, so the chain `CLAUDE.md → skill → profile` is complete without a
new sentence. `CLAUDE.md` is also 164 lines against its own 150-line budget, which says the
next change there starts by pruning — and a control may never be pruned to make room.

## The first commit in this branch

`e513ff3` added the named controls to the architecture dimension (the reason this work
started: a comparison against the Code Reviewer agent in
[msitarzewski/agency-agents](https://github.com/msitarzewski/agency-agents/blob/main/engineering/engineering-code-reviewer.md),
which names vulnerability classes where ours named only "the security surface"). That
content is not lost — it is now split correctly: the four classes that hold in any
project are in the skill, and the ScrapeX-specific ones are §2 of the profile, where they
grew from six to ten.

## What this gives up

The skill is no longer versioned in git, no longer guarded by this repository's CI, and a
fresh clone on another machine has no review skill until it is installed there. That is
inherent to a personal skill, and the fix — if it is ever wanted — is to publish the same
two halves as a Claude Code plugin from a small git repository, which changes distribution
only and none of the design above.

## Not merged

Unassigned and unmerged.

---

# Rebased onto `main`, and *What this gives up* is now answered

**Rebased** `23b5855` → `ccb810d`, no conflicts, 0 commits behind. The docs gate is green
on the rebased tree (`python -m pytest -m docs -q`, 2 skipped, 0 failed).

## The paragraph this PR ended on has a home now

It said:

> *"The skill is no longer versioned in git, no longer guarded by this repository's CI, and
> a fresh clone on another machine has no review skill until it is installed there. That is
> inherent to a personal skill, and the fix — if it is ever wanted — is to publish the same
> two halves as a Claude Code plugin from a small git repository."*

**It is wanted, and it is built.** `claude-kit` is a marketplace repository holding the
portable halves of all three skills plus a portable rules base:

```
claude-kit/
├── .claude-plugin/{plugin,marketplace}.json
├── CLAUDE-base.md               135 lines — the loop, the preferences, the traps
└── skills/
    ├── review/SKILL.md          133 lines  ← this PR's other half
    ├── record-it/SKILL.md        70
    └── study-a-source/SKILL.md  168
```

So the split this PR designed is kept exactly; only the **destination** of the portable
half changes, from `~/.claude/skills/` to a tracked repository.

## What the measurement found while this was being prepared

The state this PR describes as future **is already half-present**, and worse than the PR
knew:

| | lines | in git | actually runs |
|---|---|---|---|
| `~/.claude/skills/review/SKILL.md` | **133** | no | **yes** |
| `.claude/skills/review/SKILL.md` | **125** | yes, CI-guarded | no |

They have **diverged**, and the difference is this PR's own work — the personal copy is
already the portable version. Personal skills take precedence over project ones, so the
merge gates run on 2026-09-20 for #1020 and #1024 used **the file git has never seen**,
while CI guarded a file nothing executes.

That is the argument for merging this, not against it: the tracked copy is already
fiction.

## Sequencing — this must not merge first

Merging this removes `.claude/skills/review/SKILL.md` with nothing tracked replacing it,
so until the kit is installed the repository would depend on the untracked personal copy.

1. Create and push `claude-kit`, then `/plugin marketplace add`.
2. `rm -rf ~/.claude/skills/review` — **required**, since a personal skill shadows a plugin
   one for the bare `/review`.
3. Then merge this.

## Still to come, not in this PR

`record-it` and `study-a-source` are still tracked here. Their portable halves are already
in the kit and `.claude/study-profile.md` is written — they move in a follow-up, so this PR
stays the one change it set out to be.


muhammadbayoumi:
State for whoever takes this: #1012. It records the head this PR is green at, the two calls the merging session has to make (whether the documentation exemption applies to 201 changed Markdown lines plus a seven-line test change, and whether to merge at all), and the fact that the skill half of this split lives outside every repository.

muhammadbayoumi:
Claimed by the session holding #1042, on the owner's word, to be worked **after #1042 merges** and not before.

Why that order rather than sooner: this PR edits `.claude/skills/review/SKILL.md`, which is the merge gate #1042 is currently running. Landing it mid-gate would change the rules a pass is being judged by while that pass is in flight, and the pass would have no way to notice.

State as read today: head `ccb810d1`, CI green on all eight contexts, but **2 ahead / 4 behind main** and last touched 2026-09-21 — so that green has expired and the gate's step 1 will need a fresh one on a rebased head.

One overlap already known: `tests/test_the_documents_cite_what_they_claim.py` is also touched by #1081, which appends three rows to `DOCUMENTS` about nine lines below the row this PR edits. That session has checked it and expects git to auto-merge either order.

PR #1008, opened 2026-09-18
The crawl stops naming the tool, and its headers stop contradicting it

muhammadbayoumi:
The owner asked what a published installation tells a site about the person
running it. The answer was: the tool's name, in every request, from every user.

`DEFAULT_USER_AGENT` was `ScrapeX/0.1 (+contact: owner)`. That string buys
nothing back — "owner" is not an address anyone can write to, and no site keys a
robots rule to a name it has never seen — and it costs the one thing a request
can withhold. Everything else a site learns comes from the network (the IP, the
TLS handshake) or from politeness itself (the pace, the 304s, the honoured
`Retry-After`) and is not ours to hide. What a request says about itself is.

**This is not evasion, and nothing that refuses evasion moved.** `HttpFetcher`'s
doctrine rejects "user-agent rotation, proxy rotation, header spoofing, CAPTCHA
handling" and still does. Rotating an identity to get back past a block is a
different act from never announcing one. The crawl presents ONE stable identity
and still stops the moment a site says stop.

## What decides the agent now

One function — `connectors.base.resolve_user_agent` — with four levels, each
answering a different question:

| | level | why it exists |
|---|---|---|
| 1 | the source declares one | Zid 403s anything else (F5) |
| 2 | the owner typed one in the panel | his word beats any automatic choice |
| 3 | **the panel reported its own Chrome** | new, and the default |
| 4 | `DEFAULT_USER_AGENT` | no panel has spoken: the CLI and the tests |

Level 3 is the point. A Chrome string written into the engine ages into a
signature of its own — "who still runs Chrome 125?" — and needs bumping for
ever. The panel knows the browser it runs in, so it reports it and the agent
follows Chrome's own updates. Chrome's reduced UA is low-entropy *by design*, so
this blends into a crowd of millions in a way a hand-written string cannot; and
a site watching one IP browse and crawl now sees one agent rather than two.

It is **reported, never asked**: no field writes those keys, and it **writes
only on change** — a POST costs the engine's write lock, which a running crawl
holds, so comparing first turns one write per panel open into a handful per year.

## The agent was the smaller half

Measured, the crawl sent **5 headers where Chrome sends 13**. A request
announcing Chrome while sending `Accept: */*` and no `Sec-CH-UA` describes a
client that does not exist — arguably a sharper tell than one that never claimed
to be a browser. So `browser_headers` ships with the agent rather than after it:

```
before: 9 of 13 missing
after : 0 of 13 missing
```

**Derived from the agent in use, never from a constant.** A fixed `Sec-CH-UA`
beside an agent that moves with Chrome would announce one version while the UA
announced another. A non-Chrome agent gets no Chrome hints at all, and the
panel's brands ride only beside the panel's own agent — a source's declared
agent and an agent the owner typed each get hints describing *themselves*.

`Sec-CH-UA` is **copied from `navigator.userAgentData`, not computed**: Chrome's
GREASE brand exists precisely so nobody can derive it, and a computed one would
name a browser that does not exist.

## Three defects found on the way

**The rate refresh was sending an empty `User-Agent`.**
`HttpFetcher(**crawl_settings(conn))` splatted the settings dict into the
constructor, passing `user_agent=""` — the shipped default, since the owner has
typed none — straight through. Measured, not inferred:

```
splat path (app.py:3012) sends : ''
```

**We advertised an encoding we cannot read.** `Accept-Encoding` was first
written as Chrome's literal `gzip, deflate, br, zstd`. `brotli` is not
installed, so httpx has no `br` decoder and any server taking the offer would
have answered with a body nothing here could parse. Caught by measurement before
it shipped; the header is now derived from `httpx._decoders.SUPPORTED_DECODERS`,
so installing `brotli` starts advertising `br` with no edit. **A header that
lies about this build is worse than one that differs from Chrome.**

**The precedence chain existed three times** — in `resolve_fetcher`, in the
robots endpoint, and as a two-level version inside the Settings template. Same
knowledge, same reasons to change. The template's copy knew nothing of a
reported agent, so left alone it would have displayed a string no crawl sends.

## One existing guard was re-aimed, and why

`test_every_crawl_setting_is_a_parameter_this_fetcher_accepts` required every
`crawl_settings` key to be an `HttpFetcher` parameter. Its own docstring names
its reason: "`HttpFetcher(**crawl_settings(conn))` is a real call site." That
splat is gone, and it was hiding the empty-agent defect the matching rule could
never see.

`browser_user_agent` and `client_hints` are **inputs to** `resolve_user_agent`
and `browser_headers`, not constructor arguments. So the guard now forbids the
splat anywhere in `scrapex/`, which is strictly stronger — with no splat a key
mismatch is harmless; with one it is fatal. A second test keeps the names that
*do* reach the fetcher pinned, with the exemption named explicitly rather than
derived from "whatever does not match".

## What did not change

robots.txt, `Crawl-delay`, `Retry-After`, the 304 conditional requests, the
circuit breaker, the pacing, and `crawl_obey_disallow` — untouched. A test
asserts it directly: with the new default agent a site's `*` rules still bind
us, its `Disallow` still applies, and `names_us` stays empty. **Hiding the
tool's name must not widen what a crawl may take.**

## Evidence

- `pytest` green; `node --test` 534 extension + 13 apps_script; eslint and ruff clean.
- **28 mutations, 28 caught.** Every fix was broken on purpose and the guard
  watched to fail: the old name restored, each chain level removed, the empty
  agent allowed through, the splat reinstated, the header set dropped, `Accept`
  returned to `*/*`, the hints pinned to one version, the panel's real hints
  ignored, Chrome hints bolted onto a non-Chrome agent, the version gate
  dropped, the failure swallowed, the startup wiring cut.
- **Two mutations came back BLIND and were fixed.** A wiring assertion matched
  the function's *declaration* instead of its call site; and a hints test used a
  non-Chrome agent, so a *different* guard was satisfying it and the gating it
  named was never exercised. Both now caught.

## `VERSION` is not bumped

The two new keys join the capability that already owns `crawl_user_agent` rather
than opening a new one. A new `Capability` moves
`contracts/capability-baseline.json` and so demands a `VERSION` bump, which is
the owner's call — `export-version` confirms the 0.4.15 baseline stands.

## Known limits, stated rather than implied

Header parity is not request parity. The TLS handshake still says Python, header
order is only approximated, and the crawl still speaks HTTP/1.1 where Chrome
speaks HTTP/2 — **that last one is the next PR, by the owner's decision.** A
site reading its access log or matching header rules can no longer tell what
this is; a real anti-bot system still can. It will not learn the tool's name or
the person's identity either way, which was the ask.

The IP is untouched and is not a code question: a VPN on the owner's machine
costs nothing here and covers Playwright too.


muhammadbayoumi:
Merged without the critical review CLAUDE.md requires — the owner's decision, to close the gap now and review after. The debt is recorded in #1010, with what was and was not skipped and what to look at first.

CI green on the merged head `45b134a`: 8/8, including `test` on Linux. 28 mutations, 28 caught. What is skipped is the review pass, not the mechanical gates.

PR #1009, opened 2026-09-18
The crawl speaks HTTP/2, so the handshake stops contradicting the agent

muhammadbayoumi:
**Stacked on #1008** — review that one first; this PR's diff is only the two
files below.

#1008 made the crawl's headers match the agent it claims. This is the layer
underneath them, and the last one reachable without changing how the engine
talks to the network.

HTTP/2 is chosen during the TLS handshake, by ALPN — **before a single header is
sent**. Chrome has spoken HTTP/2 to every site that offers it for years, so a
client announcing Chrome and then speaking HTTP/1.1 contradicted itself in a
place no header could fix, visible to any server without reading one.

```
before: HTTP/1.1  (h2 was not installed, http2 defaults to False)
after : HTTP/2    (verified against a live endpoint: response.http_version)
```

## Not a demand

`http2=True` adds `h2` to the ALPN list **offered**; the server chooses. A site
that speaks only HTTP/1.1 gets HTTP/1.1 exactly as before, so nothing that works
today stops working.

## The failure this could have been

Not a crash — a setting that looks applied and reaches nothing.

httpcore calls `ssl_context.set_alpn_protocols(...)` on the context it is
**handed**, at connection time
([httpcore/_async/connection.py:146](https://github.com/encode/httpcore/blob/master/httpcore/_async/connection.py)).
This project hands every fetcher ONE shared `ssl.SSLContext`, for a measured
reason — 1633ms → 0.6ms per client. So that call mutates shared state, and a
second client sharing the context with `http2=False` would overwrite the ALPN
list: the crawl's protocol would then depend on which connection happened last.

Measured: `HttpFetcher` is the only caller of `shared_ssl_context()`. A test
pins that, because the day a second caller appears with different settings is
the day the negotiation becomes a race nobody is watching.

## Why the dependency is not optional

`httpx[http2]` is a core dependency, not an extra. The crawl presents itself as
Chrome; whether it contradicts that must not depend on how the engine was
installed, and the owner installs from the panel.

There is no quiet fallback when `h2` is missing. Absent means a broken install,
httpx's ImportError names the fix, and degrading silently would leave a crawl
contradicting its own agent with nothing said. Two tests cover the two halves —
that the extra is *declared* (CI builds from pyproject, so a local green proves
nothing) and that `h2` is *importable* (declared is not installed).

## Evidence

- `pytest` green; ruff clean.
- **4 mutations, 4 caught**: http2 switched off, the setting dropped, the extra
  removed from the dependency, and a second caller starting to share the ssl
  context.
- The negotiation itself was verified once against a live endpoint that speaks
  HTTP/2 and came back `HTTP/2`. That is a fact about a server rather than about
  this code, so it is recorded in the test module's docstring rather than
  asserted on every run — no test here reaches the network.

## What is still open

The TLS fingerprint. Python's handshake is not Chrome's, and closing that needs
either a browser (`BrowserFetcher`, already available per source) or a native
impersonating client. `curl_cffi` was considered and rejected: it impersonates a
**pinned** Chrome version, which would re-introduce exactly the mismatch #1008
removed by making the agent follow the panel's real Chrome — besides being a
second HTTP stack and a native wheel on Python 3.14.

So a site reading its access log or matching header rules can no longer tell
what this is; a real anti-bot system still can. Neither learns the tool's name
or the person's identity, which was the ask.


muhammadbayoumi:
Merged without the critical review CLAUDE.md requires — the owner's decision, same as #1008: close the gap now, review after. The debt is recorded in #1010, which covers both.

CI green on the merged head `5f14118` (rebased onto main after #1008 squash-merged): contract-parity, lint, scope, test — all pass. 4 mutations, 4 caught. Local `pytest` green.

What is skipped is the review pass, not the mechanical gates. Two things for whoever reviews: the crawl now takes a hard dependency (`httpx[http2]`), and the shared `ssl.SSLContext` is mutated by httpcore at connection time — a test pins that `HttpFetcher` is its only user, because a second one with different settings would make the protocol a race.

